PR #48, opened 2014-12-12
[linux-xenclient-dom0] enable and use HRTIMERs for better sound

jean-edouard:
Without stubdomains, ioemu talks directly to ALSA, which is too slow when using regular timers at 100HZ.
Using hrtimers speeds up ALSA, and therefore ioemu.


jean-edouard:
We don't want this for now, closing.

PR #56, opened 2014-12-31
[usbback|dom0] Use xenbus shared ring mapper functions.

rossphilipson:
The comment in the code sums up what the problem is but I will recap here.
The existing code was calling alloc_vm_area which update the page tables
for the init process (init_mm). If the process context is not init, the
current process will not have updated page tables. These tables would be
updated through normal #PF faults later. But, since the next call here is
a hypercall, there is not chance to update those table and the hypercall
fails when it cannot find the page through a PT lookup. Thus the error:

(XEN) mm.c:3889:d0 Could not find L1 PTE for address f8684000

Using the xenbus mapping with valloc and vfree handles this by passing
the PTEs to the hypercall.

OXT-69

Signed-off-by: Ross Philipson <ross.philipson@gmail.com>

### DIFF
--- a/recipes-kernel/linux/files/usbback-map-ring-valloc.patch
+++ b/recipes-kernel/linux/files/usbback-map-ring-valloc.patch
@@ -1,0 +1,203 @@
+Subject: [PATCH 001/001] Use xenbus shared ring mapper functions.
+
+From: Ross Philipson <ross.philipson@gmail.com>
+
+The comment in the code sums up what the problem is but I will recap here.
+The existing code was calling alloc_vm_area which update the page tables
+for the init process (init_mm). If the process context is not init, the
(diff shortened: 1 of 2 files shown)


pearsonk:
Built and tested this. Short-term tests are looking good:
- XENVUSB errors (the real ones) have ceased
- Proliferation of phantom USB Host Controllers in the guest have ceased

Doing a long-term test over the weekend using a modified (uncapped iterations) version of the script Ross posted on the OXT-69 JIRA ticket

rossphilipson:
Kevin tested this over the weekend running a script and reported not seeing the error. He gave his thumbs up that this fix is good (at least for fixing OXT-69).

jean-edouard:
Not really my field, but the code looks good and has been thoroughly tested, time to merge!

PR #58, opened 2015-01-03
[cdrtools] fix URL

jean-edouard:


flihp:
LGTM. Rebuilt the cdrtools package with this and everything fetched and built as expected.

PR #61, opened 2015-01-06
refpolicy: drop selinux-policy.git modules in favor of patches

cjp256:
Converted selinux modules located in selinux-policy.git
into individual patches against reference policy.

refpolicy-mcs_2.20130424.bbappend:
- added patch files to SRC_URI
- dropped gitignore.diff from SRCU_URI
- dropped SRCREV_modules usage no longer required
- dropped do_configure_prepend() as it is no longer
  required to insert module sources
- incremented PR

Dropped gitignore.diff patch as it is now just cruft.

OXT-87


cjp256:
I did not try to keep history, as there is none in the selinux-policy.git repo (just the intial commit). That and the files all needed to be converted to patches.

To convert the existing files to patches, I used the following script:

> git clone git://github.com/openxt/selinux-policy.git
> mkdir a b patches
> mv selinux-policy/policy b
> diff -drupN a b > big-patch.patch
> perl -ne 'if(/^diff -drupN (.*) (.*)/) {
>               my $f = $1;
>               $f =~ s{/}{_}g;
>               $f =~ s{^a_}{openxt_}g;
>               open $FH, ">", "patches/$f.patch" or die $!;
>           } else {
>               print {$FH} $_;
>           }' big-patch.patch

pearsonk:
Builds, patch files apply, installs, runs in enforcing, SELinux modules installed, various contexts in /config and /storage verified

flihp:
My builder is back up and building and my test was similarly successful. Good stuff!

PR #62, opened 2015-01-09
acpid: fix recipe regression resulting from oxt-92

cjp256:
acpid was misplaced directly into recipes-bsp, needs to be moved
into its own directory.

OXT-113


jean-edouard:
Tested. This indeed fixes acpid. Merging.

PR #63, opened 2015-01-14
ioemu: fix patch naming for hda audio recording

bennettj1087:
Patch file not ending in .patch was not being applied.


jean-edouard:
My bad... Thanks for the fix!

bennettj1087:
Sorry for the sucky commit message. :) @cjp256 is upset with me. :(

jean-edouard:
Your commit message is nice, short and clear. The form of commit messages is free. We do have some *guidelines*, but they shouldn't be mistaken for compulsory rules. ;)

cjp256:
Specifically, the lack of the sign-off (and the OXT-###)!

PR #64, opened 2015-01-16
acpid warning

eric-ch:
Petty change to silence acpid warning (no /etc/acpi/actions/ac* matching script).


jean-edouard:
I would think -x implies -f, no?

eric-ch:
-x can also be true for directories with search perm, but we are looking for a file here.

jean-edouard:
Fair enough! LGTM, merging.

PR #65, opened 2015-01-22
[uim] disable multi-threaded compilation, as it fails sometimes...

jean-edouard:
This about doubles the build time of the package, from less than 30 seconds to less than a minute.


cjp256:
+1, LGTM - cleaned and built OK

PR #66, opened 2015-02-05
Set installer Dom0 maximum RAM: 64GB

pearsonk:
In order to stop crashes of the installer Dom0 on systems with over 64GB
RAM, set the maximum to within the PAE-addressable range.

OXT-192


rossphilipson:
This LGTM. I will merge it tomorrow if there are no objections.

PR #68, opened 2015-02-11
[linux-i915] Silence gem-foreign debug logs.

eric-ch:
I do not think we still need those and it is a lots of noise for anyone debugging other things.

### DIFF
--- a/recipes-kernel/linux/files/gem-foreign.patch
+++ b/recipes-kernel/linux/files/gem-foreign.patch
@@ -1,8 +1,6 @@
-diff --git a/drivers/gpu/drm/i915/Makefile b/drivers/gpu/drm/i915/Makefile
-index bdcf1e4..848f09d 100644
 --- a/drivers/gpu/drm/i915/Makefile
 +++ b/drivers/gpu/drm/i915/Makefile
-@@ -14,6 +14,7 @@ i915-y := i915_drv.o i915_dma.o i915_irq.o \
+@@ -14,6 +14,7 @@ i915-y := i915_drv.o i915_dma.o i915_irq
  	  i915_gem_gtt.o \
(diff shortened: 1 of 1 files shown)


rossphilipson:
In general I like the idea of turning off that noisy logging. A couple of things though:

 - Some of the changes look odd like this which seems it would break things:
-@@ -14,6 +14,7 @@ i915-y := i915_drv.o i915_dma.o i915_irq.o \
+@@ -14,6 +14,7 @@ i915-y := i915_drv.o i915_dma.o i915_irq

and this:

-@@ -1669,6 +1679,8 @@ int i915_gem_entervt_ioctl(struct drm_device *dev, void *data,
+@@ -1669,6 +1679,8 @@ int i915_gem_entervt_ioctl(struct drm_de

 - Instead of ripping all the logging out, can't we stub it out or disable it with #defs? It was rather useful when there were problems.

rossphilipson:
Actually sorry, the first two are just truncated matching lines (I hate reading patches of patches). I guess I just worry at this point about truncated matching lines breaking some other tool or version of quilt...

eric-ch:
I can redo the patch using the drm printk interface, to it can be toggled. Seems I was a bit lazy :).

About the patch format, this is a valid concern for reviewers imho.
Having to read a patch of patch is painful enough. We might want to ask for a patch format (even provide a quiltrc, since that is likely to be used by bitbake). That would reduce the amount of change in meta-data.

For example, something like :
QUILT_DIFF_OPTS="--show-c-function"
QUILT_PATCH_OPTS="--unified"
QUILT_REFRESH_ARGS="--no-timestamps -p ab --no-index"
QUILT_DIFF_ARGS="--no-timestamps -p ab --no-index"

PR #69, opened 2015-02-16
Fix per-VM alsa config path and audio card

jean-edouard:
OXT-191

### DIFF
--- a/recipes-openxt/xctools/audio-helper/audio_helper_start
+++ b/recipes-openxt/xctools/audio-helper/audio_helper_start
@@ -31,6 +31,15 @@ ALSA_VMID=$DOMID
 ASOUND_CONF=/etc/asound/asound.conf
 ASOUND_CONF_VMID=/var/run/alsa-vm-${ALSA_VMID}.conf
 
+# sound card
+DB_PCM=`db-read /audio/playback-pcm`
+if [ "$DB_PCM" != "" ]; then
+    CARD=`echo $DB_PCM | cut -d '-' -f 1`
+    CARD=$(( 10#$CARD ))
+else
+    CARD=0
+fi
+
 cat > ${ASOUND_CONF_VMID} <<END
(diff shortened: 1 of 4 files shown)


jean-edouard:
Added a commit to make the VMs use the right soundcard, which finishes fixing OXT-191

jean-edouard:
I just tested a whole fresh build with this fix in, along with the toolstack one (https://github.com/OpenXT/toolstack/pull/2).

OXT-191 didn't repro, and I got no (related) SELinux AVC.

If the policy bits don't look too bad, I think this is merge-able, along with the toolstack one.
@flihp, got a sec for a quick look?

jean-edouard:
Actually after a couple of reboots SELinux prevented my VM from booting, with the following AVCs. I'm not sure what they mean...

Feb 20 18:33:35.393435 kernel:[ 3995.069028] type=1400 audit(1424457215.380:1462): avc:  denied  { destroy } for  pid=2434 comm="audio_helper" key=5120  scontext=system_u:system_r:audio_helper_t:s0 tcontext=system_u:system_r:qemu_t:s0:c889 tclass=shm
Feb 20 18:33:35.393441 kernel:[ 3995.069044] type=1400 audit(1424457215.380:1463): avc:  denied  { destroy } for  pid=2434 comm="audio_helper" key=5120  scontext=system_u:system_r:audio_helper_t:s0 tcontext=system_u:system_r:qemu_t:s0:c889 tclass=sem

jean-edouard:
Oh, actually I just saw this AVC on a build that doesn't include this pull request!
So the AVC mentioned above, while it's still mysterious, doesn't have anything to do with this pull request, which should then be safe to merge.

flihp:
I can dig it. Policy looks good to me.

jean-edouard:
Thanks @flihp, what about the two AVCs I mentioned above? Any idea on what they mean and how to fix them?

PR #71, opened 2015-02-25
[linux-i915] Silence gem-foreign debug.

eric-ch:
As some debug calls went into other i915 files, split the debug code
from the original functionality.

Use DRM print methods.


rossphilipson:
LGTM. I was confused at first but now I understand that Eric split the debug stuffs into another patch file.

PR #72, opened 2015-02-25
[xen-tools] Use EXTRA_OEMAKE for DESTDIR.

eric-ch:
Use Makefile variables instead of environment.
Removes redundancy and makes sure that Makefile's DESTDIR variable is
set.

It seems the comment is deprecated since ioemu is built outside of the xen-tools tree, hence disabling it.

### DIFF
--- a/recipes-extended/xen/xen-tools.bb
+++ b/recipes-extended/xen/xen-tools.bb
@@ -63,7 +63,7 @@ INITSCRIPT_PARAMS_${PN}-xenstored = "defaults 05"
 
 EXTRA_OEMAKE += "CROSS_SYS_ROOT=${STAGING_DIR_HOST} CROSS_COMPILE=${HOST_PREFIX}"
 EXTRA_OEMAKE += "CONFIG_IOEMU=n"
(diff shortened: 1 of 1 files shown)


jean-edouard:
This potentially changes behaviors, making DESTDIR "stronger".
Environment variables get superseded by the "=" operator in the Makefile.
However, passing them as EXTRA_OEMAKE makes them invincible.

It's not necessarily a bad thing, just making sure that it's what we want.

flihp:
Any interest in knocking this lose? Makes sense to me. Lines up with the way OE does things generally. I say ship it.

eric-ch:
Making DESTDIR stronger sounds interesting to me. It is expected to be set by the recipe anyway and later stages of BitBake will depend on it.

For some reason, I had a build not setting DESTDIR properly without this patch... I have not investigated much more though.

cjp256:
LGTM - packages-split is identical before & after.  merging.

cjp256:
Actually, it needs a sign-off.

cjp256:
thanks

PR #73, opened 2015-03-24
[OXT-118] Replace Ioemu with QEMU 1.4 in OpenXT

brendank310:
Replace Ioemu with QEMU 1.4 in OpenXT

https://openxt.atlassian.net/browse/OXT-118
https://groups.google.com/forum/#!topic/openxt/FX4sbWt6soI


rossphilipson:
Concerning:
https://github.com/brendank310/xenclient-oe/commit/a8869017fbc1af88a7fbe72db8a60ee15fa2e435
+1 on the first xen part that gives stubdoms privs to deal with IRQs for its guest. I can't really speak to the removal of the ioreq bits but when Eric explained how we were using it, it sounded like it was not right or too useful.

jean-edouard:
Concerning:
https://github.com/brendank310/xenclient-oe/commit/a225ffc75beed49db369ab6e634bef91b7ef4e18
This is not only new patches but also patch modifications. Commit messages and authors for these modifications are lost, which is a shame...

eric-ch:
@rossphilipson since we neither use, or plan to use again, XenGFX the ioreq implementation we had is obsolete. Also it diverges from upstream plans. The rest of the removal is mostly XenGFX and PVM support (PVM is not supported with available plugins anyway).

jean-edouard:
https://github.com/brendank310/xenclient-oe/commit/a8869017fbc1af88a7fbe72db8a60ee15fa2e435 removes the definition of HVM_PARAM_IO_PFN_FIRST, that is used by libsurfman

ktemkin:
@jean-edouard Right, and https://github.com/OpenXT/surfman/pull/3 in turn strips out the (obsoleted) pieces of libsurfman that are broken by removing the deprecated ioreq patch.

cjp256:
Cleaned up commits to meet guidelines.

PR #75, opened 2015-03-30
[xenclient-boot-sound] fix config before adjusting the volumes

jean-edouard:
update-pcm-config fixes the alsa config, which is needed to be able to access the volumes.


rianquinn:
The patch looks good to me. The only thing I would change in addition to what is there, is the set should really be "sset" instead, as amixer's proper syntax is "sset" and "sget".

jean-edouard:
I think that second fix belongs to a separate pull request

rianquinn:
thats fine... I will still do it though.

rianquinn:
There probably should be two fixes to this script. I'm going to create a JIRA ticket to address both issues, so that we can create a fix and separate pull request.

rossphilipson:
LGTM

PR #76, opened 2015-03-30
Fix missing VGABIOS in the xen-firmware package.

ktemkin:
Prior to this commit, xen-firmware is missing vgabios as a dependency;
as a result, xen-firmware is compiled without vgabios.


jean-edouard:
LGTM, thanks!

rossphilipson:
+1 merging

PR #77, opened 2015-03-31
[OXT-238] generic xenstore patch

cjp256:


jean-edouard:
LGTM.
I assume this use case is too specific to use libxcxenstore (https://github.com/OpenXT/xclibs/tree/master/xcxenstore), right?

brendank310:
+1 from me when the recipe INC_PR is bumped up.

cjp256:
Yeah, most of the xenstore functions are already provided in xen_backend.c. This is just so other bits can use them (e.g. atapi-pt) as well as to provide a way to make sure they are ready to be used (before xen_be_init() is called) with xenstore_generic_init().

cjp256:
Pushed fix for 0025 refresh.

brendank310:
+1 with the latest commits

crogers1:
+1.

PR #78, opened 2015-04-01
[OXT-215] atapi-pt

cjp256:


jean-edouard:
Re-assigning the CD from one VM to another using the UI doesn't seem to work.

jean-edouard:
Changing the CD doesn't seem to work either, the VM still "sees" the previous CD somehow...

brendank310:
Did you use this in conjunction with the associated pull request from xctools.git?

jean-edouard:
And dm-agent yes

cjp256:
@jean-edouard:
"Re-assigning the CD from one VM to another using the UI doesn't seem to work."
-> is autolock enabled? did you set the sticky bit (always use with this vm)? can you validate with 'xec -o /host list-cd-devices'? In my testing I have sometimes noticed the UI data is stale...

"Changing the CD doesn't seem to work either, the VM still "sees" the previous CD somehow..."
-> windows or linux? i'll look into that.

cjp256:
This last commit should resolve the autolock assignment issues.

As far as Windows still "seeing" the CD when flipping the lock to another VM, I'm investigating.

cjp256:
Hopefully addressed the above issues, and rebased the last two commits.

jean-edouard:
Now my basic tests succeed.

The amount of code here is too big to be reviewed line by line.
After a quick overview,  the code looks well documented, so debugging it in the future shouldn't be too miserable.

As this is the only available way to get atapi passthrough in our new qemu 1.4, I'll merge it in 24+ hours if nobody objects.

rossphilipson:
in 24 hours....+1

cjp256:
whoops, oxt-122 accidentally snuck in here unnoticed after I was testing it.  I'll need to rebase first.

cjp256:
pulled oxt-122 back out, should be gtg now.

jean-edouard:
24 hours have passed, merging this pull request and the other two in dm-agent and xctools now.

PR #79, opened 2015-04-01
Add additional resolutions (back) to vgabios.

ktemkin:
Adds a collection of common supported "resolution" modes for HVM guest display.
These allow for a more natural-appearing display on many monitors. See
added patch for additional information.

Signed-off-by: Kyle Temkin <temkink@ainfosec.com>


eric-ch:
LGTM, +1.

brendank310:
I've tested preliminary forms of the patch on affected laptops (1366x768 panels) and this works well. +1

rossphilipson:
LGTM. This patch was lost in the move to qmeu 1.4 so it makes sense to bring it with us. Merging.

PR #80, opened 2015-04-01
qemu: Refactored the ISO media changing code for QEmu to use the generic

brendank310:
Xenstore watch callback. Detailed explaination of changes is available
within 0024-Xenstore-based-ISO-media(...).patch.

OXT-122


recipes-openxt/qemu-dm/files/0024-Xenstore-based-ISO-media-change-detection-for-both-s.patch
```diff
@@ -0,0 +1,462 @@
...
++#else
++    dompath = xs_get_domain_path(xenstore, xen_domid);
++#endif
++
++    if (!dompath) {
++        fprintf(stderr, "%s: Failed to retrieve dompath", __FUNCTION__);
++        return -1;
++    }
++
++    snprintf(path, XEN_BUFSIZE, "%s/device/vbd", dompath);
```
cjp256: I generally would encourage the use of sizeof() rather than constants.

recipes-openxt/qemu-dm/files/0024-Xenstore-based-ISO-media-change-detection-for-both-s.patch
```diff
@@ -0,0 +1,462 @@
...
++                if (node && params &&
++                    node->xen_vbd_id == NULL &&
++                    node->frontend_state == NULL) {
++                    // For mapping a fired watch to a specific device later
++                    node->xen_vbd_id = strdup(be);
++
++                    // Before the guest disk change can occur, make sure the state
++                    // of the specified blkfront device is ready
++                    memset(path, 0x00, XEN_BUFSIZE);
++                    snprintf(path, XEN_BUFSIZE, "%s/device/vbd/%s/state", dompath, vbd_devs[i]);
```
cjp256: Should null terminate path.

recipes-openxt/qemu-dm/files/0024-Xenstore-based-ISO-media-change-detection-for-both-s.patch
```diff
@@ -0,0 +1,462 @@
...
++            QTAILQ_FOREACH(node, &iso_list, next) {
++                if (node && params &&
++                    node->xen_vbd_id == NULL &&
++                    node->frontend_state == NULL) {
++                    // For mapping a fired watch to a specific device later
++                    node->xen_vbd_id = strdup(be);
++
++                    // Before the guest disk change can occur, make sure the state
++                    // of the specified blkfront device is ready
++                    memset(path, 0x00, XEN_BUFSIZE);
```
cjp256: Not necessary.

recipes-openxt/qemu-dm/files/0024-Xenstore-based-ISO-media-change-detection-for-both-s.patch
```diff
@@ -0,0 +1,462 @@
...
++    char path[XEN_BUFSIZE];
++
++    if(!xenstore || !node)
++        return;
++
++    if(!node->first_watch) {
++        memset(path, 0x00, XEN_BUFSIZE);
++
++#ifdef CONFIG_STUBDOM
++        snprintf(path, XEN_BUFSIZE, "%s/state", node->xen_vbd_id);
```
cjp256: I generally would encourage the use of sizeof() rather than constants.

brendank310:
Mind clarifying? The dev_type here doesn't come from qemu, it's from the blkfront drivers' Xenstore nodes. I want to ensure I'm not scooping up a block device that isn't a CDROM. The comment about mapping is out of date, but the backend string is still used during the callback.

recipes-openxt/qemu-dm/files/0024-Xenstore-based-ISO-media-change-detection-for-both-s.patch
```diff
@@ -0,0 +1,462 @@
...
++
++int xenstore_init_iso_dev(void)
++{
++    unsigned int dirNum = 0, i = 0;
++    char *dompath = NULL;
++    char **vbd_devs = NULL;
++    char path[XEN_BUFSIZE];
++    char token[XEN_BUFSIZE];
++
++    memset(path, 0x00, XEN_BUFSIZE);
```
cjp256: not necessary?

recipes-openxt/qemu-dm/files/0024-Xenstore-based-ISO-media-change-detection-for-both-s.patch
```diff
@@ -0,0 +1,462 @@
...
++    return 0;
++}
++
++int xenstore_init_iso_dev(void)
++{
++    unsigned int dirNum = 0, i = 0;
++    char *dompath = NULL;
++    char **vbd_devs = NULL;
++    char path[XEN_BUFSIZE];
++    char token[XEN_BUFSIZE];
```
cjp256: unused?

cjp256:
None of my comments would block this, it looks good to me. +1

The only suggestion I would have is to ditch the linked list, but since the plan is to eventually dump this in favor of a QMP-based approach, it's probably not worth the effort.

cjp256:
Also, built and tested (stubdom=true).

jean-edouard:
The compile-time stubdom option is something we used to have with ioemu, and that we decided to get rid of when updating to qemu 1.4, in preference to having two (slightly) different patchqueues.

I don't necessarily think this was a good idea, but we have to make a decision one way or the other. Having both seems redundant...

Either ways, that kind of core change to the way we build qemu should be part of a separate pull request.

rossphilipson:
I completely agree we should only have one means of generating stubdom/non-stubdom versions of qemu. I think the #ifdef CONFIG_STUBDOM method is better personally. It seems like a pq that is dependent on being applied on top of another pq could get pretty messy.

brendank310:
I don't see the sense in a separate pull request if the only patches using the different patch queues are being replaced by this request. All other patches are being applied from within qemu-dm.inc.

rianquinn:
I agree that the CONFIG_STUBDOM is the better approach (which should be what is in there now). I do not agree that this needs to be a separate pull request. This is currently the only patch that needs this, so it's logical for the two to be combined

transpute:
The discussion about "what requires a separate pull request" would benefit from a mailing list discussion, so it can be added to the project guidelines.

jean-edouard:
What I mean is that if we decide to go with CONFIG_STUBDOM, there should be a separate pull request to switch the existing patches to that.

What adds the CONFIG_STUBDOM options can be in this pull request, but should be in a separate patch.

rianquinn:
this is the only patch in the queue that needs that at the moment. All of the other patches work regardless of the recipe. In other words, there are no patches to "change". Since they are tied together it's fine to leave them in the same pull request. Future QEMU patches (assuming we have any) that need to know the difference can/should use this switch

jean-edouard:
The new flag "--enable-openxt-stubdom" is a distinct feature, that may get used by other patches like you mentionned.

If in the future we decide to rewrite ISO media change detection (to use QMP for example), the new patch may not even use the stubdom flag anymore. If in the meantime other patches were written using that new flag, bad stuff will happen...

rianquinn:
This pull request will be dropped till a later date.

jean-edouard:
I rebased this pull request there: https://github.com/OpenXT/xenclient-oe/pull/149
@brendank310 please close this.

brendank310:
Haven't really had the opportunity to get this up to snuff. Thanks @jean-edouard for the rebase.

PR #82, opened 2015-04-08
[usbutils] avoid loop in linked list...

jean-edouard:


rossphilipson:
On face value the fix looks like it will prevent crashes but you might not get the correct output. But, it looks like you do get the right output and Jed said he got the fix from a diff w/  a newer version. It prevents the infinite loop, so that is a major fix. Also our package is down rev and presumably we will move to a newer one. Given all that, merging...

PR #83, opened 2015-04-08
New audio code

rianquinn:
New audio code

Long description:
These changes bring in a bunch of audio fixes, including a new audio
front/back driver model for QEMU 1.4. The following was addressed:
- General quest instability
- Performance issues with capture and playback (there were a lot
  of unneeded world switches due to v4v API)
- Security cleanup (the old code had known buffer overflow attacks
  and handle rewrite attack vulnerabilities)
- Added error validation (the old code had little to no error checks)
- Bugs with amixer patch. The patch was moved to the audio helper removing
  the need to patch ALSA. This was a relatively easy thing to do as the
  audio helper and amixer have to do the same kind of tasks
- Changes to init script (the old init script over-wrote the ALSA state
  on reboot, so settings in the UIVM would roll-back on each reboot)
- Patch queue cleanup
- Recipe cleanup
- SELinux work to support init script

Bugs:
OXT-133: Qemu 1.4 crash during tools install
OXT-220: Qemu 1.4: sometimes the VM volume defaults to 0
OXT-234: amixer patch
OXT-189: Mic mute not working
OXT-196: Audio on Dell Latitude E6540
OXT-195: Audio on HP 2570p
OXT-194: Audio Capture Setting not working

Singed-off-by: Rian Quinn <quinnr@ainfosec.com>


recipes-openxt/qemu-dm/qemu-dm.inc
```diff
@@ -12,15 +12,19 @@ DEPENDS = "xen-tools alsa-lib pciutils libpng blktap libxenacpi libv4v openssl z
 #            file://0010-xc-emulated-nic-link-state-propagation.patch;striplevel=1 \
 #            file://0012-audio-alsa.patch;striplevel=1 \
 #            file://0014-audio-alsa-stub.patch;striplevel=1 \
-#            file://0020-audio-policy.patch;striplevel=1 \
 #            file://0026-openxtaudio.patch;striplevel=1 \
+#            file://0007-pci-reserve-igd-slot.patch;striplevel=1 \
```
cjp256: Dropping 0007-pci-reserve-igd-slot.patch patch on purpose?

rianquinn: Yes

rossphilipson:
VM switching works nicely. This was a problem before. Mixing music between guests works well too.

rianquinn:
One thing that was fixed was the total number of V4V packets being sent for both playback and capture. For playback it went from 4 to 3, and for capture it went from 6 to 2. The result is you can now do things like capture and playback at the sametime, a test that always failed for us prior.

rossphilipson:
We have done a bunch of testing and the audio code is looking solid. There were two issues that came out of it that got tickets:

This one is minor and not a regression:
https://openxt.atlassian.net/browse/OXT-257

This one is a regression since it did not seem to be an issue when we were using HDA:
https://openxt.atlassian.net/browse/OXT-258

At this point we would like to go ahead and merge these changes. I am going to hold off on the merge for a short period in case there are any major objections. I will shoot to merge it at 12:30 today.

rossphilipson:
Ok Rian fixed the merge problem. Merging...

PR #84, opened 2015-04-08
[oxt-246] pciutils: update pci.ids from 20110801 to 20141108.

cjp256:
OXT-246


cjp256:
It comes out of https://www.kernel.org/pub/software/utils/pciutils/pciutils-3.3.0.tar.xz, a file named 'pci.ids' that I gzipped to match the current approach.

Normally this could be handled by uprev-ing pciutils, but I'm hesitant because it looks like someone already just backported the 2011 version IDs instead of updating the package.

PR #70, opened 2015-02-20
ca-certificates: Fix broken SRC_URI

bennettj1087:
Update SRC_URI to snapshots.debian.org. Apparently they removed
the 20120212 version from the main repo.

Signed-off-by: Justin Bennett <bennettj@ainfosec.com>

PR #85, opened 2015-04-09
Use PREMIRRORS_append instead of overriding the SRC_URI variable.

crogers1:

### DIFF
--- a/recipes-support/ca-certificates/ca-certificates_20120212.bbappend
+++ b/recipes-support/ca-certificates/ca-certificates_20120212.bbappend
@@ -1,0 +1,4 @@
+PRINC = "1"
+
+PREMIRRORS_append = "\
+ftp://.*/.* http://snapshot.debian.org/archive/debian/20120229T213309Z/pool/main/c/ca-certificates/ca-certificates_${PV}.tar.gz\n"


crogers1:
Alternate approach to PR: https://github.com/OpenXT/xenclient-oe/pull/70 that uses @eric-ch's recommendation.

jean-edouard:
Can you please explain what that variable does? What's "ftp://.*/.*"?

crogers1:
When there are extra entries in PREMIRRORS, OE will try to replace the existing url in the src_uri during the fetch step.  In the urI_replace function it decodes the urls into 2 lists and compares them.  The ftp://.*/.* is basically a regex to tell the fetcher how to parse the url and what it should match on.

For example, if you had multiple src_uris, one http, one git, and two ftp, and you wanted your PREMIRROR entry to only apply to the ftp entries, the ftp://.*/.* regex will ensure OE only replaces the ftp src_uris when attempting to fetch.

jean-edouard:
What about the following URL?
http://sources.angstrom-distribution.org/ca-certificates_20120212.tar.gz
It looks more ... permanent...

Or shouldn't that kind of old tarball be in the mirror?

rossphilipson:
@crogers1 - ping, Jed has a question on this - that is why it is stuck here.

crogers1:
To be honest I wasn't really sure on this one.  Was hoping someone with more knowledge of what we wanted would chime in.

rossphilipson:
I was told this doesn't appear to be necessary any longer. Closing.

PR #86, opened 2015-04-10
Add kernel support for fallback PCI reset.

ktemkin:
This set of patches adds support for "fallback" PCI reset techniques,
in support for PCI (and especially GPU) passthrough. See the individual
added patches for details-- documentation of the added code is present
in the patch headers.

Signed-off-by: Kyle J. Temkin <temkink@ainfosec.com>

### DIFF
--- a/recipes-kernel/linux/linux-xenclient-dom0-3.11/0001-Add-thorough-reset-interface-to-pciback-s-sysfs.patch
+++ b/recipes-kernel/linux/linux-xenclient-dom0-3.11/0001-Add-thorough-reset-interface-to-pciback-s-sysfs.patch
@@ -1,0 +1,534 @@
+From 02a1f1b44e8e81bedf7fccbe210aaec108be33b8 Mon Sep 17 00:00:00 2001
+From: Kyle Temkin <ktemkin@binghamton.edu>
+Date: Wed, 8 Apr 2015 00:58:24 -0400
+Subject: [PATCH 1/2] Add thorough reset interface to pciback's sysfs.
+
+--------------------------------------------------------------------------------
+SHORT DESCRIPTION:
+--------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


rossphilipson:
Ok I like the logic in __pcistub_raw_device_reset. That seems the safest. You are missing a check for PCI (non-express) FLR capabilities.

ktemkin:
I've updated this to use the Advanced Functionality-spec FLRs when possible. 

There shouldn't be any need for additional logic to distinguish between PCI and PCIe for the two checks-- the first check (for PCIe FLR capability) should only ever return true on a PCIe card, and the second should only return true on a "legacy" PCI card.  The inline comments (e.g. https://github.com/ktemkin/xenclient-oe/commit/bff1165fccd85e812d24b23d5b477619a02fd199#diff-9e3e8974ea7f0bafe95933c8df0f6374R209) explain this in more depth.

This is the same way the Linux kernel handles things (https://github.com/torvalds/linux/blob/master/drivers/pci/pci.c#L3358): each card tries a PCIe FLR. If that fails, it moves on to a PCI AF FLR. The probe code executed in this patch winds up being exactly the same as the probe code from the linux kernel.

ktemkin:
I plan on testing this (hopefully final!) revision in the morning. Here's an ISO with the relevant fix, for anyone who wants to test:

http://dev.ktemkin.com/openxt/pci_fix_and_vgabios_fix_installer.iso

rossphilipson:
The code LGTM and ack that you don't need to test for PCI vs. PCIe first. I will grab the ISO and test on our HP 8300.

ktemkin:
All of my quick re-tests passed over here-- this seems to fix the issue on my end.

rossphilipson:
Tested here, looking good. Merging.

PR #87, opened 2015-04-13
[VGABIOS] Disable 24-bit color modes for Surfman compatibility.

ktemkin:
The zero-copy Surfman display frontend currently does not support 24-bit
color modes. This commit modifies our extended mode patch to ensure that
24-bit color modes are not advertised to the guest.

Signed-off-by: Kyle J. Temkin <temkink@ainfosec.com>


ktemkin:
Looks like we missed one more patch when the VGAbios change occurred:

https://github.com/OpenXT/xenclient-oe/blob/aed28ed5435ebfc7d58001346348c225d80b5a1d/recipes-extended/xen/files/disable-24bit-vbe-modes.patch

Without this patch, neither Windows XP nor Windows 8 can install properly, as their installers use improper 24-bit color modes.

ktemkin:
Here's an ISO with the fix built in, if anyone wants to test:

http://dev.ktemkin.com/openxt/pci_fix_and_vgabios_fix_installer.iso

eric-ch:
Stupid me ... +1.

transpute:
Tested ISO and Windows 8 installer loaded successfully.

jean-edouard:
I'm testing this now

jean-edouard:
Tests successful (Debian installers), LGTM, merging.

transpute:
After further testing, I'm seeing a black screen when booting Win 8.1.  The initial install succeeded, and a subsequent reboot was also successful, but then all subsequent reboots briefly show the blue window icon, then a black screen -- never reaching the Windows desktop.

I also saw a failure to install Win 8 guest tools, which I initially thought was the cause of the subsequent black screen, but I have now done a new install without guest tools, and still see the black screen.  I now have two VMs exhibiting this failure (one with a failed tools install, one with no tools), let me know if you need logs or diagnostics.

PR #88, opened 2015-04-13
[udev] use pci.ids and usb.ids for providing complete device information

jean-edouard:

### DIFF
--- a/recipes-bsp/pciutils/pciutils.inc
+++ b/recipes-bsp/pciutils/pciutils.inc
@@ -37,7 +37,11 @@ do_install () {
 	# "make install" misses the debug file for the library
 	oe_libinstall -so -C lib libpci ${D}/${libdir}
 
+	install -d -m 0755 ${datadir}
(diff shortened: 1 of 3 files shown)


rossphilipson:
"...once xenmgr is patched to use the non-gzipped file"

Maybe we should fix it here?

jean-edouard:
This pull request is basically adding the uncompressed version of pci.ids. It doesn't remove the compressed one because xenmgr depends on it.
Xenmgr, and everything else that uses the compressed version, should be modified to use the uncompressed one, so that we can remove the compressed file and save 200KB of disk space for the dom0 rootfs.
This pull request can stay open until xenmgr, and potential others, are ready to use the uncompressed file.

jean-edouard:
PR submitted against xenmgr to use the uncompressed file and added a commit to get rid of the compressed one.
Please merge at the same time as https://github.com/OpenXT/manager/pull/11

rossphilipson:
Just tested, +1, merging

PR #89, opened 2015-04-15
[OXT-250] Update Xen from 4.3.3 to 4.3.4, rolling back some hand rolled XSA patche...

brendank310:
...s to use upstreams latest.

### DIFF
--- a/recipes-extended/xen/files/xc-xt-interrupt-debug-info.patch
+++ b/recipes-extended/xen/files/xc-xt-interrupt-debug-info.patch
@@ -2,7 +2,7 @@ diff --git a/xen/arch/x86/irq.c b/xen/arch/x86/irq.c
 index b0b0c65..47d3e4f 100644
 --- a/xen/arch/x86/irq.c
 +++ b/xen/arch/x86/irq.c
-@@ -2100,9 +2100,102 @@ void free_domain_pirqs(struct domain *d)
+@@ -2100,9 +2100,97 @@ void free_domain_pirqs(struct domain *d)
      spin_unlock(&pcidevs_lock);
  }
  
@@ -22,11 +22,6 @@ index b0b0c65..47d3e4f 100644
 +        printk("%04x:", apic_read(APIC_TMR + i*0x10));
 +    printk("\n");
 +}
-+
-+#define vlapic_test_vector(vec, bitmap)                         \
-+    test_bit(VEC_POS(vec),                                      \
-+                     (unsigned long *)((bitmap) + REG_POS(vec)))
-+
 +
 +static void dump_lapic_register(struct vlapic *l,char *name,int base)
 +{
--- a/recipes-extended/xen/files/xc-xt-v4v.patch
+++ b/recipes-extended/xen/files/xc-xt-v4v.patch
@@ -6,9 +6,9 @@ index ba57e5a..f2073e9 100644
      HYPERCALL(hvm_op),
      HYPERCALL(sysctl),
      HYPERCALL(domctl),
--    HYPERCALL(tmem_op)
-+    HYPERCALL(tmem_op),
-+    HYPERCALL(v4v_op)
+     HYPERCALL(tmem_op),
++    HYPERCALL(v4v_op),
+     [ __HYPERVISOR_arch_1 ] = (hvm_hypercall_t *)paging_domctl_continuation
  };
  
  #define COMPAT_CALL(x)                                        \
@@ -16,9 +16,9 @@ index ba57e5a..f2073e9 100644
      HYPERCALL(hvm_op),
      HYPERCALL(sysctl),
      HYPERCALL(domctl),
--    HYPERCALL(tmem_op)
-+    HYPERCALL(tmem_op),
-+    HYPERCALL(v4v_op)
+     HYPERCALL(tmem_op),
++    HYPERCALL(v4v_op),
+     [ __HYPERVISOR_arch_1 ] = (hvm_hypercall_t *)paging_domctl_continuation
  };
  
  int hvm_do_hypercall(struct cpu_user_regs *regs)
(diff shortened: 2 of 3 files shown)


brendank310:
Required:
https://github.com/OpenXT/openxt/pull/56

recipes-extended/xen/files/xc-xt-interrupt-debug-info.patch
```diff
@@ -23,11 +23,6 @@ index b0b0c65..47d3e4f 100644
 +    printk("\n");
 +}
 +
-+#define vlapic_test_vector(vec, bitmap)                         \
```
rianquinn: Do we know why this is removed?

recipes-extended/xen/files/xc-xt-v4v.patch
```diff
@@ -6,9 +6,9 @@ index ba57e5a..f2073e9 100644
      HYPERCALL(hvm_op),
      HYPERCALL(sysctl),
      HYPERCALL(domctl),
--    HYPERCALL(tmem_op)
-+    HYPERCALL(tmem_op),
-+    HYPERCALL(v4v_op)
+     HYPERCALL(tmem_op),
++    HYPERCALL(v4v_op),
+     [ __HYPERVISOR_arch_1 ] = (hvm_hypercall_t *)paging_domctl_continuation
```
rianquinn: Why exactly is this part changing? Just curious

recipes-extended/xen/files/xc-xt-interrupt-debug-info.patch
brendank310: It is already defined in another header.

recipes-extended/xen/files/xc-xt-v4v.patch
brendank310: Upstream change to the hypercall table has an anchor now, so in order to have the v4v patch apply, it was modified to contain the anchor.

cjp256:
LGTM. Installed build derived from autobuilder, boots OK here:
 __  __            _  _    _____ _  _              
 \ \/ /___ _ __   | || |  |___ /| || |   __  _____ 
  \  // _ \ '_ \  | || |_   |_ \| || |_ _\ \/ / __|
  /  \  __/ | | | |__   _| ___) |__   _|__>  < (__ 
 /_/\_\___|_| |_|    |_|(_)____(_) |_|   /_/\_\___|
                                                   
(XEN) Xen version 4.3.4-xc

cjp256:
Rebased on master.

brendank310:
As it stands this can't be merged. The xc-xt-v4v.patch doesn't apply correctly to hvm.c. The changes to the hvm_hypercall64 table and hvm_hypercall32 table are too similar, and patch seems to ignore the second hunk. I had done testing on 64 bit HVM so it didn't expose the problem. If I break those two hunks into separate patches they work just fine, but it's an ugly solution. Playing with fuzz options didn't yield anything, and I'm not sure how well that'd play with bitbake anyhow. Anyone have any ideas?

rianquinn:
Have we tried applying the modifications using quilt? It might generate a more sane patch.

rossphilipson:
Wow that is crazy. The patch is correct, the offsets are correct. Actually that is scary.

ktemkin:
The offsets are wrong in the _unified diff_ header for that patch. As a result, patch is viewing everything after the first close-brace as "trailing garbage" and skipping it.

Running the patch manually with --verbose yields:

```
checking file xen/arch/x86/hvm/hvm.c
Using Plan A... 
Hmm...  Ignoring the trailing garbage
done
```
This is an unfortunate result of a patch "feature" in which it ignores anything it doesn't recognize at the start or end of a patchset-- the same feature that lets us use the output of git format-patch directly as patches.

Fixing the start/end indicators makes the patch apply correctly.

```patch
index ba57e5a..f2073e9 100644
--- a/xen/arch/x86/hvm/hvm.c
+++ b/xen/arch/x86/hvm/hvm.c
@@ -3365,8 +3365,9 @@
     HYPERCALL(hvm_op),
     HYPERCALL(sysctl),
     HYPERCALL(domctl),
     HYPERCALL(tmem_op),
+    HYPERCALL(v4v_op),
     [ __HYPERVISOR_arch_1 ] = (hvm_hypercall_t *)paging_domctl_continuation
 };
 
 #define COMPAT_CALL(x)                                        \
@@ -3385,8 +3386,9 @@
     HYPERCALL(hvm_op),
     HYPERCALL(sysctl),
     HYPERCALL(domctl),
     HYPERCALL(tmem_op),
+    HYPERCALL(v4v_op),
     [ __HYPERVISOR_arch_1 ] = (hvm_hypercall_t *)paging_domctl_continuation
 };
 
 int hvm_do_hypercall(struct cpu_user_regs *regs)
```

Was this patch modified by hand? It'd be a pretty bad bug if some of our software was generating malformed patches.

brendank310:
Yes (and no). As committed yes it was modified by hand. After I saw that behavior I went down the diff -u route. Oddly enough, it still wouldn't apply to the second hypercall table. I even went so far as adding additional context (-U <number>) with no luck. 

Is there anything we can do to enforce some strict type of patch application? It bothers me that this failed silently.

ktemkin:
Unfortunately, there's no easy way that I know of: patch always allows "garbage" to be stripped without more than that note. Even if we could somehow error specifically on *trailing* garbage, that would break compatibility with a lot of our existing patches: many of them are generated with trailing garbage by git.

We might be able to come up with some kind of heuristic method to error out on these (e.g. error out if the number of hunks applied don't match the number of _unified diff_ `@@ xx @@` headers), but I can't think of anything as easy as adding a command-line option to our patch.

I think I'd chalk this up to a downside of patch-queues. =(

ktemkin:
+1.

With the corrected patch, I was able to boot a 32-bit and 64-bit guest and verify basic functionality (including Xenstore reads from inside of Windows 32-bit guest). 

(Here's an ISO with the 4.3.4 upgrade, in case anyone wants to test: http://dev.ktemkin.com/openxt/xen_4_3_4_openxt.iso)

rossphilipson:
This was tested last week and Kyle fixed the patch issue and tested over the weekend. Looks good, merging.

PR #92, opened 2015-04-17
Use ACPI guest firmware that matches upstream QEMU

rossphilipson:
We used to use the ACPI firmware for QEMU traditional but now we want the
version that matches our new QEMU. Note gpe0_blk_len = ACPI_GPE0_BLK_LEN_V1
is left intact. This matches the new QEMU and is 1/2 as big as the traditional
GPE space. This is also why all of the GPE's in the patch are changed to fit
in 16 bits now. Finally we let the Seabios ACPI builder use the new DSDT
version by removing where we patched it to use traditional.

OXT-259

### DIFF
--- a/recipes-extended/xen/files/acpi-hvmloader-pm-xt.patch
+++ b/recipes-extended/xen/files/acpi-hvmloader-pm-xt.patch
@@ -69,7 +69,7 @@ index 13af0ed..4c903b0 100644
  
 -        /* Future patches will extend AC object to better account for
 -         * AC to DC transition and more. */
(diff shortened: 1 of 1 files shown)


rianquinn:
Looks fine to me, but my vote doesn't count as I have no idea what half of this is

jean-edouard:
I got an explanation of this from Ross, and it all looks good and makes sense to me.
I will merge in a short while if nobody objects.

ktemkin:
+1, LGTM.

Just tested the basics (power on/off, sleep/resume) and haven't noticed any regressions.

PR #93, opened 2015-04-19
Qemu nic link state

eric-ch:
Have the emulated NICs mirror the link-status of the network back-end, exposed through XenStore.

Add xenstore_remove_watch() in addition to the existing XenStore helpers.

I had to refresh patches 0024/0025 and took the opportunity to remove trailing white-spaces and tabs that went in. Also, there was a dependency on syslog in the patch, I choose to remove it since we make syslog logging optional by a configure flag.

OXT-254

### DIFF
--- a/recipes-openxt/qemu-dm/files/0001-generic-xenstore-extensions.patch
+++ b/recipes-openxt/qemu-dm/files/0001-generic-xenstore-extensions.patch
@@ -56,18 +56,20 @@ Chris Patterson, <pattersonc@ainfosec.com>, 03/31/2015
 
 --- a/hw/xen_backend.c
 +++ b/hw/xen_backend.c
-@@ -658,6 +658,10 @@ static void xenstore_update(void *unused
+@@ -658,6 +658,12 @@ static void xenstore_update(void *unused
      if (sscanf(vec[XS_WATCH_TOKEN], "fe:%" PRIxPTR, &ptr) == 1) {
          xenstore_update_fe(vec[XS_WATCH_PATH], (void*)ptr);
      }
 +    if (sscanf(vec[XS_WATCH_TOKEN], "cb:%" PRIxPTR ":%" PRIxPTR,
 +               &ptr, &ops) == 2) {
-+        ((xenstore_watch_cb_t)ptr)((void *)ops);
++        if (ptr) {
++            ((xenstore_watch_cb_t)ptr)((void *)ops);
++        }
 +    }
  
  cleanup:
      free(vec);
-@@ -685,7 +689,11 @@ static void xen_be_evtchn_event(void *op
+@@ -685,7 +691,11 @@ static void xen_be_evtchn_event(void *op
  
  int xen_be_init(void)
  {
@@ -80,7 +82,7 @@ Chris Patterson, <pattersonc@ainfosec.com>, 03/31/2015
      if (!xenstore) {
          xen_be_printf(NULL, 0, "can't connect to xenstored\n");
          return -1;
-@@ -785,3 +793,88 @@ void xen_be_printf(struct XenDevice *xen
+@@ -785,3 +795,88 @@ void xen_be_printf(struct XenDevice *xen
      }
      qemu_log_flush();
  }
--- a/recipes-openxt/qemu-dm/files/0001-generic-xenstore-extensions.patch
+++ b/recipes-openxt/qemu-dm/files/0001-generic-xenstore-extensions.patch
@@ -19,6 +19,9 @@ Adds the following functions:
 - xenstore_add_watch(): provides ability to register a xenstore watch callback 
   with an optional pointer.
 
+- xenstore_remove_watch(): remove a xenstore watch callback previously
+  registered with xenstore_add_watch().
+
 --------------------------------------------------------------------------------
 UPSTREAM PLAN
 --------------------------------------------------------------------------------
@@ -35,6 +38,9 @@ None.
 CHANGELOG
 --------------------------------------------------------------------------------
 
+Eric Chanudet, <chanudete@ainfosec.com>, 04/18/2015
+- Added xenstore_remove_watch()
+
 Chris Patterson, <pattersonc@ainfosec.com>, 02/15/2015
 - Added xenstore_add_watch()
 
@@ -48,11 +54,9 @@ Chris Patterson, <pattersonc@ainfosec.com>, 03/31/2015
  hw/xen_backend.h |    4 ++++
  2 files changed, 22 insertions(+)
 
-Index: qemu-1.4.0/hw/xen_backend.c
-===================================================================
---- qemu-1.4.0.orig/hw/xen_backend.c	2013-02-15 18:05:35.000000000 -0500
-+++ qemu-1.4.0/hw/xen_backend.c	2015-03-30 17:37:35.720427701 -0400
-@@ -658,6 +658,10 @@
+--- a/hw/xen_backend.c
++++ b/hw/xen_backend.c
+@@ -658,6 +658,10 @@ static void xenstore_update(void *unused
      if (sscanf(vec[XS_WATCH_TOKEN], "fe:%" PRIxPTR, &ptr) == 1) {
          xenstore_update_fe(vec[XS_WATCH_PATH], (void*)ptr);
      }
@@ -63,7 +67,7 @@ Index: qemu-1.4.0/hw/xen_backend.c
  
  cleanup:
      free(vec);
-@@ -685,7 +689,11 @@
+@@ -685,7 +689,11 @@ static void xen_be_evtchn_event(void *op
  
  int xen_be_init(void)
  {
@@ -76,7 +80,7 @@ Index: qemu-1.4.0/hw/xen_backend.c
      if (!xenstore) {
          xen_be_printf(NULL, 0, "can't connect to xenstored\n");
          return -1;
-@@ -785,3 +793,62 @@
+@@ -785,3 +793,88 @@ void xen_be_printf(struct XenDevice *xen
      }
      qemu_log_flush();
  }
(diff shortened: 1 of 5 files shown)


recipes-openxt/qemu-dm/files/0001-generic-xenstore-extensions.patch
```diff
@@ -123,6 +127,32 @@ Index: qemu-1.4.0/hw/xen_backend.c
 +}
 +
 +/**
++ * Un-register an existing xenstore node using the specified callback and opaque pointer.
++ * @returns 0 on success, -1 otherwise.
++ */
++int xenstore_remove_watch(const char *base, const char *node,
++                          xenstore_watch_cb_t cb, void *opaque)
++{
```
rianquinn: You should validate that the pointers being provided are not NULL.

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
```diff
@@ -0,0 +1,246 @@
...
+@@ -589,6 +590,143 @@ static int xenstore_scan(const char *typ
+     return 0;
+ }
+ 
++/*
++ * OpenXT: Get type and id from \([a-zA-Z]\+\)\([0-9]\+\).
++ */
++static void xenstore_nic_parse_name(const char *name, char *type, char *id)
++{
++    if (name) {
```
rianquinn: You check for name, but forgot to check for type and id == NULL

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
```diff
@@ -0,0 +1,246 @@
...
++        while (isalpha(*name)) {
```
rianquinn: This code also looks like it assumes that the buffers are the same length, which could cause a buffer overflow. I would add a check for the length of each string.

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
```diff
@@ -0,0 +1,246 @@
...
++            *(type++) = *(name++);
++        }
++        while (isdigit(*name)) {
++            *(id++) = *(name++);
++        }
++    }
++    *type = '\0';
```
rianquinn: Same thing here, this assumes your not overflowing

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
```diff
@@ -0,0 +1,246 @@
...
++    *id = '\0';
++}
++
++/*
++ * OpenXT: Get network backend type and ID from NIC information table.
++ */
++static int xenstore_net_client_state_fetch_backend_info(NetClientState *ncs, char *type, char *id)
++{
++    size_t i;
```
rianquinn: pointer checks

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
```diff
@@ -0,0 +1,246 @@
...
++}
++
++/* Define constants for network devices XenStore data representation. */
++#define XENSTORE_NET_TYPE_LEN   5   /* Currently "vif"|"vwif" */
++#define XENSTORE_NET_ID_LEN     5
++/* OpenXT: Read the appropriate Base Register and check if we have to change
++ * the device status. */
++static void xenstore_update_nic(void *opaque)
++{
++    NetClientState *nc = opaque;
```
rianquinn: pointer checks

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
```diff
@@ -0,0 +1,246 @@
...
++        fprintf(stderr, "%s (%s%s): link status is now %s.\n", nc->name, type, id, !!val ? "down" : "up");
++        /* Notify the emulation through QMP.
++         * Note that qmp_set_link boolean is "link-up?", not nc->link_down "link-down?". */
++        qmp_set_link(nc->name, !val, NULL);
++    }
++}
++
++/* OpenXT: Register a Net Client in Xenstore. */
++int xenstore_register_nic(NetClientState *nc)
++{
```
rianquinn: pointer check

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
```diff
@@ -0,0 +1,246 @@
...
++
++    xenstore_update_nic(nc);
++
++    return 0;
++}
++
++/* OpenXT: Unregister a Net Client in Xenstore.
++ * Called when a device is removed and no longer used. */
++int xenstore_unregister_nic(NetClientState *nc)
++{
```
rianquinn: pointer check

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
```diff
@@ -0,0 +1,246 @@
...
++        return -1;
++    }
++    snprintf(base, sizeof (base), "%s/device/%s/%s", dompath, type, id);
++    free(dompath);
++
++    if (xenstore_remove_watch(base, "disconnect", xenstore_update_nic, id)) {
++        fprintf(stderr, "Could not remove xenstore watch on path `%s/disconnect'.\n", base);
++        return -1;
++    }
++    /* TODO: xenstore_update_nic() and force disconnect if not already? */
```
rianquinn: Any reason this was left as a TODO?

rianquinn:
In general this patch looks pretty good, and makes a hell of a lot more sense than what was there before. If nobody beats me to the punch, I will try to test this out on Monday if possible.

recipes-openxt/qemu-dm/files/0001-generic-xenstore-extensions.patch
eric-ch: Other XenStore helpers did not, so I assumed pointer validation was up to the caller. It made some sense to me since, in the current use-case, arguments are temporary static buffers allocated on the stack and passing cb/opaque as NULL is "safe" for add/remove helpers.
Adding a check in xenstore_update sounds like a good idea though, since the callback and opaque pointers are sscanf from the node without check.

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
eric-ch: Output arguments should be validated by the caller. But that's only my opinion.

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
eric-ch: Output arguments, although we could add length argument if you think that is required.

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
eric-ch: Do you mean checking against the macros for type/id length for type and id?

recipes-openxt/qemu-dm/files/0001-generic-xenstore-extensions.patch
rianquinn: agreed

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
eric-ch: This check would be redundant, maybe an assert for debug builds ?

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
eric-ch: Actually, I am more afraid of the xenstore_update() function parsing values from XenStore node and using them as pointer without check.
But again, debug assert?

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
rianquinn: for example, if type[2] and name[256] = {'a'}, this code could overflow. The while statement should be while(isalpha(*name) && i < strlen(type)) or something like that to validate that your not overflowing type

recipes-openxt/qemu-dm/files/0027-nic-link-state-propagation.patch
eric-ch: No I forgot to remove it.
It is not necessary as this is called from net/net.c qemu_del_nic() (link-state doesn't matter at that point). Even if we wanted to "disconnect" using that helper it would only be from the XenStore.

rossphilipson:
I don't really agree with so many pointer checks on arguments to internal functions. Passing bad arguments to internal functions is a development time screw up not a run time one. I like Eric's idea for asserts though.

rianquinn:
So long as the function is never used externally, I agree that the redundancy is optional as the functions that are being used externally can do the check and then from that point on, you have some degree of validation. Note that this assumes that internal functions can only be called after proper validation has occurred at least once (something that is not always easy to prove). That being said, I don't agree with the notion that developer screw-ups don't occur in run-time.

rossphilipson:
"That being said, I don't agree with the notion that developer screw-ups don't occur in run-time." ... now wait a minute, I never said that :)

But putting argument checks all over your internal code is not the way to go. You should write the code correctly and that is what assert helps you do.

rianquinn:
:)

Agreed with the use of asserts. To be honest, I prefer them (as you might have noticed :), as the code is a lot smaller. If you don't use asserts, you end up with if(blah) {printf("hell is upon us\n"); return 0xdeadbeaf; } over and over and over. It just gets depressing after a while.

cjp256:
Patches LGTM, +1.  However, sign-offs are missing in the commits.

eric-ch:
Added sign-offs.
Added missing assert in unregister_nic helper.

jean-edouard:
Started testing, looks good for now, will merge soon if all goes well.

jean-edouard:
All good, merging.

PR #94, opened 2015-04-21
qemu-dm: add patch to allow readonly IDE

cjp256:
Upstream qemu does not support readonly IDE devices
other than CDROM, because they do not exist.

This patch will mimic the behavior used in legacy-ioemu
to provide read-only IDE disk support, by erroring out
on write commands.  Typically, these disks will be unplugged
in favor of the PV couterpart.

OXT-264

### DIFF
--- a/recipes-openxt/qemu-dm/files/0001-readonly-ide.patch
+++ b/recipes-openxt/qemu-dm/files/0001-readonly-ide.patch
@@ -1,0 +1,152 @@
+--------------------------------------------------------------------------------
+SHORT DESCRIPTION:
+--------------------------------------------------------------------------------
+
+Provide readonly capability for IDE disk drives.
+
+--------------------------------------------------------------------------------
+LONG DESCRIPTION:
+--------------------------------------------------------------------------------
+
+Abort on IDE write commands (in a manner similarly done in legacy-ioemu),
+effectively providing a read-only IDE disk device for OpenXT.
+
+Upstream Qemu does not support this, but IDE disk devices must be used
+as Xen/Qemu will only unplug IDE disk devices in favor of PV (blkfront)
+usage within the guest.  Other options such as SCSI do not allow for this unplug
+without more intrusive changes.
+
+Can be toggled in ./configure stage (disabled by default):
+* To disable: --disable-readonly-ide
+* To enable: --enable-readonly-ide
+
+--------------------------------------------------------------------------------
+UPSTREAM PLAN
+--------------------------------------------------------------------------------
+
+Unlikely to upstream, qemu purposely does not support read-only IDE disk drives.
+
+--------------------------------------------------------------------------------
+DEPENDENCIES
+--------------------------------------------------------------------------------
+
+None.
+
+--------------------------------------------------------------------------------
+CHANGELOG
+--------------------------------------------------------------------------------
+
+Chris Patterson, <pattersonc@ainfosec.com>, 04/21/2015
+- Initial commit.
+
+diff --git a/blockdev.c b/blockdev.c
+index 63e6f1e..f6f6ab1 100644
+--- a/blockdev.c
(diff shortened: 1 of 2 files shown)


recipes-openxt/qemu-dm/files/0001-readonly-ide.patch
```diff
@@ -0,0 +1,154 @@
...
+index 3743dc3..73ae4c0 100644
+--- a/hw/ide/core.c
++++ b/hw/ide/core.c
+@@ -1188,6 +1188,12 @@ void ide_exec_cmd(IDEBus *bus, uint32_t val)
+         if (!s->bs) {
+             goto abort_cmd;
+         }
++#ifdef CONFIG_READONLY_IDE
++        if (bdrv_is_read_only(s->bs)) {
++            printf("IDE write attempted for read-only device\n");
```
rianquinn: My experience is that the printf will make it into /var/log/messages for dom0. In your testing, how often is this being triggered? Seems to me that this could end up causing QEMU to really slow down by leaving the printf in there.

recipes-openxt/qemu-dm/qemu-dm.inc
```diff
@@ -22,6 +22,7 @@ SRC_URI = "http://wiki.qemu-project.org/download/qemu-${PV}.tar.bz2"
 
 # QEMU Patch Queue
 SRC_URI += "file://0001-generic-xenstore-extensions.patch \
+            file://0001-readonly-ide.patch;striplevel=1 \
```
rianquinn: Probably should rename this to 0002

recipes-openxt/qemu-dm/files/0001-readonly-ide.patch
cjp256: Should be very rarely.  If you set your disk to readonly, and within your guest you treat it as such, then you'll get little or no prints.  But I can comment them out, if desired.

rianquinn: I guess it makes sense to comment them out just in case the user accidentally turns read-only on but doesn't setup the guest that way.

recipes-openxt/qemu-dm/qemu-dm.inc
cjp256: It has no dependencies, and can safely apply in parallel at the top.  Ordering is implied by the list so the numbers should simply be used to indicate the dependency tree, IMO.

jean-edouard:
I think readonly_ide should default to "no".

rianquinn:
Since we will always have --enable-readonly-ide set, and it's our patch (i.e. this is not being upstreamed) I would argue that it should default to --enable-readonly-ide.

cjp256:
Fixed default to no, commented out prints.

rianquinn:
+1

jean-edouard:
LGTM, will merge soon.

jean-edouard:
I installed OpenXT with patched versions of qemu-dm and dm-agent according to your pull requests.
I put SELinux in permissive mode by default and rebooted.
Then I created 2 VMs A and B, without tools nor stubdoms and halted both.
Then I added A's vhd to B's VM config as disk "2".
B then only boots when disk "2" is read/write ("mode": "w"). If I change the mode to "r", the VM spends some time in the "Starting" state, and then dies with a generic error...

jean-edouard:
When following the previous scenario, Windows stays on its splash screen forever trying to write to the read-only disc.
However, there's a high probability that the same behavior happens with ioemu (not tested).
Also, without that patch, Windows just won't have access to the disc anyway (no SCSI support).

The previous should be documented and the corresponding ticket (OXT-264) should stay open to track that. (will comment)

Merging now, to have it included in the next nightly. Will revert if needed.

PR #96, opened 2015-04-29
Remove old ACPI PM patches from the Xen pq.

rossphilipson:
This clears the way to introduce the new ACPI PM patch that is coming
and takes out 6 messy patches.

OXT-261 OXT-217 OXT-263


rianquinn:
Other than the removal of a bunch of ACPI patches to Xen, can you explain the changes in the patches. Mostly it just looks like a mod to the patch file format itself

rossphilipson:
When a patch is removed from a queue it can have side effects on other patches further down in the queue that modify the same files. Unfortunately this is often overlooked when pulling patches out. If you pop back and push the xen pq you will see a lot of patches succeeding with fuzz and incorrect offsets due to this. Even in the new qemu-dm pq there is some fuzz. In general, I always track down the patches that are effected by my patch removals and refresh them (or hand fix them when necessary). That is why those other patch files are changed. I did not fix all the other fuzzy patches in the queue at this time because the work is not related to this particular task. The xen pq is a pretty big mess.

In addition, the changes are a result of using quilt with default settings on guilt patches. We have settled on using quilt because it is the OE default. Unfortunately that brings along some baggage like losing the function scope specifiers in the diff meta-data.

rianquinn:
Makes sense to me. That's what I figured. 

+1

rossphilipson:
Yea sorry the lecture; I am sure you know all that. It was meant as more of a public service announcement to everyone to be mindful of patch removals from pq's. I have seen the issue with patch removal w/o thought of effecting other patches in the pq often over the years. Fuzzy patches suck :)

rianquinn:
np :grin:

transpute:
This excellent comment about patch queues deserves a permanent home in the coding guidelines wiki, with an illustrative link to this PR.

eric-ch:
We should definitely define a configuration to use with quilt in order to reduce the amount of change in the diff meta-data and add that to the patch etiquette on the wiki.

I find the following configuration "close enough" to guilt, but that is my preference only :
QUILT_PATCH_OPTS="--unified"
QUILT_REFRESH_ARGS="--no-timestamps --backup -p ab --no-index"

+1 otherwise ;).

rossphilipson:
erich-ch: So that is the other thing I wanted this PR to illustrate. On the one hand, your idea of using a config is great to reduce the amount of meta info changes to patches and keep useful things like function scopes. On the other hand, not just using the defaults is a bigger barrier to entry for new developers and will lead to more PR rejections because people didn't get the format right. I think you should start a discussion on this on the mailing list :)

rossphilipson:
transpute: That is a good idea. There were several other comments like this you mad in PRs about coding guidelines wiki. I think you should open a ticket and collect your suggestions there. Everyone is pretty darn busy at the moment and I suspect no one is going to do that right now. If you put that stuffs in a ticket, at least it won't get lost.

rianquinn:
:+1: that idea

jean-edouard:
I believe this can be merged independently of the patch format discussion, as we've been accepting that kind of change in the past.
Shout if you disagree, I'll merge soon otherwise.

PR #97, opened 2015-04-29
Remove broken battery patch from the Qemu pq.

rossphilipson:
This clears the way to introduce the new ACPI PM patch that is coming
and takes out a broken patch.

OXT-262 OXT-217

Signed-off-by: Ross Philipson philipsonr@ainfosec.com


rianquinn:
+1

eric-ch:
+1

PR #98, opened 2015-04-30
OXT-271: fix firstboot.sh typo

cjp256:

PR #99, opened 2015-04-30
[OXT-272]+[OXT-271]

cjp256:
Depends on oxt-271, and as such, is included in this pull (although available separately).

### DIFF
--- a/recipes-openxt/xenclient/xenclient-dom0-tweaks.inc
+++ b/recipes-openxt/xenclient/xenclient-dom0-tweaks.inc
@@ -120,4 +120,4 @@ do_install () {
 		${D}/usr/bin/selinux-state
 }
 
-INC_PR="r2"
+INC_PR="r3"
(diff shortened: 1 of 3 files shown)


cjp256:
https://github.com/OpenXT/xenclient-oe/pull/98

cjp256:
I do not believe so, because of --ignore-existing (which is probably good in most cases, just not this one).

jean-edouard:
Will cp -p preserve xattrs? The manpage wasn't clear on that...

cjp256:
Good Q.  I'd assumed the xattrs would be set correctly at file creation, being that it is copied into the right path? Perhaps I'm wrong, I'll double check it, but I didn't have selinux issues when I did before.

cjp256:
Tested it, labels look consistent between the dom0 image and the target location.

jean-edouard:
Great thanks, merging now.

PR #100, opened 2015-05-05
[qemu-dm] add a compile-time option for stubdomains

jean-edouard:
The code comes directly from https://github.com/OpenXT/xenclient-oe/pull/80.

Signed-off-by: Jed Lejosne <lejosnej@ainfosec.com>


rianquinn:
The patch should probably be applied to the beginning, not the end of the patch queue. This way, previous patches can also leverage it if needed. Other than that, it looks good.

jean-edouard:
I agree, I thought about it but didn't want to rebase the other patches on top...
I guess I'll do it then.

rianquinn:
:smile:

jean-edouard:
Here you go, enjoy!

brendank310:
LGTM, thanks for separating it out.  I'll build it, but my build tree is clean so it's a full rebuild for me. I think it's an innocuous enough change to just merge in.

rossphilipson:
I looked it over too. It is ugly but that is the cost of moving it to the bottom. Merging.

PR #102, opened 2015-05-07
Allow custom source cert filenames

pearsonk:
The previous recipe requires the source filenames to be
"prod-cacert.pem" and "dev-cacert.pem" otherwise do_install would fail
to find what it was expecting. This patch modifies do_install to use the
basename from the configuration file for these certificates and put them
into the package in their standardized format without requiring an
intermediate standard that was also different.


pearsonk:
Built recipe with both the old required filenames and different filenames. It helps to manually run up to the do_unpack step to see the files in ${WORKDIR} as the local fetcher grabs them and puts them into place, then run do_install (or finish the recipe tasks)

jean-edouard:
+1

rossphilipson:
Ok then, merging...

PR #103, opened 2015-05-07
Update Installer Dom0 RAM to 8GB

pearsonk:
Some systems were having issues with 64GB (one person cited issues with
50GB). 8GB is common for installed systems and is beyond sufficient for
the installer with plenty of RAM available for any debugging or
unpacking for derivative systems.


pearsonk:
Just some speculation on causes, the problem seen by others may stem from RAM balance and the fact that even the installer sets dom0_max_vcpus=1. My test systems had >64GB per socket available. If this is the case, a crash could still happen on a multi-socket system with <8GB per socket but why would anybody ever do that?

pearsonk:
Tested this on a Z820 with 96GB RAM and for funsies a laptop with 16GB RAM.

jean-edouard:
+1

rossphilipson:
Checked that no "----" snuck in :)

Merging...

PR #104, opened 2015-05-08
Use mac address to determine proper eth <-> vif pairing.

crogers1:
Use the xenstore to query for a list of vifs available to the stubdom
and match with the proper eth device.

OXT-275

### DIFF
--- a/recipes-openxt/qemu-dm/qemu-dm-stubdom/qemu-ifup-stubdom
+++ b/recipes-openxt/qemu-dm/qemu-dm-stubdom/qemu-ifup-stubdom
@@ -41,7 +41,22 @@ else
     echo 1 > /sys/class/net/$bridgename/bridge/break_8021d
 fi
 
-ethname="eth${ifnumber}"
+#Query xenstore for mac address of vif and match with mac found in
+#sys/class/net
+domid=$(xenstore-read domid)
+xenstore_mac=$(xenstore-read /local/domain/${domid}/device/vif/${ifnumber}/mac)
+
+for netif in $(ls /sys/class/net/); do
+    if [ "${xenstore_mac}" == "$(cat /sys/class/net/${netif}/address)" ]; then
+        ethname="${netif}"
+        break
+    fi
+done
+
+if [ -z ${ethname} ]; then
+    echo "For mac address: ${xenstore_mac} from interface: ${1} from xenstore, could not match with any known network interface."
+    exit 1
+fi
 
 if [ ! -e "/sys/class/net/$ethname" ]; then
     echo "$ethname does not exist, strange (invoked for $1)"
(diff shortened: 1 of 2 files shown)


pearsonk:
Personal nitpick, but the function is in the middle of the procedural part of the script.

crogers1:
I can move that.

cjp256:
1) I'd recommend using the form $(...) instead of `...`
2) Relying on the ordering of xenstore-list seems like it may be error prone. Perhaps scan for the mac address instead?

crogers1:
Updated.  Old version of sh doesn't support $() command expansion.  Now pairing by parsing for the mac address.

cjp256:
If it's only going to be called once, I'd skip making it a function and just put in the code.

While grepping ifconfig output works, it may be nicer to find the match it in a cleaner fashion, e.g.:
`for netif in $(ls /sys/class/net/); do `
`if [[ "${xenstore_mac}" == "$(cat /sys/class/net/${netif}/address)" ]]; then ethname="${netif}"; break; fi`
`done`

pearsonk:
Agree with @cjp256 as the current code doesn't really stand alone or appear to be targeted for reuse since it writes a single value to a pre-determined variable as opposed to stdout to be captured by the caller. Simplicity should probably win out here.

crogers1:
Updated.

pearsonk:
The script is missing some logic in case the MAC/ethname is somehow not matched. $ethname would never be set, being an empty string. This would cause the ' if [ ! -e "/sys/class/net/$ethname" ]' to resolve false and the else portion to cause a massive amount of puking. I don't know likelihood of occurrence, but it could probably happen if the xenstore got a bad value.

crogers1:
We could set ethname to something sane before the loop:
```bash
xenstore_mac=$(xenstore-read /local/domain/${domid}/device/vif/${ifnumber}/mac)
ethname= #something
for netif in ...
```bash
If there are no matches, we can safely error out, and if we match, then it will be overwritten with the correct value.

crogers1:
Eh, let's just check for zero length for $ethname and exit if it fails to match.

pearsonk:
I think the error message for a MAC not found should include the MAC being searched. After that, LGTM.

crogers1:
Updated with more descriptive error for mac not found case.

pearsonk:
Point of clarity, should also include the ${xenstore_mac} that was being searched. That way if this error makes its way to xenconsoled, then someone debugging can have a better idea what to look into.

crogers1:
I agree.  Error message now reports ${xenstore_mac} and value passed to qemu-ifup script (tapX.N).

cjp256:
+1, seems to be working well for me. LGTM

PR #105, opened 2015-05-12
[qemu-dm] try take the vga lock (or do nothing) when drawing graphics

jean-edouard:
Modify and use the vga (spin)lock so that we don't draw graphics (and potentially resize) while the vgabios is working (most likely handling a modeset).

OXT-123


rossphilipson:
+1 but I would like some other folks to take a look too.

jean-edouard:
Added a commit to take the lock earlier and increment the PR, as per @ktemkin recommendations.

cjp256:
Will test latest.

cjp256:
+1, I don't have repro of the test case, but VMs seem to be booting fine with these patches.

PR #106, opened 2015-05-12
firstboot: remove chown for tss:tss on PS_DIR

cjp256:
It really needs to be only on PS_FILE, PS_DIR is too broad. It also
appears to be unnecessary for two reasons:

1) this needs to be done in time for init.root-ro, which this is not.
2) and init.root-ro looks like it replaces every boot (if it is non-zero
sized) anyways.

In fact, as a result of 2, the firstboot install of the trousers files
can probably be safely removed altogether.

OXT-272


jean-edouard:
ACK, merging.

PR #107, opened 2015-05-13
[qemu-dm] add max-ram-below-4g parameter.

eric-ch:
Add max-ram-below-4g parameter to machine option. This option is only
implemented for xenfv machine using i440fx PCI bus controller. Emulation
will use the value to set up the PCI hole base at a lower address,
making it bigger.

This parameter makes the PCI hole consistent between the Xen memory
listener and the i440fx emulation. Also, hvmloader will process that
value again after QEMU is initialized, both have to match.

The full fix for OXT-243 includes the toolstack commit to process that value before starting QEMU and the modification in dm-agent to pass it from Xenstore to QEMU command line.

### DIFF
--- a/recipes-openxt/qemu-dm/files/0030-xenfv-i440fx-max-ram-below-4g.patch
+++ b/recipes-openxt/qemu-dm/files/0030-xenfv-i440fx-max-ram-below-4g.patch
@@ -1,3 +1,63 @@
+################################################################################
+SHORT DESCRIPTION: 
+################################################################################
+Add machine option "max-ram-below-4g" to configure the RAM limit below 32b
+address boundary.
+
+################################################################################
+LONG DESCRIPTION: 
+################################################################################
+QEMU configures the guest memory layout using assumptions that will not always
+match what hvmloader sets up later, leaving the guest with an e820 not matching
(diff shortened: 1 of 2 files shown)


recipes-openxt/qemu-dm/files/0030-xenfv-i440fx-max-ram-below-4g.patch
```diff
@@ -0,0 +1,187 @@
+--- a/hw/pc.h
```
rianquinn: Missing the QEMU patch header

rianquinn:
+1 the code, I have not tested though

recipes-openxt/qemu-dm/files/0030-xenfv-i440fx-max-ram-below-4g.patch
eric-ch: Oops, fixed.

jean-edouard:
+1
I tested this on a machine that reproes the issue.
With the fix, my 32 and 64 bits VMs booted fine with and without an AMD gpu pt-ed to them, with 3.5GB and 4GB of ram. (all 8 tests passed)

PR #108, opened 2015-05-15
[xen] fix leaks and formatting in v4v firewalls

jean-edouard:


cjp256:
LGTM

cjp256:
Noted that this is missing signoff (and ticket?).  Any test steps?

jean-edouard:
No ticket nor repro steps, this was found by looking at the code.
A possible issue would be that VMs can't shutdown after getting an XSM/viptables denial, as the denials would increment (and not decrement) the refcount on the domain. Also a big amount of denials could lead to a visible memory leak.

cjp256:
Testing looks good to me.

PR #111, opened 2015-05-19
[linux][blkback] fix memory leaks by backporting the code from Linux …

jean-edouard:
…3.18.13

OXT-289


cjp256:
Have test steps to validate that memory is not leaking between reboots? (Without rebooting until something crashes? :))

Building now for testing..

jean-edouard:
I did 900 reboots of a Ubuntu VM (no stubdom, no tools) on a laptop, and now 106 (and counting) reboots of a Windows 7 64 bits VM (stubdom + tools) on the same laptop, without ever rebooting the platform. No leak, 236MB of RAM used at the moment.

rossphilipson:
Reviewed the patch. It is rather complicated so I don't understand all of it but the bits that changed LGTM. There is a new "feature" that got added - a xenstore flag to disable discard IO. It looks like it does not change the behavior though - if the flag is not in xenstore it behaves as before.

cjp256:
LGTM and testing is going well.  +1, merging

PR #112, opened 2015-05-20
Dedicate a fixed amount of memory to dom0

eric-ch:
Follow "Xen best practice"
(http://wiki.xenproject.org/wiki/Xen_Project_Best_Practices) and make
sure to give dom0 a fixed amount of memory.

Having dom0_mem=320M leaves the possibility to dom0 to balloon up to 4G.

OXT-285

### DIFF
--- a/recipes-openxt/xenclient/xenclient-dom0-tweaks/grub.cfg
+++ b/recipes-openxt/xenclient/xenclient-dom0-tweaks/grub.cfg
@@ -31,7 +31,7 @@ fi
 # Set to drm-graphics to enable experimental DRM plugin support on pre-Haswell systems
 DRM_GRAPHICS_OPTION=""
 
-XEN_COMMON_CMD="console=com1 dom0_mem=320M com1=115200,8n1,pci mbi-video vga=current flask_enforcing=1 loglvl=debug guest_loglvl=debug"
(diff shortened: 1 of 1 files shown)


pearsonk:
LGTM

jean-edouard:
For completeness sake, and because the code in Xen looks a bit flaky, I think we should also set "<amt>".
Something like "dom0_mem=320M,min=320M,max=320M".

This will also limit the number of guests we can run. We should figure out what that number is (with and without stubdoms), and document it somewhere in the wiki.

eric-ch:
I do not see why you want quotes (also what part of parse_dom0_mem() look flaky?), neither why 320M should be added. Both the code and the syntax min:/max: makes it obvious imho.

As suggested in the associated ticket, this might as well be configured during OpenXT installation. It is a topic for another ticket though.

jean-edouard:
Sorry, ignore the quotes.

parse_dom0_mem() looks fine, compute_dom0_nr_pages() is the one I don't like. It starts with undefined values and does weird maths...

If I read the code right, the computed initial value should be the one you specify for min and max, but it would just make it clear and obvious to have it on the cmdline IMHO.

eric-ch:
Re-commit to add the value after min: and max:.

jean-edouard:
Good stuffs :)
LGTM, will merge soon.

PR #114, opened 2015-05-21
Fix segv in NIC link state propagation

eric-ch:
Use NICInfo table instead of NetClientState for watch argument.

Refresh patches applied on top.

OXT-291


cjp256:
+1, build & tested - looks good

jean-edouard:
same here, merging.

PR #115, opened 2015-05-21
format and cleanup

eric-ch:
Format 0030-xenfv-i440fx-max-ram-below-4g.patch and remove debug print.


jean-edouard:
+1, thanks :)

PR #116, opened 2015-05-22
Fix build of Icedtea on newer linux kernels.

ktemkin:
Icedtea has a hardcoded check for the kernel version prefix. We very
strongly suspect our host kernels are new enough to build icedtea,
so we'll disable the check (rather than having to change a constant
for every version).


eric-ch:
+1, otherwise icedtea won't build on 4* kernels.
There is already an OE patch for kernels 3*, so disabling the check sounds good.

PR #119, opened 2015-05-30
[acpi] QEMU portion of the ACPI PM feature.

rossphilipson:
OXT-260


jean-edouard:
LGTM and tested, will merge after 5PM today if there's no objection.

PR #120, opened 2015-06-05
[acpi] Use not present mode to allow sleep/power buttons to work on d…

rossphilipson:
…esktops.

The code was just totally disabling the ACPI PM feature when AC and batteries
were not present. But non-portable systems still have power and sleep
buttons.

OXT-304


jean-edouard:
Built here:
http://192.99.200.146:81/builds/master/ext-dev-201-master/iso/installer.iso

jean-edouard:
Tested on a desktop, writing sleep/power button nodes trigger the right behavior in the (tools-less) guest.
The guest also doesn't see any battery, as expected.
+1

jean-edouard:
Will merge soon.

PR #121, opened 2015-06-08
Push configuration data into recipes

dpquigl2:
Several recipes worth of CFLAGS LDFLAGS and ASNEEDED options were specified in local.conf. This pushes those statements into the recipes themselves

OXT-308

Signed-Off-By: David Quigley <drquigl@tycho.nsa.gov>

### DIFF
--- a/recipes-extended/pigz/pigz_2.2.4.bbappend
+++ b/recipes-extended/pigz/pigz_2.2.4.bbappend
@@ -1,2 +1,0 @@
-SRC_URI = "${DEBIAN_MIRROR}/main/p/pigz/pigz_${PV}.orig.tar.gz \
-           file://ldflags.patch"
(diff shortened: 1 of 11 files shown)


recipes-extended/pigz/pigz_2.2.4.bbappend
```diff
@@ -1,2 +0,0 @@
-SRC_URI = "${DEBIAN_MIRROR}/main/p/pigz/pigz_${PV}.orig.tar.gz \
-           file://ldflags.patch"
```
cjp256: Is this patch still being applied somewhere?  Should this be a +=?

cjp256:
Missing commit signoffs.  And just the Q about the pigz ldflags patch, but otherwise looks good to me.

recipes-extended/pigz/pigz_2.2.4.bbappend
dpquigl2: I don't think it is. Its not clear to me why it needed at all. It builds just find an works without either the modified source package or the patch.

rossphilipson:
See my comment in https://github.com/OpenXT/openxt/pull/61 about PRs being left open for long periods...

dpquigl2:
Bring picked up by other people to finish.

PR #123, opened 2015-06-09
[OXT-309] qemu: maintain guest time offset

cjp256:


recipes-openxt/qemu-dm/files/0001-generic-xenstore-extensions.patch
```diff
@@ -54,9 +54,11 @@ Chris Patterson, <pattersonc@ainfosec.com>, 03/31/2015
  hw/xen_backend.h |    4 ++++
```
rianquinn: Missing header :smile:

cjp256: No it's not. Go look at the patch.

jean-edouard:
Not a blocker here, since you're not really modifying its content, but `0024-Stubdom-support-for-ISO-media-changes-after-boot.patch` really ought to be ported to the new CONFIG_STUBDOM way of building qemu-dm-stubdom and moved to qemu-dm.inc...

cjp256:
Agreed, however, that is well outside the scope of this ticket/PR :)

jean-edouard:
I think you shouldn't update it and let the build break on it :) jk

recipes-openxt/qemu-dm/files/0001-generic-xenstore-extensions.patch
rianquinn: Rats.... thought I got you there. Better luck next time !!!

cjp256:
To test: start Windows guest.  Change timezone (to something non-UTC), note time displayed in Windows.  Reboot.  Time should be consistent after reboot.

Without this fix, time will revert to the system UTC time (e.g. off by timezone offset).  (Unless you applied registry change such as http://crashmag.net/configuring-windows-7-support-for-utc-bios-time)

rossphilipson:
Tested - works fine. +1

Merging.

PR #124, opened 2015-06-10
[iasl]: Add iasl recipe and rework xentools and seabios

dpquigl2:
Add the iasl recipe so we don't require it to be on the build box.
Then moodify recipes to require the iasl-native package instead of
relying on the host to have the program.

OXT-210

Signed-off-by: David Quigley <drquigley@tycho.nsa.gov>


rossphilipson:
Does xen-firmware.bb need the depend too? That is the recipe that actually builds the ACPI bits in hvmloader.

dpquigl2:
It's not clear that it doesn't pick up the dependency from somewhere else. The build completed fine with just those two modifications. I can add it in though if you'd like though.

rossphilipson:
LGTM now. Merging.

PR #125, opened 2015-06-10
[xcpmd] remove dependency on libxenacpi

jean-edouard:
See https://github.com/OpenXT/xctools/pull/9


rossphilipson:
This goes with https://github.com/OpenXT/xctools/pull/9

Merging too...

PR #126, opened 2015-06-11
Qemu support for acpi shutdown without pv drivers.

crogers1:
0033-acpi-pm-feature.patch:
  * Add xenstore watch on a per domain basis that can trigger
    a power button SCI to the guest.

OXT-212

### DIFF
--- a/recipes-openxt/qemu-dm/files/0033-acpi-pm-feature.patch
+++ b/recipes-openxt/qemu-dm/files/0033-acpi-pm-feature.patch
@@ -39,7 +39,7 @@ PATCHES
 Index: qemu-1.4.0/hw/Makefile.objs
 ===================================================================
 --- qemu-1.4.0.orig/hw/Makefile.objs	2013-02-15 18:05:35.000000000 -0500
-+++ qemu-1.4.0/hw/Makefile.objs	2015-06-05 09:22:08.461081630 -0400
++++ qemu-1.4.0/hw/Makefile.objs	2015-06-15 12:37:48.149121126 -0400
 @@ -195,6 +195,7 @@
  # xen backend driver support
  common-obj-$(CONFIG_XEN_BACKEND) += xen_backend.o xen_devconfig.o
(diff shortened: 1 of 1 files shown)


jean-edouard:
This should be added to 0033-acpi-pm-feature.patch instead of adding a new patch.

rossphilipson:
I agree with Jed, pop back to 0033-acpi-pm-feature.patch and add it there. We don't want to get into the horrible patch fragmentation state we were in before.

crogers1:
Will do.  Yea my C's a bit rusty, thanks for that Ross.

rossphilipson:
Oh and I know it is a PITA but if you could use quilt <pukes in mouth a bit> with default settings to update the patch, it would help keep things consistent too.

crogers1:
Updated with quilt.

rossphilipson:
Me likey +1 on fixes and format. Needs someone to test it. And the Haskell changes are O_o for me.

crogers1:
Updated PR.  Changed my mind, I don't want to share the same xenstore node as the PV drivers.  Watch a new node called "hvm-shutdown"

cjp256:
Did some basic testing, seems to work OK if Windows is running.

If you issue shutdown before Windows is up (e.g. BIOS is loading), then it indefinetely spins at "Stopping..." but the guest lives on.  Is there something else at play here, or could the signal be emitted again (when the guest is ready to receive it) while shutdown is executing?

rossphilipson:
Yea Windows has to be up and running to field the ACPI event (it is the OSPM). The event is actually going in the bit bucket since Windows has not even enabled the GPE for that event yet. You should be able to reissue the event later.

transpute:
Will the UIVM pass on a user-issued Shutdown event when the VM is already in "Stopping" mode?  I've seen VMs stuck in that mode and usually the only recourse is Force Shutdown.

rossphilipson:
Tested, looks like it is working. +1 merging.

PR #127, opened 2015-06-13
[pvusb] Add usbback support for URB cancelation.

rossphilipson:
The Linux usbfront driver must support an URB dequeue callback. When
invoked it is supposed to kill the URB in question if it can and complete
it. To do this an internal command is sent to usbback where it uses the anchor
list to find the URB and kill it if it is still being processed.

OXT-38


rossphilipson:
I know Eric is working on porting the PQ to 3.19 but I wanted to get this fix out there because it looks like  all this fixes isochronous devices (cameras, microphones, etc).

Eric, this patch can be merged into the unified usbback patch too.

jean-edouard:
LGTM, will merge soon.

PR #129, opened 2015-06-30
Remove the remaining currency expirations from the icedtea utility.

crogers1:


rossphilipson:
That looks fine to me. If someone else who is involved wants to +1 this, I will merge it.

ktemkin:
+1.

Justed tested this, LGTM. You might want to bump the PR on the relevant recipe so broken builds automatically fix themselves-- but otherwise looks good.

crogers1:
It doesn't look like we have the full recipe in our repos; there's just a bbappend, and that doesn't have a PR defined.

crogers1:
As per Kyle's suggestion, just append to the PR to indicate a new version.

jean-edouard:
LGTM, merging to fix the build

PR #132, opened 2015-07-10
Add VT-d workaround for IGD on Nehalem (Dell 980).

ktemkin:
This should correct graphics garbling on Dell 980 machines.

### DIFF
--- a/recipes-extended/xen/files/workaround-nehalem-igd-vtd.patch
+++ b/recipes-extended/xen/files/workaround-nehalem-igd-vtd.patch
@@ -1,0 +1,161 @@
+--------------------------------------------------------------------------------
+SHORT DESCRIPTION:
+--------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


transpute:
Could we rename is_xc_xt_ to is_oxt_?

ktemkin:
Name is updated. 

Also pushed up a minor change that had gotten lost in another tree, which skips this workaround if iommu=force is set on the Xen command line.

gamorg02:
Was able to confirm a Dell 980 now is able to boot and have GPU/NIC pass through with and without stubdom enabled.  Was also able to confirm on a 9010 (Ivy Bridge) there were no adverse effects when doing the same GPU/NIC passthroughs with and without stubdoms.

rossphilipson:
As long as this is restricted to this particular platform, I am fine with this. The specific device ID check on the host bridge device is sufficient. +1 merging.

PR #133, opened 2015-07-16
Allow Stubdomain configuration of a given VM through the toolstack

eric-ch:
Change kernel cmdline parsing in the dmagent case so new parameters do
not affect dmagent argument recovery.

OXT-332

See related PR in:
- manager.git
- toolstack.git
- idl.git

### DIFF
--- a/recipes-openxt/initrdscripts/initramfs-xenclient/xenclient-stubdomain/init.sh
+++ b/recipes-openxt/initrdscripts/initramfs-xenclient/xenclient-stubdomain/init.sh
@@ -59,10 +59,6 @@ echo "Command line: `cat /proc/cmdline`"
 
 ln -s /proc/self/fd/2 /dev/stderr
 
-QEMU_CMDLINE=`cat /proc/cmdline | cut -d' ' -f4- `
-
-DOMID=`echo $QEMU_CMDLINE | cut -d' ' -f2 `
-
 echo $*
 echo 1 > /proc/sys/net/ipv4/ip_forward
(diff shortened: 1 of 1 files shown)


jean-edouard:
I think this is our chance to get that TODO fixed :)
If we give the user the possibility to change the cmdline, I don't know how we'll enforce the fact that the domid has to be last...
I'm also concerned that the kernel could act on what's actually intended for qemu...
Something like "openxt_agent=dmagent openxt_domid=<domid>" would be more explicit, safer and easy to parse! :+1:

jean-edouard:
@eric-ch did you see my previous comment? Please let me know what you think about it.

eric-ch:
@jean-edouard Sorry I did not or forgot to answer. So the dmagent <domid> is appended to the cmdline of the stub-domain by the toolstack. I believe this is how we know, for now, that it's always at the end. I do agree though, it could be done more explicitely.
I think you are right, we might as well change the cmdline format now.

jean-edouard:
Commenting here about this PR along with
- https://github.com/OpenXT/manager/pull/15
- https://github.com/OpenXT/toolstack/pull/9
- https://github.com/OpenXT/idl/pull/5

I just built and tested it. The build and install went fine.
I created a VM, `xec-vm -n <vm> get stubdom-cmdline` gave me `init=/init xen_blkfront.max=8`, which looks good. I modified stubdom-cmdline, restarted the VM and verified it with `grep stubdom /var/log/messages | grep "Kernel command line" | tail -1`.

All good, +1.

jean-edouard:
Will merge soon, unless someone volunteers to give this more testing :)

PR #134, opened 2015-07-20
Remove the remaining currency expirations from the icedtea utility.

jean-edouard:
This is a cherry-pick of a master commit that's needed for legacy-ioemu to build.
If you think this should be done a different way, please shout.


rossphilipson:
This needs a bit of clarification. What Jed is asking is whether people think cherry picking is the appropriate way to get commits from one branch to another. E.g. if we have stable branches and want to pull in critical fixes from say master. If there is any discussion to be had, it really should happen on the mailing list.

As for this PR, +1, merging in a while.

pearsonk:
Well, another question could be was the legacy-ioemu intended to be maintained and if so who is responsible for doing that work? There are probably quite a few changes that should also be moved.

As for cherry-picking, seems like the only appropriate method to me. Especially for things like this which are solely intended to prevent the build from outright failing.

rossphilipson:
I don't think we are maintaining the ioemu branch other than little things like this to just make it build.

The question about cherry picking is a broader question that is more future looking. At some point we will have stable code lines (branches) and we need a mechanism to pull in important fixes etc.

PR #135, opened 2015-07-24
Update NDVM to have PCNET32

pearsonk:
Dom0/installer has the module, but this makes the NIC work with
NDVM.


pearsonk:
Tested with AT-2701-FX

cjp256:
+1, LGTM

PR #136, opened 2015-07-27
Upgrade acpid to version 2.0.23.

jctemkin:
As discussed in https://groups.google.com/forum/#!topic/openxt/uk4qVSYea2Y


rossphilipson:
This is looking good. I made a custom xcpmd and it is getting all the ACPI events expected. It is a little hard to validate that i definitely have the new acpid (no version info that I can find) but it looks like the new one installs an init.d script which is there.

I will merge this once the white-space formatting is fixed :)

jctemkin:
Great to hear. FYI, you can verify acpid's version using apid --version.

rossphilipson:
I tried --version but it did not work. But it does work if I disable selinux:
root@xenclient-dom0:~# acpid --version
acpid-2.0.23

PR #137, opened 2015-07-27
qemu: patches for XSA-138 / CVE-2015-5154

cjp256:
OXT-339


jean-edouard:
Built and tested.
All good, VM starts and can access both physical and emulated CD drives, with and without tools, with and without stubdoms.
Will merge soon.

PR #138, opened 2015-08-03
Upgrade Linux kernels to 3.18.

eric-ch:
This work is based on 3.18.16, further upgrades to follow the 3.18.y development will ensue.

Important changes include:
- switch to Linux in-tree e1000e module,
- Removal of xen-wmi-acpi interfaces (and removal of dependencies),
- Back-port Xen PV guest PAT support from Linux 3.19,
- Attempt at documenting the patch-queue.

See https://openxt.atlassian.net/browse/OXT-323 for the complete list of changes in various OpenXT repositories.


pearsonk:
Did a build and I guess I will use this to post tests run and results:

- verified installer kernel at 3.18.16
- Verified Dom0 kernel at 3.18.16
- Verified UIVM kernel at 3.18.16
- Verified NDVM kernel at 3.18.16

- Verified broadwell graphics (NUC5i5MYBE)

jean-edouard:
NDVM doesn't seem to have iptables NAT support, preventing any shared networking.

eric-ch:
@jean-edouard Oops, it is back now, should it stay as a module or do we require it built in the kernel?
Also fixed some typos and refreshed the defconfigs.

jean-edouard:
Technically, I guess only things that are not always used should be built as modules, so pretty much nothing but NIC drivers, which is the only variable for NDVM... But I think that's something we'd want to address separately from this PR, so LGTM! Thanks.

jean-edouard:
Sorry but now it misses masquerde support... (ipt_MASQUERADE.ko on 3.11)
3.11 also has ipt_ULOG.ko but I don't know what it does.
I checked to make sure that there's isn't any other module missing. The only other one I could find is pcnet32.ko, which seems to be some vmware module that we probably don't need.

cjp256:
@jean-edouard pcnet32 is required (just merged recently https://github.com/OpenXT/xenclient-oe/pull/135)

eric-ch:
@jean-edouard CONFIG_IP_NF_TARGET_ULOG (for ipt_ULOG.ko) is deprecated by CONFIG_NETFILTER_NETLINK_LOG. Added the rest.
@cjp256 pcnet32 added back.
Sorry :)

jean-edouard:
Thank you :)

rossphilipson:
Testing 249 now - so far so good...

jean-edouard:
- PCI passthrough of EHCI controllers, XHCI controllers and audio cards work perfectly

rossphilipson:
Windows 7 x64 Pro install worked fine. I am able to connect using wifi so the NATing issue seems fixed.

jean-edouard:
Don't we need to update linux-firmware too? I don't have any Broadwell laptop, but I bet iwlwifi won't be able to drive the wireless NIC without recent firmware stuffs

jean-edouard:
- GPU passthrough works perfectly

At this point, I'm ready to merge this PR and the associated ones.
Let's give it maybe just one more day of testing?

transpute:
Please create a branch (e.g. haswell) before merging,

jean-edouard:
If you wish to maintain a Linux 3.11 branch, please request one as described by Chris's RFC.
I'll hold the merge.

transpute:
I've responded on-list about the maintenance requirement, the proposed RFC is still under discussion.  

There are more considerations besides maintenance, the most obvious one being stability.  If stablity were not a consideration, the 3.18 kernel would already have been developed on master.  Given the effort involved in Eric's work and the effort to stabilize OpenXT after the last big merge (ioemu), and that it has been almost two years since a merge which adds support for new hardware, this merge needs to be clearly identified.

jean-edouard:
Sure, and this is the perfect use case for the under-discussion RFC.
Let's do it right, whatever we decide that is :)

eric-ch:
@transpute I am really reluctant at creating a branch for something that is not maintained, that looks like a tag really. I think tagging the repos with something like "last-linux-3.11" is more appropriate: it's just a point in source history; people with interest in maintaining a branch from that state can do so easily; it gives a named point in the sources history for comparison, regression testing, etc. Current build scripts might be misleading with the "BRANCH" variable that can take any revision.

Anyway that probably belongs in the RFC thread.

cjp256:
+1, i think tagging "last-linux-3.11" is an excellent way to snapshot the tree before this merge.

rossphilipson:
My initial test of Windows update ran into problems - it would just not run. But I tried it with other VMs and I cannot repro - Windows update works fine with and without disk encryption. And Eric cannot repro either so I think there is something wrong w/ that original VM and I will consider it a fluke.

rossphilipson:
+1 for the PR after a bunch of testing. It looks good.

rossphilipson:
I know discussions are taking place on:

https://groups.google.com/forum/#!topic/openxt/LlexgJ-sEtg

It does not look like there is any resolution in sight though. We have a manifest that we can do a pre-3.18 branch (or whatever) from. Can we consider going ahead and merging the 3.18 work - it is really blocking a lot of other work from progressing?

transpute:
How about tagging the repos before merging?

rossphilipson:
+1, I am fine with tagging them - we can branch off a tag later once we sort out branching.

eric-ch:
+1 (again) for tagging. The discussion can go on and branching will still be possible later from that tag if someone is interested in it.

jean-edouard:
Branch done, tag in progress, I will merge this pull request and the related ones after.

PR #140, opened 2015-08-11
Upgrade linux micro to 3.18.19

eric-ch:
Upgrade to include upstream fixes.

backport-config-Enable-NEED_DMA_MAP_STATE-by-default-when-SWIOTLB-is-selected.patch is now upstream.

PR #141, opened 2015-08-11
Upgrade linux-firmware

eric-ch:
Move to linux-firmware top revision.


rossphilipson:
See my comment here: https://github.com/OpenXT/xenclient-oe/pull/140

I think we should just get this in too.

jean-edouard:
+1 too.
This is a requirement for Broadwell support, heavily tested upstream and unpatchable anyway.

jean-edouard:
Will merge soon.

pearsonk:
I agree with @rossphilipson and @jean-edouard, makes sense to merge these ASAP to we can test it all together

PR #142, opened 2015-08-12
Changed the linux recipes' PR variable from 1 to r1 to follow OE conv…

brendank310:
…ention


jean-edouard:
Does that fix anything in particular or is it aesthetics?
+1 for both cases I guess :)

brendank310:
I was getting parse failures when using bbappends in another layer against the recipes with PR = "1", where PR = "r1" parsed fine.

jean-edouard:
Makes sense. @eric-ch no objection?

eric-ch:
@jean-edouard None :)
+1

PR #144, opened 2015-08-13
[linux] get rid of the dom0 over-patch-queue

jean-edouard:


eric-ch:
Linux 3.18 patch-queue is in recipes-kernel/linux/linux-xenclient-3.18 but not recipes-kernel/linux/files, pciback-restrictive-attr.patch has to be moved there.

Also, I strongly believe that putting patches in "files" (as it has been done in QEMU 1.4 during the layerisation) is bad practice: it makes it really hard to manage two different versions under the same subdir. Having that can be convenient for regression testing, the patch-queue being dependent on the version of the project it is applied upon.

rianquinn:
I agree. Files is probably not the best place for patches. At one point the QEMU patches were in their own folder specifically for versioning... I cannot remember why they were moved but we should probably look into moving them back at some point.

jean-edouard:
+1 for not using "files".
However I think we had this discussion in the past, and I think I remember "files" won.
I'll be glad to add a commit that moves files/* to linux-xenclient-3.18/ or something if we all agree...

eric-ch:
@jean-edouard files won? I do not remember that, but we could bring it back I guess. For now since linux-xenclient-3.18 already holds the patch-queue for the Linux 3.18 kernel, I guess we really do not want to mix both.

Also, I think you meant files/* to linux-xenclient-3.11 :). The current patch-queue for Linux 3.18 already being in linux-xenclient-3.18.

jean-edouard:
Oh, this PR is wrong then, I didn't realize...

PR #145, opened 2015-08-14
qemu-dm-1.4 Patch-queue subdir

eric-ch:
Regarding comment on https://github.com/OpenXT/xenclient-oe/pull/144, propose to make a version related patch-queue subdir for QEMU.


jean-edouard:
Yes! LGTM. Will merge soon.

PR #147, opened 2015-08-14
Added support for permitting xcpmd to send to dbus

Marsland:
Added support for xcpmd to send to dbusbouncer context to support adding
xcpmd to the UI

### DIFF
--- a/recipes-security/selinux/refpolicy-mcs/openxt_policy_modules_services_xenpmd.te.patch
+++ b/recipes-security/selinux/refpolicy-mcs/openxt_policy_modules_services_xenpmd.te.patch
@@ -1,6 +1,6 @@
---- a/policy/modules/services/xenpmd.te	1969-12-31 19:00:00.000000000 -0500
-+++ b/policy/modules/services/xenpmd.te	2015-01-05 16:03:12.793080030 -0500
-@@ -0,0 +1,73 @@
+--- /dev/null   1970-01-01 00:00:00.000000000 +0000
++++ refpolicy/policy/modules/services/xenpmd.te 2015-08-19 16:50:31.000000000 -0400
+@@ -0,0 +1,74 @@
 +#############################################################################
 +#
 +# Copyright (C) 2014 Citrix Systems, Inc.
@@ -50,6 +50,7 @@
 +dbus_send_system_bus(xenpmd_t)
 +uid_dbus_chat(xenpmd_t)
 +surfman_dbus_chat(xenpmd_t)
++dbusbouncer_dbus_chat(xenpmd_t)
 +
 +apm_stream_connect(xenpmd_t)
 +dev_rw_sysfs(xenpmd_t)
--- a/recipes-security/selinux/refpolicy-mcs_2.20130424.bbappend
+++ b/recipes-security/selinux/refpolicy-mcs_2.20130424.bbappend
@@ -1,7 +1,7 @@
 FILESEXTRA := "${THISDIR}/${PN}"
 FILESEXTRAPATHS_prepend := "${FILESEXTRA}:"
 
-PRINC := "${@int(PRINC) + 2}"
+PRINC := "${@int(PRINC) + 3}"
 
 SRC_URI += " \
     file://remove-xml-doc-gen.patch;patch=1 \


jean-edouard:
Looks like you manually hacked the patch without fixing the metadata though!
Please use quilt in your build tree to refresh the patch

Marsland:
Guilty as charged! :) I will correct the metadata.

Marsland:
I munged the metadata and updated the commit.

jean-edouard:
I guess that'll do, but next time please don't munge, and use quilt to properly update the patch.

Marsland:
Understood. The UI team hasn't been working directly with the OpenXT source with an active build box, I'm updating one Ian set up today to get that operational. Just wanted to get that SELinux patch up because we will need it in the build to get things working as we continue to flesh out xcpmd.

jean-edouard:
ack, LGTM, will merge soon.

jean-edouard:
I take it back, you don't seem to have updated the patch metadata...

Marsland:
OK, then I'll need some guidance getting this correct.

jean-edouard:
Hand-frobbing patches is risky business.
Here, after your modification, the last line of the patch doesn't get applied anymore.
If you look at the third line: `@@ -0,0 +1,73 @@` means we're adding 73 lines, so `patch` will stop after adding 73 lines. A dirty way to fix this is to replace 73 by 74.
The proper way to do this is to apply the patchqueue (OE does that for you), `quilt pop` to the patch you want to modify, make your modification and run `quilt refresh` to modify the patch.
The properly modified patch can then be found in patches/

Marsland:
Got it. The issue on this end is as UI developers we generally aren't developing against a complete OpenXT build tree from raw source. Our usual development environment is to download a completed build from the PXE to a development laptop and then use Chrome in a remote session to debug and code the appropriate HTML or javascript in a forked branch. Then we generate patch files from within git and submit them along with metadata that we normally edit directly, which is usually mundane as we almost always are simply adding patches. That's been the SOP up to now, but since we are dealing with core OpenXT code in this instance I'm getting up to speed on doing it from a standalone build box. Thanks for your enlightenment and support. :)

Marsland:
Just for expediency's sake, I'm going to suggest we abandon this pull request and do the patch the right way. It's a simple one line change to the openxt_policy_modules_services_xenpmd.te.patch to allow xenpmd messages through the dbusbouncer.
dbusbouncer_dbus_chat(xenpmd_t)

jean-edouard:
Ok, are you taking care of that?

Marsland:
Yeah, I'm going to give it a shot. Just refreshed our build box and have just this minute installed the resulting build on a test laptop successfully. So that system appears to be working, now I'll jump over to quilt and follow the steps to modify the patch. Provided that goes well I'll copy the resulting files and redo the commit.

Marsland:
Ran into a snag. I have no patches in my series when I run quilt.

jean-edouard:
Are you in `tmp-eglibc/work/xenclient_dom0-oe-linux/refpolicy-mcs*/refpolicy-mcs*/`?

Marsland:
No. Let me try that and see.

Marsland:
Nope. I don't see those paths on this system. I must not have the correct directory tree on this system. Ian set this up to build the version we've been testing/developing the UI changes for xcpmd but I believe he simply followed the steps outlined on the OpenXT wiki, so I just see the user/openxt directory tree for the actual build. I think what I need to do is go to square one and create a build box from scratch to match the development environments the rest of the team is using.

jean-edouard:
Try to `cd user/openxt/build`, the relative paths usually start from there.

Marsland:
Got it. I just sat down with Chris Rogers and he got me oriented.

Marsland:
All right. Got this done with a little help from Chris. Let me know if it gives you any trouble now.

jean-edouard:
Are you using a special quilt configuration? That's not what its default header look like...
Why don't you bump the package version anymore? That seemed like a good idea.

Marsland:
Well, Chris felt it wasn't necessary to bump the package version, so I backed that change out.

And you have an incredible eye for detail. After Chris and I ran through the entire quilt process the result was so nearly identical to the one I'd 'hand frobbed' as a backup, (in case we couldn't get quilt to go on the build machine), I decided to submit it.

If you prefer, I'll rebump the package file and copy the quilted file from the build machine to my current dev box to complete the change. If it were anything other than a one line change, I wouldn't do it this way, that's for certain. Learned a lot about quilt today though, and thanks!

jean-edouard:
I'd like someone who knows more about OE than I do to chime in about the version bump.

My understanding is that you're supposed to do it when you modify a package that doesn't live in a separate repository, which seems to be the case here. (when the sources are in a separate repository, the version number includes the HEAD git hash, that changes after each commit already)

About the patch header, the current format is not the default quilt format because the patches were originally generated by git. It's ok to leave it as it is, even if the quilt format is preferred.

Marsland:
I've staged both changes in my branch and will stand by for comments regarding the version bump.

ktemkin:
If it's changing something in xenclient-oe, there should be a bump to one of the version tokens (PV or PR). Sometimes that can happen implicitly (e.g. if you're ensuring you change the a repository somewhere else that determines the PV), but most of the time it's a good idea to bump PR.

In this case, that'd be accomplished by changing the following line in ```xenclient-oe/recipes-security/selinux/refpolicy-mcs_2.20130424.bbappend```:

```
   PRINC := "${@int(PRINC) + 2}"
```

to

```
   PRINC := "${@int(PRINC) + 3}"
```

cjp256:
Please fix git config (author) and use --signoff.

Other than that, LGTM (though the /dev/null / pathing change in the patch header looks suspect?)

Marsland:
Done as requested. Thanks.

jean-edouard:
@cjp256 this must be due to the quilt reformatting. LGTM

PR #148, opened 2015-08-17
Fix patch-queue subdir

eric-ch:
Directory names confusion... I'm sorry.


jean-edouard:
The issue was found after merging https://github.com/OpenXT/xenclient-oe/pull/145
This is needed to fix the build. Merging (very) soon.

PR #150, opened 2015-08-18
bbclasses: get makeclean to depend on force_rebuild

jean-edouard:
This will simplify working in OE.
It may not be the best way to implement it, please chime in if you know better.

### DIFF
--- a/classes/xenclient-customtask.bbclass
+++ b/classes/xenclient-customtask.bbclass
@@ -45,7 +45,7 @@ do_force_rebuild[depends] = ""
 do_force_rebuild[nostamp] = "1"
 
 addtask makeclean
-do_makeclean[depends] = ""
+do_makeclean[depends] = "${PN}:do_force_rebuild"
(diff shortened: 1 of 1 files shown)


rossphilipson:
I think there might be a couple of better ways to do this since there are plenty of packages that do incremental builds just fine and you may not want to do a full build (e.g. qemu that takes a while for a full build).

1. Add a new options like force_clean_rebuild that does both.
2. Switch it around and always make the makeclean option do a force_rebuild.

jean-edouard:
I agree, I didn't think about that use case.
Your second suggestion makes a lot of sense, since `force_rebuild` brings the package back to the "configured" state, which is exactly what you would expect after a `makeclean`.
I'll switch the dependency around and see if it works.
Thanks

jean-edouard:
Looking at the code again, "2." seems to be what I actually did.

rossphilipson:
+1 - I am good with #2. I will leave it open for a bit for other comments...

PR #152, opened 2015-09-01
Tboot uprev x2

rossphilipson:
This is the cumulative work to move to tboot 1.8.3. See the ticket for more details on the patches:
https://openxt.atlassian.net/browse/OXT-346
The individual commits and patch headers have more information.


rossphilipson:
I wanted to get this out there for people to try on as many platforms as possible. That is probably where we will see problems. I will continue testing on whatever I can find here.

pearsonk:
Is there a hosted build already on the AIS server?
Is there a confluence page or something to track tested platforms, CPUs, chipsets? It would be good to document the breadth of testing here so we can ensure we cover the bases.

rossphilipson:
There is one running that should pop out here in 4 hours:
http://openxt.ainfosec.com:81/builds/master/ext-dev-288-master

No confluence page. Do you want a page with a table? Or just line items? In the past we have tried to capture this sort of information and never had a standard for it.

pearsonk:
For this I think a table would suffice. Also, any test procedures to verify (or log messages to watch for) to see if it is working or not working would greatly help. With this, I should be able to get some non-technical help to assist.

rossphilipson:
Testing reveals that S3 is badly broken with this new version of TBOOT. This PR should not be merged until this is sorted out.

rossphilipson:
Closing for now - this cannot be merged with the major S3 regressions.

PR #153, opened 2015-09-14
[linux-pq] Remove incorrect memsets.

eric-ch:
pending_reqs and pending_segments are both allocated through kzalloc().
No need to memset() those to 0, also the size given to memset is invalid
and will cause recent gcc versions to output some warnings.


rossphilipson:
ACK I will merge it in a while...

PR #154, opened 2015-09-21
[ralink] install in correct images

eric-ch:
NDVM does not have USB support so rt3572 and rt2870-firmware are not
required there.
On the other hand, rt3572 could find its use in the installer.

### DIFF
--- a/recipes-core/images/xenclient-installer-image.bb
+++ b/recipes-core/images/xenclient-installer-image.bb
@@ -40,6 +40,7 @@ IMAGE_INSTALL = "\
     kernel-module-e1000e \
     linux-firmware \
     rt2870-firmware \
+    rt3572 \
     ${ANGSTROM_EXTRA_INSTALL}"
 
 IMAGE_FSTYPES = "cpio.gz"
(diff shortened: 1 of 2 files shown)


jean-edouard:
I agree that NDVM shouldn't have any USB driver.

However, I'm not sure we should allow installing using an unsupported NIC. With its Xen, tboot, SELinux support and various checks, the installer has always doubled as a compatibility checker. I'm not sure we still care about that though...

rossphilipson:
I agree with with Jed. If our platform doesn't support a NIC the installer shouldn't. As a user I would be confused that the install went just fine but in the end I have no network connectivity with my new install. Also we sort of want to discourage the use of USB NICs with OpenXT in general.

eric-ch:
Makes sense, in this case, rt* recipes should go I guess. That will be another PR.

rianquinn:
The be honest, I would prefer USB support for install, even is the NDVM does not support it. Specifically for the Ultrabook use case. Having USB support would provide PXE install support. Once OpenXT is installed, you could then use WIFI from that point on. Removing USB support for the installer forces the use of a USB stick, or USB CD-ROM to install.

transpute:
If the installer retrieves a signed installation image from the network, the local USB boot device could be relatively static.

rossphilipson:
@rianquinn I can see your point. I think though, it goes beyond this (closed) pull request. The questions becomes do we want to add all the USB NIC drivers to the installer. I think it is a question for the mailing list TBH.

eric-ch:
Since that PR cannot be re-open, please consider https://github.com/OpenXT/xenclient-oe/pull/156 the following.

PR #155, opened 2015-09-21
[ralink] Remove module from images.

eric-ch:
NDVM does not support USB, so remove the ralink USB Wifi dongle modules
and firmware from it.

While OpenXT installation could be done using a USB dongle, NDVM
would not handle network after that leading to confusion. So remove the
USB dongle support from the installer as well.


eric-ch:
Lets keep the discussion in one place:
https://github.com/OpenXT/xenclient-oe/pull/154

PR #156, opened 2015-09-22
[ralink] Move from NDVM to installer.

eric-ch:
NDVM does not support USB, so remove the ralink USB Wifi dongle modules
and firmware from it.

This most likely belongs in the installer to allow PXE installation on
platforms without ethernet.

Discussion started in: https://github.com/OpenXT/xenclient-oe/pull/154, but changes in the branch have locked the PR.


jean-edouard:
+1

rossphilipson:
+1 We are going to start a general conversation on adding more USB NIC support to the installer on the mailing list. This is fine for now since it is mainly about removing this driver from the NDVM which makes sense.

PR #159, opened 2015-09-23
[ralink] Remove module and firmware.

eric-ch:
rt2870/rt357x devices are USB Wifi dongles only.
There is no Wireless support in dom0's kernel shared by the installer. There is no claim to support installation through such means and no infrastructure in OpenXT either.

Currently, rt3572 will break the build in dom0's kernel (CONFIG_WIRELESS=n -> CONFIG_WEXT_PRIV=n).


rossphilipson:
+1 lets unbust the build

PR #160, opened 2015-09-23
Kernel uprev to 4.1

eric-ch:
Upgrade Linux kernel of dom0 and service VMs to 4.1.7.
Upgrade libdrm to 2.4.64 so headers, used in drm-plugin for example, reflect the ioctl changes.

OXT-376


pearsonk:
I'm assuming this should be merged alongside https://github.com/OpenXT/openxt/pull/77

aikidokatech:
Is this stalled?  Should it be swapped out for 4.3?

eric-ch:
It has been delayed at least.
I will try to maintain it, if only for testing purposes. I did some preliminary work for 4.3, but nothing ready to be PR'ed yet.
4.1 being longterm, it is still a valid candidate for an upgrade after 3.18, imho.

aikidokatech:
Were there changes elsewhere for 4.1 that would cause an issue with 3.18?  Could we have both in the repo and then a person could pick which to use at build time?

eric-ch:
@aikidokatech Only libdrm being upgraded to match the ioctl table (that will overlap with our added i915 foreign ioctl). That's easily dealt with using the DISTRO env. Also it needs a rebase. Last PRs were aiming to ease having "testing" kernel.

aikidokatech:
Ah, I do keep forgetting about that one.  A way to resolve that could be to pick an higher ioctl for both.  0x40 or so.

eric-ch:
Indeed, though I was thinking of keeping both recipes for libdrm and choosing one or the other in a distro.conf, depending on the kernel version used. Actually that would only solve problems we do not have... choosing a consequent offset for the ioctl() (since libdrm ioctl list changes once in a while) should keep us safe for a while.

eric-ch:
In the end, lets split this in different PRs: one to change the ioctl and upgrade libdrm, one to introduce Linux 4.1 patches & recipe.

eric-ch:
Created ticket https://openxt.atlassian.net/browse/OXT-444 to track libdrm upgrade. I will re-open a PR to upgrade Linux 4.1 with a rebased branch once https://github.com/OpenXT/xenclient-oe/pull/191 is in.
Until then, I will maintain a fork with Linux 4.1: https://github.com/eric-ch/xenclient-oe/tree/linux-4.1 if someone needs this version of the kernel.
.

PR #162, opened 2015-09-24
xenclient-deb-kernel.bbclass: be nice to Linux 3.18

jean-edouard:
The version of Lintian that is in Debian Wheezy doesn't seem to be compatible with Linux 3.18.
This override seems to make it work.


rianquinn:
was this the reason debian was crashing as a Linux Guest in OpenXT?

jean-edouard:
I don't think so. Debian is crashing as a Linux Guest because of ac97 though :)

rianquinn:
Yeah, that we are aware of. That's likely going to need a patch to QEMU to blacklist the kernel from running that automatic sample rate detection code since it's clearly broken. I could be mistaken, but I thought there might be something else causing issues with Debian

eric-ch:
LGTM.

PR #163, opened 2015-09-25
xen-tools: enable xz compression support

jean-edouard:
This is needed to boot standard Debian kernels as PV guests.

All the credit goes to @rossphilipson for the find :)


rossphilipson:
+1

PR #164, opened 2015-09-29
fix config file permissions for the vusb daemon

jean-edouard:
WARNING: this is stable-4, not master

Without this fix, the vusb daemon can't create and write to /config/etc/USB_always.conf, losing auto-assign settings on reboot.

This may not be the correct/best way to fix it, but that's my best shot and it seems to work.


cjp256:
LGTM, merging

PR #165, opened 2015-09-30
linux: dom0: enable I2C HID support, and drivers for Haswell-ULT cont…

jean-edouard:
…rollers

OXT-396

### DIFF
--- a/recipes-kernel/linux/linux-xenclient-dom0-3.18/defconfig
+++ b/recipes-kernel/linux/linux-xenclient-dom0-3.18/defconfig
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/x86 3.18.16 Kernel Configuration
+# Linux/x86 3.18.19 Kernel Configuration
 #
 # CONFIG_64BIT is not set
 CONFIG_X86_32=y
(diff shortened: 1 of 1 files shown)


eric-ch:
LGTM, please add OXT-396 in the commit message though :).

ktemkin:
+1; LGTM

rossphilipson:
Has anyone tried this fix out - tested it?

rossphilipson:
Also if we are ultimately going to move to say 4.3, won't we pick this stuff up there?

ktemkin:
@rossphilipson:
I haven't built with this PR exactly, but I do have a build with those options turned on-- they enable touch support for more modern touchscreen platforms (Dell Venue 11, Dell 7350, etc.). 

I'm not sure what you mean by "won't we pick this stuff up [in 4.3]". This looks to me like it's just turning on configuration options that aren't set by default-- this change should be necessary no matter what kernel version we use.

rossphilipson:
Oh sorry I got this confused with something else (a back port). My bad.

jean-edouard:
@eric-ch, done

rossphilipson:
Ok then. Two +1s and @ktemkin has been running w/ those options on. Merging.

PR #166, opened 2015-10-05
remove upower from dom0

jean-edouard:


rossphilipson:
ACK. We don't need that anymore now that everything talks to xcpmd. Will merge soon.

PR #167, opened 2015-10-05
[tboot] Get rid of LCP bin and configure script.

rossphilipson:
OXT-349


rossphilipson:
Did a build, ext-306, finished without problems. Installed the build with TXT on - no problems. I think it is safe to merge this.

PR #168, opened 2015-10-14
Adding recipes for a couple of HID tools

jean-edouard:
Those two tools work together to dump and translate HID descriptors and streams.
They are very useful when working with HID devices.
This PR just adds the recipes, they won't get built by default.
They should never be added to the dom0 task, but it would make sense to add them to task-xenclient-extra.bb, to be able to simply `opkg install` them on "production" builds.


eric-ch:
I would have thought those to belong in recipes-devtools.

jean-edouard:
Good point, makes sense, moving them in a few.

jean-edouard:
Done

PR #169, opened 2015-10-16
[vusb] URB_SHORT_NOT_OK should only be set on IN URBs

rossphilipson:
OXT-394


jean-edouard:
+1, I know about this, will merge soon.

PR #170, opened 2015-10-19
Add YAJL as dependency of XCPMD.

jctemkin:
Since there seems to be community consensus to use YAJL as OpenXT's standard C JSON parser, I've written this pull request to add the newest version of YAJL to recipes-devtools.

In related news, since [xctools PR15](https://github.com/OpenXT/xctools/pull/15) depends on YAJL, I modified XCPMD's recipe to reflect this.


rossphilipson:
I will start a build with this and the XCPMD work

cjp256:
Recipe LGTM (pulled from upstream).

jean-edouard:
Our meta-openembedded repo already has a recipe for yajl (v2.0.3).
Do we really need 2.1.0 for some reason?

It's usually preferable to use the existing recipes when possible, as there's more chances of them being compatible with everything else. (think building something from source for your dev machine, instead of apt-get installing the thing)

jctemkin:
We don't strictly need 2.1.0, it just includes some bugfixes and performance improvements that would be nice to have. But if you think that it may introduce some compatibility problems, sticking with 2.0.3 should work fine also.

rossphilipson:
Build succeeded with this change: http://openxt.ainfosec.com:8020/builders/extbuilder/builds/312

I will merge this soon.

PR #173, opened 2015-10-30
[qemu-dm] Remove lingering patches.

eric-ch:
Patches using old naming system got left around but unused. Remove to
avoid any confusion.


rianquinn:
+1

rossphilipson:
+1 definitely cleanup is good - will merge soon

PR #174, opened 2015-11-03
linux: remove seemingly useless and harmful mass storage patch

jean-edouard:
This patch is about preventing usbstorage from binding to USB mass storage devices.
I can see how it made sense with the old usb backend, since we didn't use to reset USB devices before/after passing them to VMs. USB device binding was disabled system-wide, and the USB daemon would bind some devices to dom0.
With the new USB daemon and USB backend, we're now doing things the normal way, and this patch doesn't seem to make sense anymore.

Practically, this patch sometimes prevents USB drives from working properly, which is annoying.


rossphilipson:
+1 but I will leave it open for comments...

eric-ch:
+1, if it makes sense with the new usb-daemon.

PR #177, opened 2015-11-08
Fix kernel OOPS when changing backend for XenBus device with no driver.

ktemkin:
Fixes a kernel OOPS that disrupts a VM's kernel if it winds up with XenBus device (backend or frontend) that it can't provide a driver for.

See https://openxt.atlassian.net/browse/OXT-420.

OXT-420


rianquinn:
Please add the patch documentation

ktemkin:
@rianquinn:
This is a modification of an existing patch, which already has the patch documentation header.  

See the full file here: 
https://github.com/ktemkin/xenclient-oe/blob/fix_xenbus_oops/recipes-kernel/linux/linux-xenclient-3.18/xenbus-move-otherend-watches-on-relocate.patch

rianquinn:
+1

rossphilipson:
@ktemkin I assume you tested this? The fix makes sense so I am tempted to merge it w/o a whole lot of custom builds and testing...

ktemkin:
@rossphilipson I tested this for a short time-- I didn't see any adverse effects using OpenXT for a few hours, and it does fix the bug.

rossphilipson:
Sounds good, merging...

PR #178, opened 2015-11-10
rhino: fix URL

jean-edouard:
OXT-415


Marsland:
The fix I was given is https://ftp.mozilla.org/pub/js/rhino1_7R2.zip but +1, this https://ftp.mozilla.org/pub/mozilla.org/js/rhino1_7R2.zip appears to work fine too

ktemkin:
+1

PR #179, opened 2015-11-10
Allow PCI USB3 keyboards (most 2-in-1 keyboards) to be used for reseal.

ktemkin:


cjp256:
LGTM, though missing signoff. :)

jean-edouard:
+1

rossphilipson:
I commented on a priv chan about this:

> ktemkin: i am fine with that though given that the pci glue drivers depend on the hcd drivers maybe we should just modprobe all the *hci-pci.ko drivers instead
> that should drag up all the *hci-hcd.ko drivers and all the other dependencies
> on debian they just cut out the middle man and build the pci glue right into the hcd driver

jean-edouard:
I tested with just those 4 (see below), and both the integrated laptop keyboard and the one on the USB3 port worked fine. As Ross suggested, *hci-pci depend on *hci-hcd (apart from uhci, that includes the pci bits inside the hcd driver).

     modprobe ehci-pci
     modprobe uhci-hcd
     modprobe ohci-pci
     modprobe xhci-pci

ktemkin:
@rossphilipson: I agree-- I'll update the patch.
@jean-edouard: Thanks for testing; I'll update the PR accordingly. :+1

ktemkin:
Modified per the suggestions above.

rossphilipson:
Thanks for fixing that. That is what Jed tested so I think this can go in. Will merge soon...

PR #180, opened 2015-11-13
initramfs: use busybox instead of busybox-static

cjp256:
Upstream oe-core does not use busybox-static for their
images, and it does not appear to be required for the
dom0 initramfs to function.

As the initramfs is the only user of busybox-static,
we can remove busybox-1.11.3 entirely now.


cjp256:
Tested this by building initramfs (and rebuilding dom0).  System booted OK.

Configured measured launch, caused it to trip, ensured that input continued to work during reseal/recovery process.

rossphilipson:
I think this one needs a full build too - started one here: 
http://openxt.ainfosec.com:8020/builders/extbuilder/builds/322

cjp256:
@rossphilipson Thanks!

I did a network install of ext build 322 and it booted fine for me.

jean-edouard:
I bisected the two config files, and by switching to the new busybox we lose:
- syslog
- syslogd
- vi
- netcat
- telnetd
- traceroute
- dc (reverse polish calculator)

It's probably ok, I can't think of anything using any of those, but if anyone knows about a part of the initramfs using something from that list, please chime in.

jean-edouard:
I couldn't find anything using those, I think we're good, I will merge soon.

PR #181, opened 2015-11-17
Define recipes versions in distro.

eric-ch:
PREFERRED_VERSION variables are split between the general local.conf
file (in openxt.git) and machine specific configurations.
Regroup everything under a distro configuration (openxt-main in this
case), leaving machine specific configuration with PREFERRED_PROVIDER.

This is only a small attempt at making recipe selection a little bit easier than it is using the DISTRO configuration of Bitbake. I might have missed the canonical way of doing so.

See related commit in https://github.com/OpenXT/openxt.

### DIFF
--- a/conf/distro/openxt-main.conf
+++ b/conf/distro/openxt-main.conf
@@ -1,0 +1,34 @@
+#
+# This is how bitbake will establish its configuration:
+#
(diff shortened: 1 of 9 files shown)


cjp256:
LGTM, built fine here.  Will merge.

PR #182, opened 2015-11-18
Set XenVESA IO ports for xenvesa driver.

eric-ch:
Use IO ports 0x383A-0x383D from the shadow bda region defined in QEMU to
pass graphic information to windows guest driver.

This patch was left out during the QEMU 1.4 upgrade.

### DIFF
--- a/recipes-openxt/vgabios/vgabios-0.7a/vbe-xenvesa.patch
+++ b/recipes-openxt/vgabios/vgabios-0.7a/vbe-xenvesa.patch
@@ -1,0 +1,116 @@
+################################################################################
+SHORT DESCRIPTION: 
+################################################################################
+Use the "shadow-bda" IO ports region (0x383A-0x383D) to pass a table to the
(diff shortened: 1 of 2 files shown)


transpute:
If possible, associate this PR with an existing or new ticket in the Qemu 1.4 epic, https://openxt.atlassian.net/browse/OXT-118

smeisner:
+1

jean-edouard:
LGTM, will merge soon.

rossphilipson:
We should leave it open so Eric can add a ticket number to it. I think using the epic itself is fine for something like this (OXT-118)

eric-ch:
I prefer to create a separate ticket linked to that epic (OXT-430). The epic scope seems too general.

rossphilipson:
Ok I will fix my commit that goes along with this to use that ticket number...

jean-edouard:
LGTM, will merge soon.

PR #183, opened 2015-11-18
[xenvesa] Improve comments and fix incorrect length of register block.

rossphilipson:
The comments need to include a reference to the registers used to store
the xenvesa support structure too.

OXT-430


eric-ch:
I think it should be in its own ticket (linked to OXT-118), being related to one behaviour in particular.
I took the liberty to create OXT-430 for this one and amend the related patch for vgabios.
+1 otherwise :).

rossphilipson:
Modified to use OXT-430

jean-edouard:
LGTM, will merge soon.

PR #184, opened 2015-11-18
Update dmidecode to the latest version (3.0).

rneilturner:
Upgrade the dmidecode package to 3.0 to remove warnings generated for systems with SMBIOS 2.7 or later. The warnings had two side effects:

1.) Spammed the hardware info tab within the UIVM
2.) Laptop systems were no longer being determined as laptops. This was due to the toolstack not being able to parse the chassis-type due to the comments.


cjp256:
With your git commit, use --signoff.

You should also remove (or use git mv) the old version.

Otherwise, LGTM.  I can test it after the commit is fixed up.

transpute:
If you create a JIRA ticket for this change, you can set the "Component" field so this change is grouped with other changes that affect the user interface.   JIRA will automatically associate this PR with the ticket, if the ticket number is included in the PR description.

cjp256:
@transpute : a jira ticket for every commit / PR is a bit overboard.  That's the point of the git history.  Perhaps micro-managing this for stable branches is useful, but not every day master changes.

@rneilturner : this updated commit looks good. will give it a spin.

ktemkin:
LGTM, as well.

transpute:
@cjp256 No one said that a JIRA ticket is needed for every PR.  This PR fixes a user-visible bug, which was not reported in JIRA prior to the fix.  Some projects track user-visible bugs in JIRA.  Among other benefits, it makes it easy for a stable release to generate a list of user-visible bugs that have been fixed.  Some users and customers place value on a list of fixes in release notes.

cjp256:
@transpute Well just about every commit is either a feature change or bug fix (which likely came about because a user/developer noticed the issue).  I'm not sure what dictates the requirement to generate the additional JIRA ticket (in addition to the commit and PR documentation).  It seems to be unclear per https://openxt.atlassian.net/wiki/display/OD/How+to+contribute. 

You should RFC this topic so we can standardize the guidelines and update the contribution wiki accordingly.

rneilturner:
See OXT-429 for the JIRA ticket, move on.

rossphilipson:
+1 ... I will try building this - if it builds then I think it can go in.

rossphilipson:
Looks good, will merge soon...

PR #185, opened 2015-11-18
Master fix clean targets

cjp256:


jean-edouard:
Can you please provide more details about what this is doing? Thanks

aikidokatech:
I believe this is to fix the fact that at least three recipes file if attempting to build a second time without an explicit clean.  Examples are xenfb2 and v4v-module.  Being that development traditionally was always done with a builder that uses a fresh tree, it was never fixed before.

cjp256:
aikidokatech is right.

if fixes clean steps, which I have hit doing jethro-based work.

cjp256:
ping! :)

jean-edouard:
I still don't get it, but @aikidokatech seems to +1, I'll merge soon.

PR #186, opened 2015-11-19
An attempt to tidy up recipes-kernel/linux before new versions come along.

eric-ch:
The linux subdirectory is getting messy. Try to tidy up a little bit by
splitting things in versions related subdirectories.
Also:
- Move machine specific patches to a common patch-queue for each kernel.
- Move all the defconfigs in one subdirectory to avoid having
subdirectories for each machine to have the file where bitbake expects
it to be (this add a step in linux-xenclient.inc)
- Unify how 3.11 and 3.18 recipes work for good measure.

OXT-433


cjp256:
I like this cleanup.

Do we even need linux-xenclient-[dom0|ndvm|syncvm|..]_3.11.10.4.bb ?  Can we simplify further and just have one with separate def configs per machine name?

jean-edouard:
That would be awesome. A long time ago, when we switched to OpenEmbedded, I asked the guy who did the initial commit why he did multiple kernel recipes. He had a reason, that I think I didn't understand at the time, that may or may not make sense...

eric-ch:
@cjp256 I do not see why we couldn't. I'll give it a try and amend the PR if that "looks good".
@jean-edouard I seem to remember there was a reason as well, I can't remember though...

aikidokatech:
The only reason I can think of is if you have differences across images with the same machine type.  Otherwise you should be able to affect it based on the machine type.  I did that with the initscripts.

eric-ch:
This last commit only use one recipe for the kernel all the machines.
Also rebased on top of master upstream.

jean-edouard:
May I suggest amending that last commit to rename the recipe to something like "linux-openxt"? :)
If there's more implied than just the scope of the existing commit, please ignore this.

aikidokatech:
@eric-ch Right, I believe it automatically will become machine specific.  Therefore we can easily keep separate configs and such.  Only trouble will be when we need images that have the same machine type to have different kernel settings.  An example of something difficult (as I see now) would be a service VM kernel template.  More likely others will copy that and make a new machine type as service VM kernels will most likely differ.

eric-ch:
@aikidokatech OE kernel.bbclass sets PACKAGE_ARCH = MACHINE_ARCH, that should make it machine specific. Thanks for pointing out the service VM template issue, that would definitely be difficult having a separate recipe.
@jean-edouard There is not much more really, if nobody has any concern about that, I can do it (for mulltiple recipes or only one).

ktemkin:
I'm very much for having a single linux-openxt recipe-- and this looks like some awesome work towards making this happen.  Definitely +1.

One thing to note: the ```SRC_URI = "defconfig-{$MACHINE}"``` magic isn't strictly necessary; and IIRC it's not the recommended way to do things in openembedded. OE already has a mechanism for per-system file overrides: when looking for a file from ```SRC_URI```, it automatically checks a variety of package, machine, architecture, and toolchain-specific subfolders for the file. 

For our ```linux-xenclient-oe``` recipe, it actually checks ~120 default paths to see if there are any more specific *defconfig*s than the ones we're provided. An excerpt from that path list, as found in ```FILESPATH``` on my machine:

```
<...>/xenclient-oe/recipes-kernel/linux/linux-xenclient-dom0-3.18.19/linux:
<...>/xenclient-oe/recipes-kernel/linux/linux-xenclient-dom0-3.18.19/i586:
<...>/xenclient-oe/recipes-kernel/linux/linux-xenclient-dom0-3.18.19/build-linux:
<...>/xenclient-oe/recipes-kernel/linux/linux-xenclient-dom0-3.18.19/pn-linux-xenclient-dom0:
<...>/xenclient-oe/recipes-kernel/linux/linux-xenclient-dom0-3.18.19/xenclient-dom0:
<...>/xenclient-oe/recipes-kernel/linux/linux-xenclient-dom0-3.18.19/x86:
<...>/xenclient-oe/recipes-kernel/linux/linux-xenclient-dom0-3.18.19/:
<...>/xenclient-oe/recipes-kernel/linux/files/i586:
<...>/xenclient-oe/recipes-kernel/linux/files/build-linux:
<...>/xenclient-oe/recipes-kernel/linux/files/pn-linux-xenclient-dom0:
<...>/xenclient-oe/recipes-kernel/linux/files/xenclient-dom0:
<...>/xenclient-oe/recipes-kernel/linux/files/x86:
<...>/xenclient-oe/recipes-kernel/linux/files/
```

These work a lot like a normal linux ```$PATH```-- bitbake searches for the relevant file by looking through these paths in left-to-right priority. The canonical OE way to specify a per-machine defconfig would be to create per-machine subfolders in e.g. your ```3.18``` directory, and then place a defconfig in each of those. 

Here's an example from my experiment-with-Xen-on-ARM repository:
https://github.com/MobileXT/meta-mobilext/tree/master/recipes-kernel/linux

eric-ch:
@ktemkin Indeed, this is how we have machine specific defconfigs currently (e.g, https://github.com/OpenXT/xenclient-oe/tree/master/recipes-kernel/linux/linux-xenclient-ndvm-3.18 for NDVM). I did not like this as it makes a subdirectory for each machine (we could  have that be a subsubdirectory in a subdirectory easily, but still) only for the defconfig (being machine specific anyway in our case), hence the step to have all defconfigs in a single subdirectory.

So I'm not sure about the 2 cosmetic changes:
1- Should we have 1 kernel recipe per kernel version + patch-queue?
2- Should we split machine-specific defconfigs in machine-specific subdirectories (OE canonical way and current method) or have a step (possibly hackish/convoluted do_rename_defconfig in this PR) to have them selected on the machine name, since the kernel+defconfig is machine specific anyway?

My personal opinion is:
1- yes,
2- I could go with both now that I think about it...

aikidokatech:
@eric-ch I think the mention of "hackish/convoluted" points to which we should pick.  I don't see a reason to avoid the OE way.

eric-ch:
@aikidokatech :) I agree.

eric-ch:
Current PR status:
1- One linux-openxt kernel recipe for each machine,
2- Machine specific defconfigs using standard OE/Bitbake (no added step),
3- linux-xenclient renamed to linux-openxt,
4- One patch-queue per kernel version (major.minor) applied, common for all OpenXT machines.

jean-edouard:
This looks pretty good now, I'll kick a build.

jean-edouard:
This build and installed fine (http://openxt.ainfosec.com:8020/builders/extbuilder/builds/329)
Unfortunately it now conflicts :(
Please fix and I'll merge, sorry about that...

eric-ch:
@jean-edouard done, usbback-base.patch was updated upstream.

jean-edouard:
Awesome, thanks, merging now.

PR #187, opened 2015-11-19
Backport XSA-156.

eric-ch:
Add XSA-156 patch to Xen patch-queue.
See: http://xenbits.xen.org/xsa/advisory-156.html


rossphilipson:
The port looks straight-forward, the code looks correct to me and Eric said he tried it. I guess to be 100% tested it we would need to make sure HVM debugging and debugging with the gdbsx thingy all works right but I think we would find that out in due course.

+1 and I am for merging. Comments?

jean-edouard:
+1

rossphilipson:
Will merge in a while...

PR #189, opened 2015-11-24
Allow QEMU to be notified when guests wake from S3.

ktemkin:
**Test note:** Can be merged in any order, but should be tested with: https://github.com/OpenXT/toolstack/pull/11

Prior to this patch, QEMU requires a wake signal to be communicated
from the input server via the DMbus, but:
  - This is nonideal, as it requires the input server to be running, and
    requires the input server to circumvent the toolstack for waking guests.
  - This means that guests are not woken properly when using toolstack calls,
    such as "xec-vm -n <domain-name> resume".

This commit modifies QEMU to also respond to wake requests via a XenStore node.
Ideally, this information would be communicated via a protocol like QMP; this is
to be considered a stand-in until QMP support is integrated.

NOTE:
While the added patch could be merged into acpi.patch, in this case it seems to
make more sense to leave it separate. Ideally, this patch will eventually be
succeeded by a QMP-based solution, and this patch can just be removed without
having to extract anything from acpi.path.

### DIFF
--- a/recipes-openxt/qemu-dm/qemu-dm-1.4/0035-add-acpi-wakeup.patch
+++ b/recipes-openxt/qemu-dm/qemu-dm-1.4/0035-add-acpi-wakeup.patch
@@ -1,0 +1,153 @@
+################################################################################
+SHORT DESCRIPTION: 
+################################################################################
+
+Allow QEMU to be notified when guests awake from S3.
+
+################################################################################
+LONG DESCRIPTION: 
+################################################################################
+
+Prior to this patch, QEMU requires a wake signal to be communicated
+from the input server via the DMbus, but:
+  - This is nonideal, as it requires the input server to be running, and
+    requires the input server to circumvent the toolstack for waking guests.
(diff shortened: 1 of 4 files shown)


jean-edouard:
That looks much better than my input_server signal. Don't you think we should remove that?

ktemkin:
Assuming you're suggesting that I also remove the dmbus wake signal, I agree-- though that seemed like a separate issue to me (i.e. the toolstack wakes guests manually, rather than asking the toolchain to do it), so I was planning on doing that in a subsequent ticket/PR. 

Splitting it into a separate JIRA ticket and PR would likely also make this easier on our testers.

If it's preferred, I can do it all at once.

jean-edouard:
Separate PRs make sense, thanks for your reply, will test and merge.

ktemkin:
Thanks. Opened OXT-446 to track the extraction of the now-duplicate waking code from the input server.

jean-edouard:
It built, and a VM went to S3 and back just fine.
LGTM, will merge both PRs soon.

PR #190, opened 2015-11-26
drm-plugin, Fix QA issue.

eric-ch:
Trivial change to avoid:
QA Issue: non -staticdev package contains static .a library


jean-edouard:
Nice, will merge soon.

PR #192, opened 2015-11-27
Make 4f15 01 returns a valid EDID.

eric-ch:
Remove the EDID copy on the stack and modify the returned EDID in place
seems to avoid memory corruption. The returned EDID is now correct,
though lack of a complete explanation makes this a work-around.

See associated ticket for more details.

OXT-445

### DIFF
--- a/recipes-openxt/vgabios/vgabios-0.7a/vbe-edid-interface.patch
+++ b/recipes-openxt/vgabios/vgabios-0.7a/vbe-edid-interface.patch
@@ -46,9 +46,11 @@ Depended on by VGABIOS: vbe-extensions.patch
 ################################################################################
 PATCHES
 ################################################################################
---- a/vbe.c
-+++ b/vbe.c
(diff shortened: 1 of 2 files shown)


rossphilipson:
I agree it would be nice to understand why this started not working at some point. It worked correctly for years. Perhaps in the newer vgabios we are using the layout is different enough that just by bad luck we are are close to the end of the stack seg when we make the EDID fetching call and the stack buffer does not fit...

Ultimately though, I think the code is cleaner to load the EDID into the caller supplied buffer at ES:DI and fix it up in place. Building it on the stack and copying it seems superfluous.

rossphilipson:
Will do a build and test soon...

rossphilipson:
Build with fix:
http://openxt.ainfosec.com:81/builds/master/ext-dev-331-master/iso/

Installed it and a Debian VM. Used get-edit to fetch it. The fix is working - I am getting the updated EDID with correct timings and checksum.

+1 will merge soon....

PR #194, opened 2015-12-05
add build step to switch git clones to the proper branch

jean-edouard:
This is an attempt at having git work directories on the right branch, instead of the current "(no branch)" state.
If you know a better way to do this, please chime in.

I got a successful build of this.

### DIFF
--- a/classes/xenclient.bbclass
+++ b/classes/xenclient.bbclass
@@ -1,3 +1,28 @@
 do_rm_work() {
         :
 }
+
+python do_checkout() {
+    from bb import note,build
+    import re
+
+    # The following like will list the whole bb environment, very useful!
+    # all = bb.data.keys(d)
+    # for x in all:
+    #     note("%s = %s" % (x, bb.data.getVar(x, d, True)))
+
+    src = bb.data.getVar("SRC_URI", d, True)
+    branches = re.findall('branch=[^ \t\n\r\f\v,;]+', src)
+    if len(branches) == 1 :
+        branch = branches[0].split('=')[1]
+        command = 'git checkout ' + branch
+        wd = bb.data.getVar('WORKDIR', d, True)
+        note('Switching the git clone to %s. You\'re welcome.' % branch)
+        note('  Directory: %s/git' % wd)
+        note('  Command: %s' % command)
+        os.chdir(wd + "/git")
+        bb.process.run(command, shell=True)
+        #runfetchcmd(command, d)
+}
+
+addtask checkout after do_unpack before do_patch


ktemkin:
This seems like it breaks a few of OE's git-fetcher features-- such as setting up a ```SRCREV```. To test, I tweaked my Surfman recipe to have SRCREV of ```aa95a96``` and tried to build; with your change, bitbake still build _master_.

This makes sense; even though bitbake correctly checked out ```aa95a96```, your step subsequently went and checked out ```master```.  (Unfortunately, the process of coming up with the commit that's fetched is more complex than just looking at the branch.)

jean-edouard:
Yes, initially my plan was to add that do_checkout step to all recipes and check the SRCREV / git HEAD.
But then I decided to do it only for openxt (xenclient) recipes, which are virtually the only ones that we care about. Afaik, openxt recipes never use a special SRCREV, so I didn't check for it.

But you're right, it could happen. I can modify this PR and add a condition like:
[ `git rev-parse $branch` = `git rev-parse HEAD` ]

Would that make this an acceptable PR?

Thanks

ktemkin:
> Afaik, openxt recipes never use a special SRCREV, so I didn't check for it.

Yeah, I don't think we use it in our recipes. I would like to preserve the ability for people to override SRCREVs in their local configuration using the "override" format:

```
SRCREV_pn-surfman = "<commit-hash-here>"
```

Checking the revision before you switch sounds like a better idea. I don't see any technical issues with the commit-- though I'd prefer to see some of comments and bit-bake notices stripped out.

I'm not sure if this is worth the additional maintenance surface area-- we differ from upstream OE in enough (often confusing) ways-- but then again, the change doesn't affect my workflow either way. If this is something that you think will significant help developers, I'm fine with it.

cjp256:
OE still uses the srcrev for packaging / environment purposes.  AUTOREV detects that srcrev for you.

I'm tending to agree with Kyle on this one, I'd avoid making changes like this that competes with OE's source and packaging mechanisms. This may cause other issues down the road.

This feature is a good example of something that should be done in upstream (openembedded-core), first, and then we can absorb it.

rossphilipson:
How about another approach where we make a build step that does this? Then if a developer wants to explicitly checkout master on the detached head repos before working, they can do that knowing what the result will be.

rossphilipson:
Jed, I think you should close this and make a ticket.

jean-edouard:
The ticket is here: https://openxt.atlassian.net/browse/OXT-353
However, it looks like this is not an issue anymore with Jethro.
Closing and commenting on the ticket.

PR #196, opened 2015-12-10
Add a distro to use Linux 4.1

eric-ch:
Create a DISTRO (openxt-linux-4.1) that uses Linux 4.1 kernel for OpenXT dom0 and service VMs.

By default, OpenXT builds will use DISTRO=openxt-main and not use Linux 4.1.

Change the configuration openxt.git/build/conf/local.conf, DISTRO variable, to change the default behaviour.


jean-edouard:
Adding support for a more recent kernel is clearly a good idea.
As it is optional, and not set by default for now, all that matters to me is that building without it doesn't change anything.
So I started a build that includes this PR but doesn't set the DISTRO variable. If it comes out fine, I'll merge. (http://openxt.ainfosec.com:8020/builders/extbuilder/builds/342)
Any objection?

pearsonk:
I really like this approach, so will +1 that. Haven't had time to do a build or test.

cjp256:
Didn't test it, but it LGTM. 

Is the primary differentiator for having multiple distro options just these two lines?
PREFERRED_VERSION_linux-openxt = "4.1%"
PREFERRED_VERSION_linux-libc-headers = "4.1%"

Perhaps in the future we can just document how to set those two (in say, site.conf) if you want to override the default?  May be easier than creating a new conf every time and keeping them in sync.

Probably could be cut down one line if:
PREFERRED_VERSION_linux-libc-headers ?= "${PREFERRED_VERSION_linux-openxt}"

eric-ch:
I started using DISTRO because some other bits could require upgrades (libdrm for example) as well, and those could be incompatible with the previous kernel version. This is no longer a problem, so I am fine doing it differently.

pearsonk:
+1 to the suggestion by @cjp256 as it will eliminate the possibility of only 1 of the 2 being set causing unforeseen behaviors or problems. I can't think of a reason why they should ever differ.

jean-edouard:
I'm ready to merge once the previous discussion is settled.

eric-ch:
@cjp256 After some reading on the meaning of these configuration files, I do not think we should put PREFERRED_VERSION_linux-openxt definition outside of a distro.conf (also applies to other packages) event if this is "all" the difference between the distributions, for now. Though, redundant information could be centralised in one common configuration file for good measure.
(mostly what I understand from https://git.yoctoproject.org/cgit/cgit.cgi/meta-yocto/tree/meta-yocto/conf examples).

Totally agree on having PREFERRED_VERSION_linux-libc-headers ?= "${PREFERRED_VERSION_linux-openxt}" though, I can't think of a reason not to have that.

eric-ch:
@cjp256, @pearsonk, Is commit https://github.com/eric-ch/xenclient-oe/commit/e7153e34a86f74e9658a402bef0af9c43eee02f8 enough regarding your comments?

pearsonk:
Looks good to me

jean-edouard:
Merging imminently.

PR #198, opened 2015-12-31
Jethro merge

aikidokatech:
These are the changes for building OpenXT using OE's Jethro (2.0) release.  It must be merged with the corresponding PR in openxt.  See https://openxt.atlassian.net/wiki/display/BS/Jessie+systemd+container for instructions on creating a new Jessie based build container.


rossphilipson:
I think before we can go further with considering the Jethro PRs we a the least need clear instructions for making a build machine. Traditionally we have described how to do it on bare metal like here:

https://openxt.atlassian.net/wiki/display/OD/How+to+build+OpenXT

As to whether later you want to do this in a container or VM should not really matter. I would like to see a wiki page like the one above for setting up a build machine. It can be a separate page for now and we can put it in Gettting Started later.

PR #200, opened 2016-01-25
Remove the ioemu recipe, since it seems to break the build

jean-edouard:
OE seems to expect all the repositories for all the recipes to have the current branch.
ioemu doesn't have a stable-5 branch, so its recipe seems to break the build just by being there...


rossphilipson:
That gets rid of a lot of unused junk. We should do that on master too at some point.

PR #202, opened 2016-01-26
Fix dmi product_uuid endianness for SMBIOS >= 2.6

eric-ch:
Backport ff4319d to fix UUID endianness for SMBIOS >= 2.6

For test installations, product_uuid is used as passphrase to cipher /config. This regression broke upgrade to Linux 3.18 for OpenXT installs using Linux 3.11.


jean-edouard:
This is pretty straight forward since it is a very small backport, in a file that's not patched elsewhere.
Will merge soon.

eric-ch:
This will also break upgrade from unfixed kernel to fixed ones (with this merged).

A quick work-around is to amend /etc/init.d/xenclient-config-access to not use get-config-key as keycmd, but "echo <the-wrong-dmi-product-uuid>". This dmi product_uuid is only used for test setups anyway.

richpersaud:
Could we document how to use this in testing, either in JIRA or the wiki?

jean-edouard:
Oh, not that straight forward then... I'll give this more time.

eric-ch:
@richpersaud I will write something to explain how to work-around the problem in both case. I could not find an existing ticket for this issue, but I am pretty sure someone mentioned it already... I might as well make a new one to make it easier to test.

eric-ch:
See openxt.atlassian.net/browse/OXT-465 for one possible work-around.

jean-edouard:
Thanks @eric-ch, will merge soon.

PR #203, opened 2016-01-26
Get rid of update-modules, since all it does is the bogus "depmod -ae"

jean-edouard:
just after every "depmod -a"


jean-edouard:
This need a build, I'll start one.

jean-edouard:
This worked before the jethro merge, but it may not make sense anymore, closing.

PR #205, opened 2016-02-09
Move labeling of /boot/system to selinux-load.sh.

stephensmalley:
There is no reason to do it prior to the switch_root, and
evidently this was not working correctly.  At the least, it would
have required passing -r /root to setfiles for an alternate root,
and even then it did not appear to work for some reason.  Just take
it to selinux-load.sh, where we can do it after loading policy
and use restorecon instead.

OXT-463

### DIFF
--- a/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
+++ b/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
@@ -154,12 +154,6 @@ boot_root() {
     mount --bind /proc /root/proc
     mount --move /sys /root/sys
 
-    # set selinux contexts for /boot/system on firstboot
-    if [ -f /root$FIRSTBOOT_FLAG ]; then
-        POL_NAME=$(sed -n 's&^[[:space:]]*SELINUXTYPE[[:space:]]*=[[:space:]]*\([A-Za-z0-9_-]\+\)[[:space:]]*$&\1&p' /root/etc/selinux/config)
-        echo "initramfs: setting file contexts for boot partition"
-        setfiles /root/etc/selinux/$POL_NAME/contexts/files/file_contexts /root/boot/system
-    fi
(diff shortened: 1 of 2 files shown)


jean-edouard:
I started a build with this as an override, I will test it tomorrow.

jean-edouard:
The build succeeded. I installed it and checked /boot/system, and most files are labelled right, except for the /boot/system/tpm directory, that gets labeled as boot_t instead of tcsd_var_lib_t (the file inside it is labelled right though).
Any idea why?

stephensmalley:
No, I'm afraid not.

jean-edouard:
Do you have a test machine handy that includes your modifications?
If you do, could you tell me what the label of /boot/system/tpm is?
I get:

    root@xenclient-dom0:~# ls -lZd /boot/system/tpm
    drwx------. 2 tss tss system_u:object_r:boot_t:s0 1024 Feb 25 11:21 /boot/system/tpm

Maybe it could be some kind of race condition...

stephensmalley:
I see the same output.  Yet running restorecon -Rv /boot/system by hand does relabel it.  So I don't currently have an explanation as to why it isn't being relabeled in the first place (assuming it exists at the point where init.sh performs its restorecon).

jean-edouard:
TL;DR: Will merge soon.

selinux-load.sh runs right between init.sh in initramfs and init.root-ro in the rootfs.

Then init.root-ro creates /boot/system/tpm (https://github.com/OpenXT/xenclient-oe/blob/master/recipes-openxt/xenclient/xenclient-root-ro/init.root-ro#L212).
I guess at that point the directory inherits the xattr from its parent directory, which is boot_t.

Bottom line, the labeling of /boot/system/tpm seems to be a separate issue, and I think this PR is valid and good to go, so I will merge it soon.

I created a ticket, please comment and feel free to take it over!
https://openxt.atlassian.net/browse/OXT-486

Thanks

PR #206, opened 2016-02-10
Fix neverallow failures the right way.

stephensmalley:
During the uprev to jethro, Adam encountered a set of neverallow failures
between neverallow rules in the reference policy and allow rules added by
OpenXT.  neverallow rules are build-time policy invariants that prevent
building the policy if any allow rule would allow a permission prohibited
by one of the neverallows. As a temporary workaround, Adam simply
commented out the neverallow rules in the refpolicy that were being
violated.

The allow rules that cause the violations are all introduced via the
*unconfined_readonly* interface calls added by OpenXT, where these
interface calls appear to allow read-only access to all objects of a given
class (e.g. all networks, devices, domains, files, etc).  This appears
to be a slightly scoped down version of unconfined domains in refpolicy,
just reduced from read-write to read-only access.  There are two domains
defined in the OpenXT policy patches that uses these interfaces:
unconfined_readonly_t and statusreport_t.

Rather than commenting out the neverallow rules altogether, this change
adds appropriate refpolicy interface calls to the definition of
unconfined_readonly_domain_noaudit() to exempt domains using it from
these neverallow rules.  This preserves the neverallow checking for all
other domains and avoids modifying neverallow rules in the refpolicy.

We should later investigate why these domains exist in OpenXT policy and
whether they truly need these permissions at all.


richpersaud:
Please create a JIRA ticket to document this workaround and the need for future investigation.

stephensmalley:
Added OXT-475

jean-edouard:
Builds fine, LGTM, merging soon.

PR #207, opened 2016-02-12
Avoid iPXE ROM init when not required

eric-ch:
Use the ROM PNP extension header to avoid loading the iPXE ROM
automatically if we have no BEV to boot from it.


richpersaud:
Doc question:  is it possible to PXE boot a VM on OpenXT?  If so, what config properties need to be set?

gamorg02:
Yes and add Network in the boot order after this commit.  Just need a PXE server to connect to.

rneilturner:
Tested with networking in the boot order and was able to PXE boot. Removed networking from the boot order and was not able to get a PXE boot (ctrl-b). LGTM. +1.

jean-edouard:
LGTM, merging soon.

PR #208, opened 2016-02-12
Move labeling of /boot/system to selinux-load.sh.

stephensmalley:
There is no reason to do it prior to the switch_root, and
evidently this was not working correctly.  At the least, it would
have required passing -r /root to setfiles for an alternate root,
and even then it did not appear to work for some reason.  Just take
it to selinux-load.sh, where we can do it after loading policy
and use restorecon instead.

OXT-463


jean-edouard:
Building now, will test on Monday.

jean-edouard:
Built successfully and behaves the same as in master.
Will merge soon.

PR #209, opened 2016-02-12
Delete old refpolicy recipes and patches.

stephensmalley:
These were for older versions or for policy variants that are
not used in OpenXT.


eric-ch:
+1, LGTM and builds fine.

jean-edouard:
+1, will merge soon.

PR #210, opened 2016-02-12
Fix test for unlabeled files.

stephensmalley:
The old test relied on ls -Z to check for unlabeled files.
However, commit 7e6b232828130514fd38091a34e29a4c3c12f105
(selinux: Fix do_install failures on libsepol/libselinux) added
a dependency on coreutils-native to ensure support for ln --relative,
and coreutils-native is built without SELinux support, so ls -Z
was always reporting "?".  It doesn't appear to be supported to
build coreutils-native with SELinux support as this creates a
dependency loop.  So just switch the test to use getfattr instead
(and we already have attr-native as a dependency).  Unfortunately
this also has some false positives due to the handling of hard links
by upstream pseudo xattr support, but at least it isn't every file.

### DIFF
--- a/classes/selinux-image.bbclass
+++ b/classes/selinux-image.bbclass
@@ -1,4 +1,4 @@
-DEPENDS += "policycoreutils-native"
+DEPENDS += "policycoreutils-native attr-native"
 IMAGE_INSTALL += " \
     libselinux-bin \
     policycoreutils-loadpolicy \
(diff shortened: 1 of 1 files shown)


eric-ch:
+1.

jean-edouard:
@stephensmalley, do you think it would make more sense to make libselinux and libsepol depend on `attr-native` instead of `coreutils-native`? I do...
That way maybe we can build avoid the potential circular dependency.

@stephensmalley, if `ls` was built properly, would it work better than `getfattr`?

stephensmalley:
The libselinux/libsepol dependency on coreutils-native is for ln --relative support, used by their Makefiles to create the .so symlinks.
If built properly, ls -Z will return the same result as getfattr; it will still be subject to the same problem with upstream pseudo xattr support not returning the same result on multiple hard links to the same file (only the link that was explicitly passed to setxattr will return a result).

jean-edouard:
Ok, thank you, I'll give this PR a quick test and merge soon.

jean-edouard:
Yup, 12854 warnings before, 1037 after, much better!
Merging very soon.

PR #211, opened 2016-02-16
Linux 4.4 DISTRO

eric-ch:
Upgrade OpenXT dom0 and service VMs to use Linux 4.4 kernel.
Introduce DISTRO=openxt-linux-4.4 to make testing easier.

OXT-476


rossphilipson:
We posted the basic plan on the ML here:
https://groups.google.com/forum/#!topic/openxt/3vdGW7G3hoc
Jed has the builders mostly ready so they can do DISTRO builds. I plan to merge this and the surfman change later today if there are no objections.

rossphilipson:
Jed ran a build with this distro with no problems. Merging now. Jed will merge in the surfman dep shortly.

PR #212, opened 2016-02-17
Port CVE-2015-7547 patch from upstream

eric-ch:
Source: https://sourceware.org/ml/libc-alpha/2016-02/msg00416.html

"During upstream review of the public open bug 18665 for glibc, it was
discovered that the bug could lead to a stack-based buffer overflow."

This ports git://sourceware.org/git/glibc.git:e9db92d


jean-edouard:
Builds fine, will merge soon.

PR #213, opened 2016-02-17
[usbback] Prevent usbback from trashing kernel memory. [OXT-481]

jctemkin:
This driver currently calls alloc_page() with flags that may return
lowmem *and* highmem pages. Throughout the remaining code,
pfn_to_kaddr() is used to produce virtual addresses for those pages, but
this function only produces valid addresses for lowmem pages. Calling it
for highmem pages produces an incorrect address that frequently lies
within the vmalloc range, potentially resulting in trashing another
piece of kernel memory. Requesting only lowmem pages prevents this from
happening.

### DIFF
--- a/recipes-kernel/linux/3.11/patches/usbback-base.patch
+++ b/recipes-kernel/linux/3.11/patches/usbback-base.patch
@@ -2347,7 +2347,7 @@ index 0000000..3f6bff2
 +
 +	for (i = 0; i < mmap_pages; i++) {
 +		pending_segments[i].grant_handle = USBBACK_INVALID_HANDLE;
-+		pending_pages[i] = alloc_page(GFP_KERNEL | __GFP_HIGHMEM);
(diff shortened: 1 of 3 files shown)


ktemkin:
+1

rossphilipson:
Does SV increase dom0 memory beyond say 950 or so?

rneilturner:
@rossphilipson Give me a call and I can answer the question, this would be an answer we couldn't give on this forum.

eric-ch:
+1

rossphilipson:
+1 since currently all dom0 memory is well within the logical lowmem range. I want to put a ticket though to perhaps enhance this if and when we move to a USB device domain. It might make sense to give such a domain more memory and use highmem to better support very large bulk or isoch transfers.

This should go in stable-5 too. Please create a ticket to fix it there (and if you have the time do a PR there too), thanks. If not, assign it to me.

I will merge it in a while.

PR #214, opened 2016-02-19
Linux 4.4 DISTRO for OpenXT stable-5

eric-ch:
Upgrade OpenXT dom0 and service VMs to use Linux 4.4 kernel.
Introduce DISTRO=openxt-linux-4.4 to make testing easier.

OXT-476


jean-edouard:
I suggest we merge this PR. Since it's a separate DISTRO, it won't disturb stable-5.
If a PR gets submitted later to make Linux 4.4 the default kernel in stable-5, then we'll test it.
How does that sound?

richpersaud:
If we merge without testing, we can indicate on the ticket what devices or features are targeted and require testing.

eric-ch:
@richpersaud I wrote a small comment to answer your request, to the best of my understanding, in the relevant ticket. This introduces a lots of changes and it can affect a large amount of things.

jean-edouard:
I will merge this tomorrow if there is no further objection.
Obviously, regular stable-5 builds will still be exactly the same, since this DISTRO will not be the default one.
I will however kick a custom build with this custom DISTRO for people who want to start testing.

PR #216, opened 2016-02-25
Fix glibc CVE-2015-7547 build break

eric-ch:
This reverts commit 91f999a2303b6d08232196799ae71fd3a1d03ce3.

What was I thinking...

glibc-2.22 follows upstream glibc repositories release/2.22/master, this
was obviously being patched upstream and has rightfuly ended up in the
release branch, breaking our unecessary patch.

If anything, with recipes tracking release branches/upstream repository,
we should submit the patch, through the relevant channel, to the
upstream project.


jean-edouard:
I can't believe OE does that in what they call a "release"...
Merging soon to fix the build...

cjp256:
That is because you are now tracking an upstream branch.  No different anywhere else... If you want to lock it in, you should be keying off a source revision for oe-core/meta-oe/etc instead of the branch.

eric-ch:
I agree with @cjp256, following a release branch is fine and more efficient to have security and bug fixes in asap. In this case, if we had a fix that looked upstream-able, we should have submitted it on libc-alpha ML directly.

Regarding build reproducibility, I would be very surprised if the ref of the release branch head used to build the package at time T is not written somewhere.

PR #218, opened 2016-02-26
xc-sha1sum redundant with busybox

eric-ch:
sha1sum is already provided in the initramfs by busybox, we should that
instead of having our own.


jean-edouard:
Oh yeah, we don't need that one anymore, since we just added busybox to the initramfs like 4 years ago :)

rossphilipson:
+1

jean-edouard:
`sha1sum` prints out the hash followed by the filename, as opposed to `xc-sha1sum` that just prints the hash.
It's probably preferable to use only the hash, so some cutting action might be needed here.

eric-ch:
@jean-edouard indeed... fixed. I'll test it correctly this time.

jean-edouard:
LGTM now, included in http://158.69.227.117/builds/master/custom-dev-14-master/
Will merge later today.

PR #219, opened 2016-02-26
QA warnings: Packaging

eric-ch:
This is only a starting effort to fix packaging in many of the xenclient-oe layer packages.

It makes a lots of noise mostly, making it harder to find important things in the output. So, deal with what is built in a better way:

- Split things into different packages (e.g, -dbg, -utils, -dev),
- do not leave things behind,
- try to make sure everything is where it belongs,
- Haskell binaries are stripped, silence the warnings about it,
- Missing RDEPENDS here and there...


jean-edouard:
LGTM, builds fine and doesn't seem to change the contents of the rootfs.
Will merge soon.

PR #220, opened 2016-02-26
QA warnings: Licences

eric-ch:
Those licenses are "Proprietary", we could make them available in the layer as well. Also there is more than one license applying to those packages.


rossphilipson:
+1

jean-edouard:
LGTM, builds fine, will merge soon.

PR #221, opened 2016-02-26
QA warnings: Linker

eric-ch:
Blktap does not need to override -fPIC.

OCAML-cross should generate position-independent code.
RPATH should not store absolute path from the build environment. Someone made a patch for that (http://caml.inria.fr/mantis/view.php?id=5943) as it is apparently not default.


jean-edouard:
LGTM, builds fine, will merge soon.

PR #222, opened 2016-02-27
Linux 4.4 missing bits

eric-ch:
@jctemkin found missing commits, present in 3.18 patch-queue but not in 4.4. Many thanks for reporting this :).

Apparently 5 of them, also missing from 4.1 (another PR will take care of that), mostly related to usbback fixes.

PR #224, opened 2016-02-29
Linux 4.1 missing bits

eric-ch:
Same as https://github.com/OpenXT/xenclient-oe/pull/222 for Linux 4.1


rossphilipson:
Same argument here: the 4.1 distro is already present and incomplete without these (even though we are not really using 4.1). Will merge in a while.

PR #225, opened 2016-02-29
Stable 5 linux 4.1 missing bits

eric-ch:
Commits went into 3.18 but were left out of 4.1.
Same as https://github.com/OpenXT/xenclient-oe/pull/224 in stable-5.


rossphilipson:
I will merge this in a while since the 4.1 distro is already present and incomplete without these (even though we are not really using 4.1).

PR #226, opened 2016-03-02
xenclient-licenses : Fix the path searched for control files.

dozylynx:
The directory containing package metadata information has moved
from /usr into /var in the current dom0 filesystem. This change
updates the licence collection step search path.

refs OXT-460

### DIFF
--- a/classes/xenclient-licences.bbclass
+++ b/classes/xenclient-licences.bbclass
@@ -34,7 +34,14 @@ python do_licences() {
 
         package = {}
 
+        # Supply an empty default value for optional field:
+        package['Homepage'] = ''
+
         for line in file:
+            if line and line[0] == ' ':
+                # Only take the first line of multi-line fields
+                continue
+
             key, value = split_re.split(line.rstrip(), 1)
             if key in fields:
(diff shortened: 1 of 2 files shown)


dozylynx:
Hold off on pulling this one in just yet. I still think it's correct but it looks like it will need another change to accompany it.

dozylynx:
OK, this is ready to go.
It fixes the license extraction for:
* dom0
* ndvm
* syncvm
* uivm

while sysroot, installer and installer-part2 still produce empty license spreadsheets. This is most likely because those filesystems do not contain the metadata about their packages. This should be addressed one way or another, but no need to block this PR for that.

jean-edouard:
Building now, will test on Monday.

jean-edouard:
Whoops, looks like you added a commit to the PR branch.
New commits get automatically added to PRs, which is why it is recommended to use a separate branch for pull requests.

dozylynx:
That commit's fine as part of this PR.

jean-edouard:
Builds fine, licenses are populated, will merge soon.

PR #227, opened 2016-03-03
Simplify xenclient-oe's selinux-image.bbclass.

stephensmalley:
Phil upstreamed a subset of the xenclient-oe selinux-image.bbclass file
to meta-selinux, in particular the portion dealing with labeling
the rootfs.  Consequently, at present, both meta-selinux and xenclient-oe
define a selinux-image.bbclass file with different, overlapping definitions.
Rename the xenclient-oe .bbclass to openxt-selinux-image, change
openxt-selinux-image to inherit from selinux-image, and drop the
mostly redundant selinux_set_labels definition and usage from
openxt-selinux-image.  The only thing we lose in using the meta-selinux
definition of selinux_set_labels are the warnings about unlabeled files,
which are not even accurate due to the previously noted pseudo xattr
issue (see commit eea606f5bf3350b92c16a3fbe7ac6d47b4fe8cbb).  If we truly
want such checking, we can revive it via a separate function in
openxt-selinux-image but we need a reliable way of checking it.

Also drop the selinux_policy_link definition and usage from
openxt-selinux-image.  The policy is linked as part of the refpolicy-mcs
build, so this is completely unnecessary and actually incorrectly
generates a policy.29 policy file in addition to the policy.28 file
shipped in refpolicy-mcs since it does not set policy-version in its
temporarily generated semanage.conf file.


jean-edouard:
LGTM, removing custom code is always nice!
Building now.

jean-edouard:
Builds and boots fine, will merge soon.

PR #228, opened 2016-03-03
Drop legacy cruft from refpolicy recipe.

stephensmalley:
Linux 3.0 introduced /sys/fs/selinux as the preferred mount point
for selinuxfs instead of /selinux, so /selinux is no longer needed.

There are no ${datadir}/selinux/*/*.bz2 files to package, and this
directory is handled by the meta-selinux refpolicy recipe anyway, so
drop it.

The EXTRA_OEMAKE flags seem be some legacy of older recipes; removing
them seems to have no effect.


jean-edouard:
Makes sense, building now.

jean-edouard:
Builds and boots fine, will merge soon.

PR #229, opened 2016-03-03
Fix labeling of files/directories created on firstboot.

stephensmalley:
/boot/system/tpm is created via install -d from init.root-ro.
coreutils install is supposed to look up the proper security context
and assign it but there seems to be a regression for directory creation,
leaving it in the parent directory type.

Several /config files are created by firstboot.sh via shell redirection.
These will default to the type of the parent directory without an explicit
label assignment.

Add restore calls as appropriate (restore is defined as a shell
wrapper for restorecon in /etc/init.d/functions).
The alternative would be to use name-based type transitions in policy,
but explicit restorecon seems more appropriate here given that these
are being created manually by an init script and are essentially one-time
operations.

OXT-486

### DIFF
--- a/recipes-openxt/xenclient/xenclient-dom0-tweaks/firstboot.sh
+++ b/recipes-openxt/xenclient/xenclient-dom0-tweaks/firstboot.sh
@@ -99,6 +99,7 @@ fi
 if [ -r ${INSTALL_CONF}/keyboard.conf ] ; then
     KEYBOARD="$(awk -F\' '/^KEYBOARD=/ { print $2 }' ${INSTALL_CONF}/keyboard.conf)"
     echo "KEYBOARD='${KEYBOARD}'" > /config/keyboard.conf
+    restore /config/keyboard.conf
  
     DEFER_KEYBOARD="$(awk -F\' '/^DEFER_KEYBOARD=/ { print $2 }' ${INSTALL_CONF}/keyboard.conf)"
     if [ "${DEFER_KEYBOARD}" = "true" ]; then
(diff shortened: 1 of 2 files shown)


jean-edouard:
So we actually do have to relabel every file we create...
LGTM, building now.

stephensmalley:
Just to clarify: If these were runtime files created by a daemon or application program and if they needed to be labeled differently than their parent directory, then we'd normally define type transition rules in the policy and the kernel would apply the new type.  But since they are just being created on first boot and from a shell script that isn't running in any unique domain of its own (probably just initrc_t), it seems more practical to explicitly relabel them.

jean-edouard:
Oh that makes sense, thank you for the clarification!

jean-edouard:
So, even with this fix in, /boot/system/tpm still ends up in the wrong domain (boot_t).
Is it expected?

stephensmalley:
init.sh only executes init.root-ro if measured is passed on the kernel cmdline.
And init.root-ro only creates it and runs restore if it doesn't already exist.
So, to test, I either make sure I manually pass measured on the first boot after install or I rm -rf /boot/system/tpm and boot it with measured on the cmdline (couldn't get it to enable measured boot for real on my current test machine).  Then it gets the right label.  Not really clear on where it gets created in the case of a non-measured boot.  But I don't see any denials for it either way.

stephensmalley:
BTW, there is still one lingering mislabeled file that I haven't figured out - /config/xenclient.conf.  Labeled db_backup_t instead of xc_config_t, which is weird because the parent directory is xc_config_t.  /config/backups is labeled db_backup_t, so I was wondering if it is created there and then moved or something.  I haven't yet found where it is being created.

stephensmalley:
Oh, I see - manager/upgrade-db/Upgrade.hs copies /etc/xenclient.conf to /config/xenclient.conf, and the policy defines a type transition for db_upgrade_t on xc_config_t to db_backup_t, so any files created by upgrade-db in /config default to db_backup_t.  Don't know if that's a problem or not.  Can probably fix that one in policy if it matters.

jean-edouard:
Tested this properly and got the expected result.
Will merge soon.

PR #230, opened 2016-03-04
refpolicy: policy fixes

stephensmalley:
Address various denials seen in dom0 when booting OpenXT.
Each patch includes its own description.


jean-edouard:
All these permissions seem fair.

I wish i915modprobe.sh was in a more specific domain, to avoid `files_exec_usr_files(udev_t)`.

I guess that's typically the kind of decision that's hard to make without a proper security plan...

I'll merge this next week if there's no objection (in other words, if someone thinks udev shouldn't be allowed to execute any /usr binary, please chime in!).

rossphilipson:
I don't think we need i915modprobe.sh anymore. I added that back in the days when we had our proprietary igfx driver and it was used to switch between that and the drm-plugin. I would make a ticket to remove it and I think there is a udev rule that goes with it.

jean-edouard:
Well perfect then, let's remove that script next week before merging this PR.
@stephensmalley please remove `files_exec_usr_files(udev_t)` from this PR, unless it's needed for something else.

rossphilipson:
I will do a PR now - I found the original commits and know what to remove. We can fire up a build to make sure everything is happy.

rossphilipson:
PR: https://github.com/OpenXT/xenclient-oe/pull/231

jean-edouard:
After looking more closely at your commit, I realized you're adding a 1-line patch for each thing you fix.
There's not really any established rule about how patchqueues should look like, but don't you think we'll end up with an unmanageable amount of patches by doing this?

rossphilipson:
I kind of agree with @jean-edouard - we have had patch queues grow enormous and un-maintainable in the past. Maybe this is a good time to think about how to re-organize the entire patch queue since are there are many many patches that are for just 1 new file in the policy.

jean-edouard:
I suggest using patches only for existing refpolicy files, and having all new files as plain files.
The directory structure in xenclient-oe could match the refpolicy one, and all the patches would be in their own directory.

For example, some of the files would be:

    recipes-security/refpolicy/refpolicy-mcs-2.20140311/policy/modules/services/dbd.te
    recipes-security/refpolicy/refpolicy-mcs-2.20140311/policy/modules/apps/xec.fc
    recipes-security/refpolicy/refpolicy-mcs-2.20140311/patches/policy.modules.contrib.qemu.diff
    recipes-security/refpolicy/refpolicy-mcs-2.20140311/patches/policy.modules.contrib.xen.diff

I'm willing to do the change if it works for everybody (including the contents of this PR).
That way, no need to modify patch files for OpenXT specific bits anymore.
Also, in the future, patches against refpolicy could be more feature-oriented rather than module based...
How does that sound?

stephensmalley:
Sounds good to me.

richpersaud:
This topic deserves a wiki page and/or mailing list thread to guide contributors on our preferred approach, taking into account optional layers for specific use cases.

jean-edouard:
Agreed, I'll start with a pull request to implement my suggestion, and if/when it gets merged I'll try to write down some patchqueue guidelines.

jean-edouard:
See https://github.com/OpenXT/xenclient-oe/pull/235
Please review.

jean-edouard:
First wiki page draft: https://openxt.atlassian.net/wiki/display/~jed/OpenXT+patchqueue+guidelines
Please review.

stephensmalley:
Patchqueue guidelines make sense to me.

jean-edouard:
https://github.com/OpenXT/xenclient-oe/pull/235 went in, and breaks this PR, as expected...
@stephensmalley if you want me to, I'll be happy to rewrite your PR on top of the new layout, since I'm the one who broke it.

stephensmalley:
See if the one I just pushed is what you wanted.

jean-edouard:
Yup! Thank you!
I will merge this soon, since it was basically already reviewed.

PR #231, opened 2016-03-04
Remove the DRM enable experimental plugin support and i915 modprobe b…

rossphilipson:
…locker.


eric-ch:
+1, will merge on monday if nobody opposes.

rossphilipson:
I think Jed started a build with it and just wanted to smoke test build/install before merging. But yea it is dead functionality and should go. We should spend some time trying to think of other dead stuffs we can rip out related to the old igfx days.

jean-edouard:
Builds, installs, boots. Will merge soon.

PR #232, opened 2016-03-07
Remove ioemu, as the repo is gone now, so it breaks the build scripts

jean-edouard:


jean-edouard:
More info on that:
Everytime we call bitbake, it runs checks on all the OpenXT git repositories from all the recipes.
If one check fails, bitbake dies
When building directly from github (not recommended), github.com/OpenXT/ioemu resolves to github.com/OpenXT-Extras/ioemu, so the check on ioemu passes.
However, the build scripts lists all the repos in github.com/OpenXT and mirrors them. ioemu is not part of that list, so it doesn't get mirrored and the build fails.

rossphilipson:
Nice you overloaded github with the shear amount of changes. It cannot display all you changes :)

I am fine with removing all these dead files as long as no one else wants something done with them (they are in history after all).

cjp256:
Cleanup is good. LGTM, merging.

PR #233, opened 2016-03-07
Remove the status-server tool and its instructions in status-report

dozylynx:
fixes OXT-498


rossphilipson:
Are there any concerns with merging this PR and closing the ticket out?

dozylynx:
No concerns from me.

richpersaud:
Are we moving all dormant/inactive components to the OpenXT-extras repo, e.g. all the tickets under https://openxt.atlassian.net/browse/OXT-500?  If not, we should document which components are appropriate for OpenXT-extras.

rossphilipson:
OpenXT-Extras is not a repo, it is an organization. Repositories that are no longer in use in OpenXT get moved there (though that is not its only use/purpose). The main reason for this (other than historical value) is to allow previous version of OpenXT to build via the Github mechanism to cache prior URLs to repositories.

jean-edouard:
+1

rossphilipson:
+1 builds and installs - I say merge if my answer above is satisfactory for @richpersaud

richpersaud:
Thanks for the OpenXT-extras org explanation, ok to merge.

PR #234, opened 2016-03-08
libselinux: Apply two upstream fixes required for installer.

stephensmalley:
selinux upstream commits 9df4988 and 5a8d8c4.
Otherwise, when switching /etc/selinux/config to enforcing for dom0,
the installer fails when attempting to mount selinuxfs, with:
    Mount failed for selinuxfs on /sys/fs/selinux:  No such file or directory

The scenario is that the installer is booted with selinux=0, but
has the same /etc/selinux/config as dom0, and thus has SELINUX=enforcing
if we make dom0 enforcing.  The expected behavior is that kernel
parameters take precedence (override) the /etc/selinux/config setting,
so this used to work.  When the preferred selinuxfs mount point moved from
/selinux to /sys/fs/selinux (in Linux 3.0), attempting to mount selinuxfs
on a kernel with SELinux disabled (via selinux=0 or via kernel config)
started returning ENOENT rather than ENODEV because the /sys/fs/selinux
mount point is only created by the kernel if SELinux is enabled.  This
then confused libselinux into thinking that the mount failed for some
reason other than SELinux being disabled in the kernel, and caused it
to fail-closed, thereby causing init to exit and the system to panic.
An early fix for this was b3b19fdce, which checked /proc/filesystems
to see if SELinux was disabled (i.e. no selinuxfs entry).  But this
fails if init has not already mounted /proc, which seems to be the case
for sysvinit.

Commit 9df4988 fixed this issue by mounting /proc before reading /proc
filesystems, but triggers unnecessary attempts to mount /proc for other
users of libselinux besides init.  Commit 5a8d8c4 fixed that issue by
making the mounting of /proc conditional on it not already being mounted.

Arguably if the installer is going to run with SELinux disabled, then
it shouldn't have any /etc/selinux/config or other SELinux policy files
present, but that is left to a separate change.

### DIFF
--- a/recipes-security/selinux/libselinux/libselinux-mount-procfs-before-check.patch
+++ b/recipes-security/selinux/libselinux/libselinux-mount-procfs-before-check.patch
@@ -1,0 +1,74 @@
+commit 9df498884665d79474b79f0f30d1cd67df11bd3e
+Author: Ben Shelton <ben.shelton@ni.com>
+Date:   Wed Apr 15 15:56:57 2015 -0500
+
+    libselinux: Mount procfs before checking /proc/filesystems
+    
+    In the case where the SELinux security module is not loaded in the
+    kernel and it's early enough in the boot process that /proc has not yet
+    been mounted, selinuxfs_exists() will incorrectly return 1, and
+    selinux_init_load_policy() will print a message like this to the
+    console:
+    
+    Mount failed for selinuxfs on /sys/fs/selinux:  No such file or directory
+    
+    To fix this, mount the procfs before attempting to open
+    /proc/filesystems, and unmount it when done if it was initially not
+    mounted.  This is the same thing that selinux_init_load_policy() does
+    when reading /proc/cmdline.
+    
+    Signed-off-by: Ben Shelton <ben.shelton@ni.com>
+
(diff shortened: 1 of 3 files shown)


jean-edouard:
Thanks for the detailed description!
Wouldn't it make sense to upgrade our SELinux libs?
It sounds like kernel upgrades call for SELinux ones too...

jean-edouard:
Also, why doesn't this seem to be a problem in stable-5?
It's based on Linux 3.18 and in enforcing mode by default.

stephensmalley:
I think there we are falling back to trying to mount on /selinux after failing the /sys/fs/selinux mount, and then we get ENODEV and libselinux does the right thing.  So the problem only shows up after removing the no-longer-used /selinux mount point, which I did in f9b758615.

I submitted patches to meta-selinux yesterday to upgrade their selinux userspace so we could look to moving to that in the future, but presently we're tied to a specific commit and not sure whether that upgrade would get cherrypicked to jethro.

jean-edouard:
Understood.

I didn't realize those changes were that recent!

I will test this PR soon.

jean-edouard:
Builds, installs, boots, merging soon.

PR #235, opened 2016-03-08
refpolicy-mcs: replace OpenXT file creation patches by the files themselves

jean-edouard:


jean-edouard:
The patch is pretty big. For review purposes, it might be easier to visually compare these two:
https://github.com/OpenXT/xenclient-oe/tree/master/recipes-security/refpolicy
https://github.com/jean-edouard/xenclient-oe/tree/refactor_refpolicy/recipes-security/refpolicy

According to my tests, the end results (the patched refpolicies) are exactly the same, but don't take my word for it!

stephensmalley:
I see a difference in alsa.te:

-optional_policy(`
-	statusreport_create_tmp_files(alsa_t)
-	statusreport_write_tmp_files(alsa_t)
-	statusreport_getattr_tmp_files(alsa_t)
-')
-

Seems to have been dropped from statusreport-interfaces.diff.

jean-edouard:
My bad, this chunk was giving me trouble so I removed it and forgot to put it back...
Good catch, thanks!
Commit amended.

stephensmalley:
Ok, now I get the same result from building both.
Also looks like the split is correct.
We may want to migrate handling of modules.conf (from patch to separately managed file) at some point; for example, Red Hat ships their own complete modules.conf with a specific list of policy modules they want included for each of their policy variants and just drops that into the refpolicy directory.  Otherwise you may be picking up additional modules from refpolicy each time you rebase without really wanting them in your policy.

rossphilipson:
@stephensmalley that sounds like wise advice and a good candidate for a ticket.

stephensmalley:
+1

eric-ch:
LGTM, +1.
Will merge soon.

PR #236, opened 2016-03-09
Remove syslogd from our busybox config

jean-edouard:
This looks like a huge amount of changes, but in the end all it should do is remove syslogd from busybox. Most of those changes were happening automatically when the busybox build system was preparing the config.

Here's all I did:
- Build busybox
- Go to its work directory
- make menuconfig
- uncheck syslogd and klogd
- save and exit

This needs a build, I'll kick one up.

Note: this introduces 2 harmless build warnings that I could not get rid of:

    WARNING: busybox: NOT adding alternative provide /etc/init.d/syslog: /etc/init.d/syslog.busybox does not exist
    WARNING: busybox: NOT adding alternative provide /etc/syslog-startup.conf: /etc/syslog-startup.conf.busybox does not exist

OXT-502


jean-edouard:
With this PR in, OpenXT hangs on the "Loading..." splash screen. Closing for now.

jean-edouard:
This PR actually works... I was just running into this more consistently than usual: https://openxt.atlassian.net/browse/OXT-504

jean-edouard:
Here's the build: http://openxt.ainfosec.com/builds/master/custom-dev-20-master/

stephensmalley:
Works for me, and avc denials no longer truncated. +1

eric-ch:
+1 as well. Will definitely have a look at OXT-504.

jean-edouard:
Thank you both for the tests!
@eric-ch good luck with OXT-504, I could be wrong but it looks like a xenfb2 bug...

PR #237, opened 2016-03-10
[xen-tools] Packaging.

eric-ch:
Put out some QA warnings and stop shipping things that cannot run on the
target.
Flask is not required either, since the xsm policy is built/shipped
separately.
meta-virtualization recipe is much nicer, but cannot be used as is just
yet.

### DIFF
--- a/recipes-extended/xen/xen-tools.bb
+++ b/recipes-extended/xen/xen-tools.bb
@@ -12,7 +12,7 @@ SRC_URI += "file://xenstored.initscript \
 DEPENDS += " gettext ncurses openssl python zlib seabios ipxe gmp lzo glib-2.0 iasl-native xz "
 DEPENDS += "util-linux"
 # lzo2 required by libxenguest.
-RDEPENDS_${PN} += " lzo"
+RDEPENDS_${PN} += " lzo bash"
 
 PACKAGES = "${PN}-libxenstore ${PN}-libxenstore-dev ${PN}-libxenstore-dbg ${PN}-libxenstore-staticdev   \
             ${PN}-xenstore-utils ${PN}-xenstore-utils-dbg                                               \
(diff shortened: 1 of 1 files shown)


jean-edouard:
As mentioned in private on IRC, I'm not a big fan of `+RDEPENDS_${PN}-xentrace += " python"`...
I think we should remove python dependencies, not add some, so that one day maybe we can stop building it!

eric-ch:
@jean-edouard Makes sense, I think only the xentrace_format tool requires python. And that is not a useful thing to deploy.

jean-edouard:
So, if I understand correctly, right now nothing depends on xen-tools-xentrace, so it won't be installed, and the python dependency will not be built, right?

eric-ch:
xen-tools-xentrace will not be installed in dom0 image by default, but python will definitely be built.

jean-edouard:
Right, because other recipes depend on it, but not because of `+RDEPENDS_${PN}-xentrace += " python"`, correct?

eric-ch:
`RDEPENDS_${PN}-xentrace += "python"` will require `do_package_write` to be ran for python before xen-tools package can run `do_build` (at least that's how I understood it from : https://lists.yoctoproject.org/pipermail/yocto/2013-August/015792.html).
So that RDEPENDS will gets up python package built, even if we do not deploy it afterwards.

jean-edouard:
Then it's a -1 for me.
We should not increase our python dependency count.
I'd be interested to know what other people think about this...

eric-ch:
Since the xenclient-oe Xen recipe tries to skim what is not shipped, it could as well be modified to not build what will not be deployed, removing as much dependencies as possible.

rossphilipson:
@eric-ch what is the status of this PR?

eric-ch:
@rossphilipson I left it behind for a while... I'll fix it up to remove the python dependency.

PR #238, opened 2016-03-10
GHC QA warnings

eric-ch:
Add GHC License in the layer so it is provided for relevant packages.

INSANE_SKIP installed-vs-shipped:
Currently, install puts files in staging absolute path, making this very noisy in the logs. The fix should be to avoid that, the very loud logs are reporting a side effect of the real issue.


jean-edouard:
LGTM.
No other ghc package is GHCL?

eric-ch:
Other Haskell libraries are (ghc-network, ghc-deepseq, ghc-mtl, ghc-transformers, ...), and already have LICENCE = "GHCL". I found no other Glas[cg]ow (no kidding :)) mentioned.

jean-edouard:
Hah, didn't see the "Glascow" at first, nice one :)
Will merge soon.

PR #239, opened 2016-03-10
[conf] xenclient-customtasks

eric-ch:
xenclient-customtasks.bbclass introduced two tasks for xenclient
recipes:
- do_force_rebuild: Remove stamps file of any tasks not in a whitelist
    Currently, that will end up doing "bitbake -c clean <target>" in the
    end.
- do_makeclean: Call "oe_runmake clean", and depends on force_rebuild
    making it a bit misleading...

The first really looks like "bitbake -C compile <task>"
I am not sure what the second provides, it is fine to do make clean in
devshell.

### DIFF
--- a/classes/xenclient-customtask.bbclass
+++ b/classes/xenclient-customtask.bbclass
@@ -1,57 +1,27 @@
-# This class contains convenience tasks for working with XenClient
+#
+# Convenience tasks for corner cases.
+#
 
-# makeclean:
-# Run "make clean" and remove the do_compile stamp if it exists
-
-__do_xcmakeclean() {
+__makeclean() {
     oe_runmake clean
 }
-
-FORCE_REBUILD_TASK_WHITELIST = "do_fetch do_unpack do_unpack_xc_repos do_patch do_apply_patchqueue do_populate_lic do_configure"
-
 python do_makeclean() {
-    from bb import note
-    note("Running make clean")
-    try:
-        bb.build.exec_func("__do_xcmakeclean", d)
-    except bb.build.FuncFailed:
-        pass
+    workdir = d.getVar('B', True)
+    makefiles = [ '/Makefile', '/GNUmakefile', '/makefile' ]
+    if reduce(lambda x, y: x or os.path.exists(workdir + y), makefiles, False):
+        bb.build.exec_func('__makeclean', d)
+        bb.build.write_taint('do_compile' , d)
+        sstate_clean_cachefiles(d)
 }
-python do_force_rebuild() {
-    from bb import note,build
-    from os import unlink, listdir
-    from glob import glob
+addtask do_makeclean
+do_makeclean[nostamp] = "1"
+do_makeclean[doc] = "Run `make clean', if applicable and mark do_compile as tainted."
 
-    # clean stamps...
-    task_whitelist = d.getVar("FORCE_REBUILD_TASK_WHITELIST", True).split()
-    stampglob = bb.data.expand('${STAMP}.*', d)
-    for stamp in glob(stampglob):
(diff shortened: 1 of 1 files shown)


jean-edouard:
-1, my workflow is pretty much based on these two steps...

I don't know do_clean, so maybe it's the same as force_rebuild and we could remove force_rebuild, but no makeclean would cause a lot of overhead by requiring to use devshell, which doesn't always work great...

cjp256:
@jean-edouard @rossphilipson 

now that jethro is in, does the workflow I previous outlined in https://openxt.atlassian.net/browse/OXT-455 work for you?

jean-edouard:
Seems like `./bb -c configure -f` could replace force_rebuild, but I really like to stay away from devshell, so makeclean is kind of a must...
Is there a motivation for removing these steps other than cleanup?

eric-ch:
@cjp256 Any use of force_rebuild (on its own or because makeclean was called and then will enforce force_rebuild afterward) will trigger and do_clean, effectively removing the sources in the workdir and any unsaved work in there ...
@jean-edouard ... To me this is a trap :), so that is the main incentive behind that removal.

./bb -c configure -f will only rerun configure and mark the step as tainted.
./bb -C configure will rerun configure, mark it tainted and run the default task?

We could probably have do_makeclean back in (and mark compile as tainted as a result instead of calling force_rebuild), though if this is required, there is likely something to fix in the build system of the project compared to just "./bb -c compile -f" or "./bb -C compile".

jean-edouard:
@eric-ch force_rebuild does nothing to the workdir, it just takes the "stamps" back to the configure step.

cjp256:
AFAICT, bitbake -c configure -f <pkg> should perform clean on relevant packages.

https://github.com/openembedded/openembedded-core/blob/master/meta/classes/base.bbclass#L289

I would imagine you can skip to -c compile or something if you didn't want that behaviour?

eric-ch:
@jean-edouard force_rebuild does not do anything to the workdir, but by doing something inadequate with the stamps files, it triggers "something" that looks really like a do_clean. Try it and be prepared.

@cjp256 oh! indeed, so that disqualifies do_makeclean :).

jean-edouard:
force_rebuild is indeed broken, I still want to keep makeclean

eric-ch:
I guess there is no harm in having those tasks around, and if people find those convenient we might as well fix them.
Closing related PR https://github.com/OpenXT/openxt/pull/110 as well.

jean-edouard:
In your new version, you left this off: `do_makeclean[depends] = "${PN}:do_force_rebuild"`.
Is it on purpose?

jean-edouard:
Thanks a lot by the way for fixing these steps even when you don't seem to have a use for it :)
Much appreciated!

eric-ch:
My last attempt partially wrong and definitely done backward (force_rebuild => makeclean instead of the opposite)...
`do_force_rebuild` will mark the `do_compile` task as tainted and clear the sstate.
`do_makeclean` will run `make clean`, mark `do_compile` task as tainted and clear the sstate.
... As it used to do.

`do_makeclean[depends] = "${PN}:do_force_rebuild"` is not required as I would like to avoid tainting a `do_compile` task if running `make clean` is irrelevant for the target (no Makefile).

jean-edouard:
I will test this.
In the meantime, would it be possible to either fix the commit short descriptions or merge all three into one? Right now, all 3 are titled "[conf] xenclient-customtasks".

eric-ch:
@jean-edouard It is now one commit.

jean-edouard:
Thank you, I tested it, it worked fine, will merge soon.

PR #240, opened 2016-03-11
grub: mount / and /boot/system as ext3

jean-edouard:
After the jethro merge, / and /boot/system get mounted as ext2 instead of ext3.
I suspect it is due to a change in busybox, that now somehow defaults to ext2.
According to the man page (http://linux.die.net/man/1/busybox), mount will:

    Mount a filesystem. Filesystem autodetection requires /proc be mounted.

Since the partition are mounted by busybox in the initramfs, /proc is probably missing at that point...

The initscript can use kernel options for choosing the filesystem, so I'm adding those to the Linux command line in Grub.

Maybe at some point we'll want the initscript to be able to figure things out by itself...

### DIFF
--- a/recipes-openxt/xenclient/xenclient-dom0-tweaks/grub.cfg
+++ b/recipes-openxt/xenclient/xenclient-dom0-tweaks/grub.cfg
@@ -30,7 +30,7 @@ fi
 
 TBOOT_COMMON_CMD="min_ram=0x2000000 loglvl=all serial=115200,8n1,0x3f8 logging=serial,memory"
 XEN_COMMON_CMD="console=com1 dom0_mem=min:320M,max:320M,320M com1=115200,8n1,pci mbi-video vga=current flask_enforcing=1 loglvl=debug guest_loglvl=debug"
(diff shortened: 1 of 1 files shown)


jean-edouard:
Actually /proc should be there, unless it failed to mount...
Either way, I think this PR is the right thing to do, at least for now.

richpersaud:
Can we use at least one ticket number to link Jethro-related cleanup PRs?  Or separate tickets under the Jethro epic?

jean-edouard:
Technically I'm just fixing a bug in master... I'm not sure how the fact that it appeared right after the Jethro merge is useful information...

richpersaud:
It would be useful to have somewhere (wiki page, ticket) that links PRs related to busybox and jethro, and your notes from https://openxt.atlassian.net/wiki/pages/viewpage.action?spaceKey=~jed&title=Busybox+overview .  If you want, I can create a busybox ticket under the jethro OE upgrade epic.

jean-edouard:
What I mean is that there's no link between this issue and jethro.

stephensmalley:
Tested, LGTM, +1

jean-edouard:

rossphilipson:
+1 if it builds and installs - merging

PR #241, opened 2016-03-14
[dm-agent] Move patches to the repository.

eric-ch:
dm-agent patches were left in the layer. They belong to the dm-agent repository.


jean-edouard:
This and https://github.com/OpenXT/dm-agent/pull/8 LGTM.
A quick visual diff showed no code difference, I don't think we need more than that.
Merging soon.

PR #245, opened 2016-03-18
Dom0 policy fixes

stephensmalley:
Various fixes for SELinux denials seen in dom0.

PR #251, opened 2016-03-18
Ndvm policy fixes

stephensmalley:
Policy fixes for denials in the ndvm.

### DIFF
--- a/recipes-security/refpolicy/refpolicy-mcs-2.20140311/patches/policy.modules.conf.patch
+++ b/recipes-security/refpolicy/refpolicy-mcs-2.20140311/patches/policy.modules.conf.patch
@@ -1,7 +1,7 @@
-diff --git a/policy/modules.conf b/policy/modules.conf
-index 66ae3ea..d533408 100644
---- a/policy/modules.conf
-+++ b/policy/modules.conf
+Index: refpolicy/policy/modules.conf
+===================================================================
+--- refpolicy.orig/policy/modules.conf
++++ refpolicy/policy/modules.conf
 @@ -149,7 +149,7 @@ su = module
  #
  # Execute a command with a substitute user
@@ -904,7 +904,7 @@ index 66ae3ea..d533408 100644
  
  # Layer: contrib
  # Module: mandb
-@@ -1318,56 +1318,56 @@ mandb = module
+@@ -1318,42 +1318,42 @@ mandb = module
  #
  # Linux hardware error daemon.
  # 
@@ -953,14 +953,7 @@ index 66ae3ea..d533408 100644
  
  # Layer: contrib
  # Module: modemmanager
- #
- # Provides a DBus interface to communicate with mobile broadband (GSM, CDMA, UMTS, ...) cards.
- # 
--modemmanager = module
-+modemmanager = off
- 
- # Layer: contrib
- # Module: mojomojo
+@@ -1367,7 +1367,7 @@ modemmanager = module
  #
  # MojoMojo Wiki.
  # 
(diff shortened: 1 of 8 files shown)


jean-edouard:
LGTM.
I started a build with this + the dom0 fixes + the enforcing PRs.
I'll merge all 4 PRs if the build succeeds, installs and runs fine.

jean-edouard:
The build is green, and the installation works.
I keep running into OXT-504, which doesn't help testing...
But after 4 reboots I managed to get a UI!

I'm getting only one AVC (!), 14 times in a row when NDVM starts:

    Mar 22 15:47:45.471605 xenconsoled: Network (1): [    2.275046] audit: type=1400 audit(1458661665.463:4): avc:  denied  { ipc_lock } for  pid=508 comm="db-cmd" capability=14  scontext=system_u:system_r:network_slave_t:s0 tcontext=system_u:system_r:network_slave_t:s0 tclass=capability permissive=0

Do you know about it?

stephensmalley:
No.  kernel version?  I'm using 4.4.0.

jean-edouard:
I'm using OpenXT as it is, so 3.18.25.

stephensmalley:
Ok, I think this is http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=a5a6579db33af91f4f5134e14be758dc71c1b694
So I believe we can ignore it, or at most add a dontaudit of it to silence it on 3.18.

jean-edouard:
Yeah, we should add a dontaudit just to keep /var/log/messages cleaner, since it seems to happen more than just a few times on startup. I'll write a ticket.

jean-edouard:
Great work on the policy fixes, I never thought I'd be so happy to see SELinux enforced on my laptop :)

I think these pull requests are good to go:
https://github.com/OpenXT/xenclient-oe/pull/245
https://github.com/OpenXT/xenclient-oe/pull/251
https://github.com/OpenXT/xenclient-oe/pull/255
https://github.com/OpenXT/xenclient-oe/pull/256

I'll merge them tomorrow if there's no objection.

Here's the build that includes all 4 PRs:
http://158.69.127.133:8020/builders/openxt-custom/builds/23
http://openxt.ainfosec.com/builds/master/custom-dev-23-master/

stephensmalley:
I have a fix queued up for the ipc_lock noise at https://github.com/stephensmalley/xenclient-oe/commit/80f33af00d053eb5d2a3899afdd6885513771712.  Will submit after these are merged.
The other alternative would be to back-port the kernel patch to 3.18.
Also, was unclear on which kernel I should be testing against - I had the impression that 4.4 was the target kernel for the 6.0 release, so I had previously switched to building with DISTRO=openxt-linux-4.4.  Do you envision keeping it optional for the final 6.0 release?

oxtbuilder:
That's a good point, I'm pretty sure we'll want Linux 4.4 for OpenXT 6.0.
In general however, I believe a pull request against a given OpenXT branch should target whatever components are currently built by default for that branch.

In this case, the delta seems to be, at least for now, just that ipc_lock AVC, so it shouldn't be a big deal either way.

Maybe we should start a thread on the mailing list about switching master to 4.4...

Merging this in a few minutes.

jean-edouard:
That's a good point, I'm pretty sure we'll want Linux 4.4 for OpenXT 6.0.
In general however, I believe a pull request against a given OpenXT branch should target whatever components are currently built by default for that branch.

In this case, the delta seems to be, at least for now, just that ipc_lock AVC, so it shouldn't be a big deal either way.

Maybe we should start a thread on the mailing list about switching master to 4.4...

Merging this in a few minutes.

PR #252, opened 2016-03-19
[blktap] Review, cleanup and remove existing blktap2 patches in the X…

rossphilipson:
…en PQ

The Xen PQ currently has patches in it from XenClient Enterprise that OpenXT
is not using (because we are using our own blktap2 fork at the moment). This
PR removes most of those patches since they are related to XCE features we
don't use. A couple of patches that looked like legit fixes were kept.

Note the xen-tools.bb recipe is now building blktap2 though not installing
it. That will come later when blktap.git goes away.

OXT-511


jean-edouard:
All good, builds fine, merging soon.

PR #253, opened 2016-03-21
Removal of pci-dm-helper

eric-ch:
This helper is no longer useful. It used to handle PCI IO to config.
space and resources from stub-domain'ed QEMU to manage PCI pass-through
in that context.
Passed-through PCI devices are now passed to the stubdom as well using
pci-back/front PV drivers, with some constraint on PCI bus addressing,
that allow the stubdomain those access to the device. Permissions are
handled in the PV drivers and the toolstack should ensure that sensitive
devices are not passed-through.


rossphilipson:
Big +1

eric-ch:
@rossphilipson linked to OXT-490

jean-edouard:
I thought pc-dm-helper was there mostly for accessing to the host bridge (00:00.0), since some drivers need that. How do we handle that kind of access without pci-dm-helper?

eric-ch:
That was igfx-PVM related iirc, so defunct.

jean-edouard:
Yes, Intel GPU drivers are an example it, but I don't think they're the only ones. I think some NVIDIA drivers also do that.
From a security point of view, it doesn't sound like a good idea to allow host bridge access though, so maybe it makes sense to just drop support for these devices/drivers.
It would be nice to get a message in the logs when that happens though.

jean-edouard:

jean-edouard:
@eric-ch, could you document, maybe on a wiki page, what you think would be needed to enable Intel GPU passthrough in OpenXT? I think that information could be verry useful some day!

rossphilipson:
+1 on documenting how it worked and how we would fix it if we needed to do IGD pt.

rossphilipson:
will merge after stuffs is documented.

eric-ch:
Regarding the pci-dm-helper, it was hooked with ioemu (https://github.com/OpenXT-Extras/ioemu-pq/blob/master/master/stubdoms#L1876). This patch is not in qemu-dm because these accesses are now possible from the stub-domain, and I can think of nothing else talking with it for now. This daemon is running, but nothing talks with it.
Along with that it was giving specific IO permissions to the PCI device, those are now attributed by the toolstack when passing the device to the stub-domain.

I can document what I think would be required for IGD pass-through, but that will take some digging, and it is not directly related to this PR. Upstream has seen quite some work to support IGD pass-through with Xen/QEMU, possibly deprecating some component used to achieve it in XenClient.

rossphilipson:
Wiki put up:
https://openxt.atlassian.net/wiki/display/DC/PCI-DM-HELPER+and+IGD+Passthrough

rossphilipson:
Well there have not been any comments for a few days. I think this can go in. Will merge soon...

PR #254, opened 2016-03-21
Remove monit

eric-ch:
While monit could be useful in production for error recovery and resource starvation management, currently it is only restarting daemons in dom0 on failure which can hide (and
already has hidden) issues.

I understand this is debatable, as testing should be done on the daemons initialy and monit, only being a failsafe for now, could provide more help on resource based events and the likes. But, for now, I believe this is not used as such and can get in the way of development.


rossphilipson:
+1 on this one too. Please create tickets for these.

eric-ch:
@rossphilipson linked to OXT-514

rossphilipson:
Some concerns were raised about removing this. Let's hold off until they are resolved.

jean-edouard:

rossphilipson:
Will give it some testing on Monday.

rossphilipson:
Testing a build with monit gone. Seems to be fine.

rossphilipson:
Got clearance from concerned parties. Merging soon...

PR #256, opened 2016-03-22
Make the ndvm default to SELinux enforcing.

stephensmalley:
This change restores the ndvm to SELinux enforcing mode by default,
as it used to be prior to the jethro merge.   Depends on the ndvm
policy fixes.


jean-edouard:
Merging soon, see: https://github.com/OpenXT/xenclient-oe/pull/251#issuecomment-200003754

PR #260, opened 2016-03-28
[xen-pq] Remove patches added in recipes including xen.inc

rossphilipson:
Doing this sort of stuff is really messy. There is no reason to split the
patch queue up like that. Also nuked a few dead patches.

OXT-515


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/27
LGTM, merging soon.

eric-ch:
+1 as well.

PR #261, opened 2016-03-29
linux: uivm: disable xen-fbfront

jean-edouard:
There's a race at UIVM boot time between xen-fbfront and xenfb2.
For some reason, on most OpenXT version and test machine combos, xenfb2 wins and everything works fine.
But on my laptop (HP EliteBook 850 G2), xen-fbfront wins 4 times out of 5, and the UI never comes up.

This PR removes xen-fbfront from UIVM, fixing the problem for good.

OXT-504


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/28

rossphilipson:
Good find, good fix :)

I will test that build.

rossphilipson:
Well I have installed it and rebooted numerous times and no problems. I have not had a lot of luck reproing the original problem but I guess since it is a race I could spend all day trying to do that. It clearly fixed things on Jed's system where it always happened so I am inclined to merge soon.

PR #262, opened 2016-03-29
Change configure_measured_launch() to use luksOpen --test-passphrase.

stephensmalley:
configure_measured_launch() was invoking cryptsetup with a custom-defined
luksCheckKey action to test the key it just added.  The patch implementing
luksCheckKey was not forward-ported during the jethro merge, so this
was always failing.  As cryptsetup has changed its internal API and
data structures, the patch no longer applies, even manually, and would
require a rewrite.  However, research discovered that luksOpen now
supports a --test-passphrase option that seems to provide equivalent
functionality.  So simply switch from luksCheckKey to luksOpen
--test-passphrase.

With this change, I can successfully run tpm-setup after install
and transition to a measured launch.  This does not fix being
able to initially install with measured launch enabled; some other
issue is still breaking that functionality.

### DIFF
--- a/recipes-openxt/xenclient/xenclient-tpm-scripts/tpm-functions
+++ b/recipes-openxt/xenclient/xenclient-tpm-scripts/tpm-functions
@@ -245,7 +245,7 @@ configure_measured_launch()
         rm -f ${key_file}
         return 1
     }
-    cat ${key_file} | cryptsetup -q -d - -S 1 luksCheckKey \
+    cat ${key_file} | cryptsetup -q -d - -S 1 luksOpen --test-passphrase \
         ${config_dev} >/dev/null 2>&1 || {
         echo "failed to verify new crypto key in config LUKS" >&2
         rm -f ${key_file}


jean-edouard:
According to cryptsetup --help, `luksOpen` seems to be a legacy alias for `open`.
Could it be safer to use `open` here, in case the alias gets removed from a future version of cryptsetup?

jean-edouard:
Feel free to ignore the previous comment, we use luksOpen all over OpenXT anyway...
I'll test this today.

jean-edouard:
These scripts also use `luksCheckKey`:
https://github.com/OpenXT/xenclient-oe/tree/master/recipes-openxt/xenclient/xenclient-sec-scripts

I'm not sure what they're for, but we should probably either fix them or remove them...

rossphilipson:
A grep through the world seems to indicate that xenclient-sec-scripts is not used anywhere.

rossphilipson:
I would like to leave this open for a day or two in case anybody else wants to comment on it.

jean-edouard:
This PR fixed tpm-setup on my HP EliteBook 850 G1, nice to see a post-Jethro-merge measured boot! :)

+1

jean-edouard:
Here's a ticket for this: https://openxt.atlassian.net/browse/OXT-518

jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/30

jean-edouard:
Works great, merging soon.

PR #263, opened 2016-03-29
Add RDEPENDS for libgcc to trousers.

stephensmalley:
tcsd depends on libgcc_s.so.1 for pthread_cancel.
The absence of libgcc_s.so.1 from the installer rootfs was
causing tcsd to abort during installation, and thereby
preventing installing with measured launch enabled.
libgcc_s.so.1 was already being included in the dom0 rootfs,
which is why tpm-setup could work.

With this fix and the luksCheckKey fix applied, one can
once again install with measured launch enabled.

OXT-473
OXT-518

### DIFF
--- a/recipes-openxt/trousers/trousers_0.3.2-1.bb
+++ b/recipes-openxt/trousers/trousers_0.3.2-1.bb
@@ -49,6 +49,7 @@ do_install_append() {
 }
 
 RPROVIDES_${PN} =+ "${PN}-data"
+RDEPENDS_${PN} = "libgcc"
 PACKAGES =+ "${PN}-data"
 FILES_${PN}-data = "${datadir}/trousers/system.data.auth \
 	${datadir}/trousers/system.data.noauth \


jean-edouard:
- running ldd on tcsd and libpthread doesn't say libgcc. Did I miss something?
- if pthread functions need libgcc, shouldn't libc6 RDEPEND on libgcc?

stephensmalley:
I debugged it by trying to install with measured launch enabled.  When it failed, I switched to a virtual console, and tried running tpm_* commands from tpm-functions by hand.  Discovered that tcsd was dying on first command.  Ran tcsd in the foreground with -f and got this message before it aborted:
libgcc_s.so.1 must be installed for pthread_cancel to work
Looked for that message and found the following:
http://comments.gmane.org/gmane.linux.embedded.yocto.general/28671
https://lists.yoctoproject.org/pipermail/meta-virtualization/2014-January/000592.html
http://patches.openembedded.org/patch/49633/
among others. Also seems to have been submitted to trousers in meta-measured, although not clear it was merged:
https://lists.yoctoproject.org/pipermail/yocto/2013-November/017112.html
Speaking of which, we should look at migrating to meta-measured for trousers and tpm-tools at some point, or if not, at least updating to a newer version.

jean-edouard:
Thanks for the links!
Looks like libc doesn't RDEPEND on libgcc because it causes circular dependencies.
I guess it's fine to do it at the application level then.
+1

jean-edouard:
rsyslog also uses pthread_cancel, should we add that RDEPEND to it too?

jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/30

jean-edouard:
Just installed that build with successful install-time sealing! :)
Merging soon.

PR #264, opened 2016-03-29
[blktap2] Add split our recipe for blktap2 to avoid circular deps in OE.

rossphilipson:
OXT-519


jean-edouard:
LGTM, not activated yet, so merging soon.

PR #265, opened 2016-03-30
Remove libpciemu defunct recipe.

eric-ch:
This recipe is no longer used. The repository should belong to OpenXT-Extra.

OXT-522


rossphilipson:
See https://github.com/OpenXT/libpciemu/pull/1

These two go in together.

rossphilipson:
Will merge both soon....

PR #268, opened 2016-03-31
Manage policy/modules.conf as files rather than patches.

stephensmalley:
policy/modules.conf controls which policy modules are built and
therefore included into the final policy.  Upstream refpolicy releases
ship with a default modules.conf generated via make conf that enables
all modules.  OpenXT was then disabling specific modules via one patch
and then adding its own new modules via a second patch.  However,
this means that as OpenXT updates to new refpolicy releases, it
may unintentionally include additional policy modules enabled upstream
that are not needed or desired.  It already appears to enable many
modules that appear to be unnecessary for dom0 or ndvm operation, making
the policy larger than needed and potentially allowing undesired
functionality.

Change this approach to manage modules.conf as files rather than
patches.  Provide one file that enables modules from upstream refpolicy
and another that enables OpenXT policy modules, and concatenate them
to produce the final modules.conf.  A similar approach is used in
the Fedora selinux-policy source RPM to produce a modules.conf.

This change enables the same set of modules as before, although
modules.conf order is not preserved (or important).  Disabling
modules is left to future changes.


stephensmalley:
Note that we could further prune modules-upstream.conf to omit all of the modules that are being disabled (= off), because any module not listed in modules.conf is set to off by default.  Not clear as to which is better style; keeping them in modules-upstream.conf shows that there was an explicit decision to disable them versus just not realizing they existed.

stephensmalley:
Looking at what additional policy modules could be disabled, I noticed that the su policy module is enabled.  And su is present in dom0 and the ndvm.  However, running su, even after nr, immediately fails due to a SELinux denial.  Which I could fix.  But I was wondering what use case there is for su in dom0 or the ndvm.  If none, then might as well just disable it and possibly remove the program too.

jean-edouard:
I verified that the module list is the same before and after this PR.
The separation makes sense to me, merging soon.

About the off modules, I vote for keeping the `<module> = off` lines.

About `su`, it is provided by the shadow package, which doesn't seem to have an option for not building `su`. I think the fact that SELinux blocks it is a good thing, so I would be in favor of leaving things as they are.

PR #272, opened 2016-04-01
Disable unnecessary policy modules.

stephensmalley:
Disabling unused policy modules is helpful in reducing disk
and kernel memory footprint and ensuring that we are not allowing
unexpected behavior/functionality.  The policy modules disabled
in this change do not appear to correspond to any packages installed
in dom0 or the ndvm.

In order to disable the hal policy module, we also need to wrap
any references to it in other modules with optional_policy blocks
in order to allow the policy to build without it.  This required
updating the policy patch for the init module and a couple of the
OpenXT-added policy modules.  Using optional_policy blocks is
preferred regardless for any module that is not a hard dependency.


jean-edouard:
Looks great! Testing today.

jean-edouard:
No AVC, merging soon!
Note: if/when we enable SELinux in UIVM, some of these modules will have to be re-enabled.

PR #273, opened 2016-04-02
OXT-524 : fix the software licence of linux-firmware package

dozylynx:
Changes:
* Change ipkg metadata to indicate Intel and Broadcom, rather than
  Proprietary. The two licenses are Open Source compatible.
  Note that the build does indicate warnings about no generic
  licence file existing for those licences, which seems fair.

* Ensure that the Broadcom licence ends up in the package, as
  required by the licence. Previously it was missing.

* Since the upstream licence for the Broadcom firmware is unclear,
  include both possible licences.

Signed-off-by: christopher.clark6@baesystems.com
On behalf of BAE Systems.

### DIFF
--- a/recipes-kernel/linux-firmware/LICENCE.broadcom_bnx2
+++ b/recipes-kernel/linux-firmware/LICENCE.broadcom_bnx2
@@ -1,0 +1,8 @@
+ This file contains firmware data derived from proprietary unpublished
+ source code, Copyright (c) 2004 - 2010 Broadcom Corporation.
+
+ Permission is hereby granted for the distribution of this firmware data
+ in hexadecimal or equivalent format, provided this copyright notice is
+ accompanying it.
+
+Found in hex form in kernel source.
--- a/recipes-kernel/linux-firmware/linux-firmware_git.bb
+++ b/recipes-kernel/linux-firmware/linux-firmware_git.bb
@@ -4,9 +4,11 @@ SECTION = "kernel"
 # Notes:
 # Based on the OE one but with the new git repository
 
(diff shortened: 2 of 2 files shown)


eric-ch:
Shouldn't these licenses be added to the [additional license files](https://github.com/OpenXT/xenclient-oe/tree/master/files/additional-licenses) to avoid the warning (as explained [here](https://www.yoctoproject.org/blogs/khem/2014/suppliment-common-licenses-yocto-project)) and have the license handled by bitbake?

cjp256:
It seems to me that the ideal solution is to simply drop this recipe in favour of the one already in oe-core.

eric-ch:
@cjp256 Indeed.

dozylynx:
I'm happy to do either of the above.

The current OpenXT linux-firmware package only inserts two firmware blobs into the build product and so only introduces two software licenses (ignoring the lack of clarity around which Broadcom license applies).

Would using the upstream recipe allow us to filter which firmware blobs get installed or would we have to carry all the firmware blobs on the filesystem and comply with all of the software licenses?

richpersaud:
Did the jethro merge result in the use of any upstream OE packages and the retirement of OpenXT-unique packages?  If so, is a list of those packages available?

We have often discussed the many benefits of using upstream OE packages.  If linux-firmware can be migrated to an upstream OE package, are there other packages which can be transitioned to upstream OE?

dozylynx:
Retiring this pull request, replacing it with a newer version:
https://github.com/OpenXT/xenclient-oe/pull/278

PR #275, opened 2016-04-05
Fix the SRCREV to point to the current tip rev.

rossphilipson:
OXT-517


jean-edouard:
Yup, merging very soon to fix the build.

PR #277, opened 2016-04-07
Move the do_compile_prepend to their own tasks

rneilturner:
With the current do_compile_prepend there is no order precedence for derived projects to add their own modules or make changes before compile to existing upstream policies. By adding specific tasks for these steps, derived projects can better extend the module and policy changes.


jean-edouard:
Builds fine, merging soon.

PR #279, opened 2016-04-08
OXT-524: linux-firmware : append to upstream oe, fix license packaging and install all bnx2 firmware

dozylynx:
The previous xenclient-oe version of linux-firmware indicated
"Proprietary" as license type, rather than Intel and Broadcom
licenses that are Open Source compatible.
The Broadcom license was also missing from the filesystem.

Retire the recipe in favour of the upstream linux-firmware pkg.

Signed-off-by: christopher.clark6@baesystems.com

fixes: OXT-524


cjp256:
@dozylynx This is a good step 1.  But I believe you will find that dom0, installer, and ndvm are probably now using a much larger amount of disk space.  This would be because the "linux-firmware" package is pulling in all of the firmware packages shipped in linux-firmware.  You want to update the install lists to instead specify the ones that OXT cares about (apparently iwlwifi* and bnx2).   On a related note, dom0 should probably pull in the skylake graphics firmware (and drop the networking related firmware).

To see what firmware is shipped in what package, you can look under the packages-split in OE's linux-firmware work directory.

dozylynx:
@cjp256 : Two more commits pushed to make the changes you recommended. It looks like linux-firmware isn't currently present in dom0 image definition, unless I've missed it, so I've just modified the installer and NDVM images to specify the iwlwifi* and bnx2 firmware packages.

Upstream doesn't individually package the bnx2 firmware yet and it's less clean than some of the other firmware packages since the software license is contained in the common WHENCE file. It may still be worth an upstream submission though.

rossphilipson:
@dozylynx / @cjp256 is this one ready to be tested? Any guidance on testing - e.g. what firmware modules we should expect to see there?

dozylynx:
@rossphilipson : the code in this PR is ready to test.

These files should be present in the installer and NDVM filesystems:
/lib/firmware/bnx2/bnx2-mips-09-6.2.1b.fw
/lib/firmware/WHENCE

and also the iwlwifi firmware.

The bnx2 work means that we do end up with some changes in OpenXT versus the upstream OE linux-firmware recipe and that work could be sent upstream, but that can be a separate change.

rossphilipson:
Ack. We will build this and try it out.

rossphilipson:
Actually it is in build 44. I will smoke it tomorrow.

recipes-kernel/linux-firmware/linux-firmware_git.bbappend
```diff
@@ -0,0 +1,25 @@
+PR .= ".1"
```
cjp256: This is fine, but shouldn't be required any more because of the jethro uprev (with the prserver).

cjp256:
LGTM, @dozylynx did you submit the bnx patch upstream to oe-core?  If not, that would be desirable so we can drop it in the next uprev :)

dozylynx:
@cjp256 :Yep, I subscribed to the openembedded-core mailing list today (after your nudging here!) and sent a patch via email. Have yet to see it surface in their patch tracker but will keep an eye on it.

cjp256:
@dozylynx Awesome, thanks! :)

dozylynx:
http://patches.openembedded.org/patch/120321/

rossphilipson:
OK couple of things:
1. Everything you list seems to be in the NDVM
2. The WHENCE file is missing in the installer rootfs but the other files are there.
3. There is a bunch of network firmware in dom0 too including more bnx files than you noted.

dozylynx:
OK, will take a look.

dozylynx:
aha, looks like there's a reference to linux-firmware that I missed in xenclient-oe/recipes-core/packagegroups/packagegroup-xenclient-dom0.bb

rossphilipson:
Issue 2 above about the WHENCE file missing is not an issue, it was just due to some muppetry on my part. It looks like 3 is fixed so we will do another build with this.

rossphilipson:
I tried to do a build of this but it failed....
http://158.69.127.133:8020/builders/openxt-custom/builds/52

It looks like you need to rebase this on some recent changes for surfman, sorry 'bout that.

dozylynx:
rebased, cleaned up a little and rolled in the change to include all bnx2 firmware files.

dozylynx:
Upstream patch submitted to break out all bnx2 firmwares and add a bnx2 package that installs them all.
http://patchwork.openembedded.org/patch/121833/

jean-edouard:
LGTM, will merge soon.

PR #280, opened 2016-04-11
Enable the fdisk advanced disk option for expert mode

rneilturner:
In our default configuration of busybox, we don't enable the CONFIG_FEATURE_FDISK_ADVANCED, which enables the expert mode needed for realigning the partition during install. During initial install the realignment step fails and the first sector is not moved, thereby causing issues for dmsetup. 


cjp256:
LGTM

richpersaud:
Given the complexity of triage and implications for upgrade, it would be helpful to have a JIRA ticket where relevant information can be compiled and linked to additional work that may be needed.

Does this affect only master, e.g. do we know if/why previous releases did not exhibit this failure mode?  

Does this affect only the Dell 9010, only specific disks/controllers on the 9010, or all devices/disks?

jean-edouard:
+1
See https://openxt.atlassian.net/wiki/display/~jed/Busybox+overview

jean-edouard:
Will merge soon.

richpersaud:
Can we please have further documentation/discussion of the upgrade implications before merging, either on the mailing list or in a JIRA ticket?

jean-edouard:
Sure, but the installer is basically broken without this fix, so please don't take too long.

rneilturner:
See OXT-534.

@richpersaud see your answers below

Does this affect only master, e.g. do we know if/why previous releases did not exhibit this failure mode?

RT - This option was enabled in previous versions of OXT (pre-Jethro). During the up rev to Jethro this option was disabled causing the installer issue.

Does this affect only the Dell 9010, only specific disks/controllers on the 9010, or all devices/disks?

RT - This problem is not platform specific. It really came down to how large your core boot loader ended up being. If it was greater than 24k it would break your install and not work. In our case the boot loader image was 32k. Also, this change does not break the upgrade case.

jean-edouard:
(moved "wiki" page to https://openxt.atlassian.net/wiki/display/DC/Busybox+overview)

richpersaud:
Thanks for responses, ok to merge.

jean-edouard:
Thanks, merging soon!

PR #281, opened 2016-04-11
linux: uivm: disable xen-fbfront in Linux 4.4

jean-edouard:
OXT-504


rossphilipson:
It is already in 3.18 so let's get it into 4.4, merging soon...

PR #282, opened 2016-04-12
busybox: add missing options and remove potentially harmful components

jean-edouard:
OXT-534


jean-edouard:
See https://openxt.atlassian.net/wiki/display/DC/Busybox+overview

jean-edouard:
I just amended the commit, after finding an issue with it.
The first version enabled `ip`, which somehow "won" over the binary from `iproute2`.
The busybox `ip` applet clearly misses a feature NDVM uses, since using it breaks the network in OpenXT.

I also noticed that `su`, `passwd` and `login` were getting automatically re-enabled by OpenEmbedded.
We don't want that, so I added an empty file to remove login-related auto-inclusions.

eric-ch:
+1, smoke-tested, it looks fine. Will merge soon.

PR #289, opened 2016-04-18
udev: fix for Linux 4.4

jean-edouard:
Fix udev by adding 2 patches found there:
http://lists.openembedded.org/pipermail/openembedded-core/2016-January/115866.html


eric-ch:
Was fixed for the openxt-linux-4.4 distro only: https://github.com/OpenXT/xenclient-oe/blob/master/recipes-core/udev/udev_182.bbappend#L39
Otherwise the 3.18 would not build.

jean-edouard:
@eric-ch thanks for commenting, I didn't see your fix, which also explains why Linux 4.4 builds used to work.
The version of the patch I found should be more flexible, so I suggest we keep it and remove your "HACK", if that works for you. (see second commit)

eric-ch:
@jean-edouard Yes, better use the upstream patch. Thanks for removing the hack :).

PR #291, opened 2016-04-19
refpolicy: Update SELinux definitions for Linux 4.4

stephensmalley:
Linux 4.4 includes additional SELinux security classes that
are not defined by the refpolicy release used in OpenXT.

Consequently, on boot, one gets the following warnings in dmesg:

SELinux:  Class netlink_iscsi_socket not defined in policy.
SELinux:  Class netlink_fib_lookup_socket not defined in policy.
SELinux:  Class netlink_connector_socket not defined in policy.
SELinux:  Class netlink_netfilter_socket not defined in policy.
SELinux:  Class netlink_generic_socket not defined in policy.
SELinux:  Class netlink_scsitransport_socket not defined in policy.
SELinux:  Class netlink_rdma_socket not defined in policy.
SELinux:  Class netlink_crypto_socket not defined in policy.
SELinux:  Class binder not defined in policy.
SELinux: the above unknown classes and permissions will be allowed

In Linux 4.2, additional, finer-grained netlink socket security classes
were defined for various netlink protocols to permit specific control
over them rather than falling back to the general netlink_socket security
class.  The binder security class was also finally upstreamed from Android
for control over the binder driver.

Add two patches from upstream refpolicy to define these security
classes, and one patch from upstream refpolicy to allow NETLINK_GENERIC
usage by networkmanager.  These patches silence the above warnings,
ensure proper SELinux enforcement over these netlink protocols (and
over the binder driver if one were to ever enable it), and allow operation
of NetworkManager.

Adjust the existing patch for the apmd domain used by acpid to also
allow it to use NETLINK_GENERIC sockets to address denials triggered
once the new classes are defined such as:

avc:  denied  { create } for  pid=860 comm="acpid" scontext=system_u:system_r:apmd_t:s0 tcontext=system_u:system_r:apmd_t:s0 tclass=netlink_generic_socket permissive=0

There may be other domains previously allowed access to netlink_socket
that will now need access to one of the more specific classes, but
I do not see any other denials at present.

These changes have no impact on older kernels (e.g. 3.18).  The existing
rules on netlink_socket are left in place for compatibility with
such kernels.

### DIFF
--- a/recipes-security/refpolicy/refpolicy-mcs-2.20141203/patches/policy.modules.contrib.apm.diff
+++ b/recipes-security/refpolicy/refpolicy-mcs-2.20141203/patches/policy.modules.contrib.apm.diff
@@ -2,7 +2,15 @@ Index: refpolicy/policy/modules/contrib/apm.te
 ===================================================================
 --- refpolicy.orig/policy/modules/contrib/apm.te
 +++ refpolicy/policy/modules/contrib/apm.te
-@@ -100,6 +100,8 @@ dev_rw_apm_bios(apmd_t)
+@@ -64,6 +64,7 @@ dontaudit apmd_t self:capability { setui
+ allow apmd_t self:process { signal_perms getsession };
+ allow apmd_t self:fifo_file rw_fifo_file_perms;
+ allow apmd_t self:netlink_socket create_socket_perms;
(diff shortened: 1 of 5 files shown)


jean-edouard:
LGTM, does what it says, will merge soon.

PR #293, opened 2016-04-26
Preserve label of /var/lock/xen-atapi-pt-lock* on restorecon.

stephensmalley:
atapi_pt_helper creates `/var/lock/xen-atapi-pt-lock-<a>_<b>_<c>_<d>`,
and policy defines a type transition to assign it atapi_helper_lock_t,
but there is no file_contexts (.fc) entry for it.  Thus, a restorecon
will incorrectly reset it to var_lock_t.  Add an entry for it.
After this change, restorecon -Rv /var/lock reports no changes.

OXT-306


jean-edouard:
LGTM, works fine, merging soon.

PR #294, opened 2016-04-27
refpolicy: rename ctxusb to vusb

jean-edouard:
OXT-357


stephensmalley:
Any need to preserve file label compatibility for upgrades?
If so, then you could define typealias statements to alias the old names to the new names for any types used for files on disk so that the existing files won't have invalid labels on upgrades.
Not sure what OpenXT's upgrade compatibility requirements are, if any.

jean-edouard:
Good question, I'm not sure either... I'll figure it out, thanks for pointing this out!

jean-edouard:
@stephensmalley, I looked at the upgrade code, and I'm not sure I see how that could be a problem, regardless of whether we support upgrade from 5.0.x or not.

I think the upgrade will create a new dom0 root logical volume, using the new labels and the new policy, then reboot to that. In that case, I don't see how an new policy would have to deal with old contexts, nor the the other way around.

Am I missing something here?

stephensmalley:
I don't really know how OpenXT upgrades work, so I was just asking whether it would be a problem.  With Fedora/RHEL, they always have to define typealiases to preserve compatibility on updates/upgrades.  Android replaces the system partition wholesale but does a recursive restorecon of the userdata partition on upgrades to reflect any policy changes.  Does OpenXT upgrade run with SELinux disabled in the kernel so that it can set the labels without having them defined in the currently loaded policy, or with the old policy loaded?  If the latter, then the upgrade process needs CAP_MAC_ADMIN / mac_admin and permissions to unlabeled_t to set unknown labels on disk.

jean-edouard:
In our case, I think we just dd the new rootfs, with xattrs set at build time, so should be good to go.

stephensmalley:
Ok, SGTM.

rossphilipson:
Built and installed. No problems. The AVCs when I ran usb-state with SELinux permissive said vusbd_t. Looks good, will merge soon.

PR #295, opened 2016-04-27
[dm-agent] Remove daemon from dom0

eric-ch:
This is no longer required.

OXT-544


eric-ch:
Break support for guests without stub-domain.
See https://openxt.atlassian.net/browse/OXT-544.

PR #298, opened 2016-05-04
Have one dm-agent for both stubdom and dom0

eric-ch:
Without dmbus there is no need to split dm-agent in two flavor.

OXT-544

### DIFF
--- a/recipes-core/images/xenclient-stubdomain-initramfs-image.bb
+++ b/recipes-core/images/xenclient-stubdomain-initramfs-image.bb
@@ -10,7 +10,7 @@ DEPENDS += "dialog"
  
 IMAGE_FSTYPES = "cpio.gz"
 IMAGE_INSTALL = "busybox bridge-utils initramfs-xenclient"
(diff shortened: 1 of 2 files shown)


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/58

rossphilipson:
Tested. Everything seems to be happy. I know there are already errors in the log but this particular one seems like it could be related to these changes:

`May  5 14:01:13.288217 surfman: Error:/home/build/160504-2/openxt/build/tmp-glibc/work/core2-32-oe-linux/surfman/0+gitAUTOINC+58fc0efc42-r0/git/surfman/src/domain.c:domain_set_visible:212: failed`

It seemed to happy just once early during startup of the platform.

rossphilipson:
Ok so I see that error in builds without these changes so it is unrelated. In fact I see numerous errors in the log so I am going to open a ticket to try to track these down and see if they are serious or not. Anyway, this PR seems to be in good shape, merging soon.

PR #300, opened 2016-05-06
fstab: make /var/cache big enough for opkg downloads

jean-edouard:
OXT-553


rossphilipson:
+1 the 2M tmpfs was unusable for opkg installs

PR #305, opened 2016-05-17
[xen] Pull in upstream patch to support xsave on Skylake.

jctemkin:
This patch was originally posted on the Qubes mailing list for Xen 4.4, and has been backported to Xen 4.3.4.

### DIFF
--- a/recipes-extended/xen/files/xen-fix-xsave.patch
+++ b/recipes-extended/xen/files/xen-fix-xsave.patch
@@ -1,0 +1,146 @@
+From e8121c54ec8cc6a92ae2d7219e7a7fede6dfd137 Mon Sep 17 00:00:00 2001
+From: Jan Beulich <jbeulich@novell.com>
+Date: Tue, 25 Feb 2014 09:34:04 +0100
+Subject: [PATCH] x86/xsave: enable support for new ISA extensions
+
+Intel has released a new version of Intel Architecture Instruction Set
+Extensions Programming Reference, adding new features like AVX-512,
+MPX, etc. Refer to
+http://download-software.intel.com/sites/default/files/319433-015.pdf
+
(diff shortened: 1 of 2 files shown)


jean-edouard:
https://openxt.atlassian.net/browse/OXT-568

jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/72
Installed successfully on Skylake! :)
Now I get this error that I also got with a Xen-ified Debian:

    [    0.000000] IOAPIC[0]: Unable to change apic_id!

Do you know about that one?

Either way, I think this PR is good to go, will merge tomorrow.

jctemkin:
Excellent, glad hear the install went well :) However, that apic message is totally new to me, sorry! Hopefully the fix isn't too difficult to find.

rianquinn:
You might want to check to see if your hardware supports changing the APIC ID, as that is not supported on all hardware

jean-edouard:
@rianquinn I'm not sure how to check that, I'm using a Dell E7470.
However I looked at the code and the error should not be fatal... I'll keep looking.

jean-edouard:
Nevermind, the machine was still running Debian when I saw the error... Sorry.

rianquinn:
My guess is your fine. The only way I know to check would be to write the code to try and change it. But this is from the Intel Manual (section 10.4.6): 

> In MP systems, the local APIC ID is also used as a processor ID by the BIOS and the operating system. Some processors permit software to modify the APIC ID. However, the ability of software to modify the APIC ID is processor model specific. Because of this, operating system software should avoid writing to the local APIC ID register.

jean-edouard:
Thank you @rianquinn.
I got OpenXT to boot to the UI after installing correctly :)

jean-edouard:
Merging soon.

PR #306, opened 2016-05-17
[OXT-417] Disable Avocent KVM autosleep.

jctemkin:


jean-edouard:
I can't really test this, but since the code runs only for once specific device it's probably fine...
2 small issues though:
- The indentation of SRC_URI is broken
- The script uses `-eq` on strings, it should probably use `=` instead.

jean-edouard:
`[[ "on" -eq "off" ]] && echo yes` :)

jean-edouard:
LGTM, can't test but will merge anyway, since I don't see how it could make anything worse.

PR #311, opened 2016-05-25
viptables: xen: fix list logic and add missing static

jean-edouard:


eric-ch:
LGTM, will merge soon.

PR #312, opened 2016-05-25
sshv4v and scpv4v: bind client port

jean-edouard:
OXT-558


jean-edouard:
Depends on https://github.com/OpenXT/v4v/pull/9

PR #314, opened 2016-05-27
OXT-575: package and install two iwlwifi firmwares in NDVM and installer filesystems

dozylynx:
This PR is deliberately omitting installation of these two firmwares into the dom0 filesystem due to OXT-576. This should enable validation of correct functionality of the wifi device without the firmware present in dom0. If this succeeds, all wifi firmware can then be removed from the dom0 filesystem.


jean-edouard:
LGTM, will merge tonight/Monday

PR #319, opened 2016-06-02
Qemu pq rearrange

eric-ch:
Apply backported patches first, then feature patches.
Fix a possible regression introduced by db7e71c.
Small refactoring in vbe-xt-extensions as it seemed a bit confusing (introduce 3 helpers).


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/86
LGTM, merging soon.

PR #323, opened 2016-06-09
Upgrade linux kernel to 4.4.13

eric-ch:
1588a8b upstreamed, removing from the patch-queue.

- Xen related
aa1cc4d xen: use same main loop for counting and remapping pages
6232876e xen/events: Don't move disabled irqs
18875bf mm: use phys_addr_t for reserve_bootmem_region() arguments
3bf7d03 xen/x86: actually allocate legacy interrupts on PV guests


jean-edouard:
LGTM, built and tested, merging soon.

PR #324, opened 2016-06-13
answer files: update certificates for unattended installs

jean-edouard:
passphrase is "openxt" without the quotes

OXT-605 (see first comment)


jean-edouard:
Just added a commit to remove the hardcoded drive name.
This PR should compensate for it: https://github.com/OpenXT/installer/pull/17

rossphilipson:
Built here:
http://158.69.127.133:8020/builders/openxt-custom/builds/99
Testing now...

rossphilipson:
@jean-edouard: This may not be working or something else may be broken. I had some trouble getting the unattended install to configure tboot during install time. I ended up having to use tpm-setup to make it work. I guess it is possible the TPM was in a bad state but I thought I had everything setup. Once I got measured launch working it kept making me reseal on every boot.

rossphilipson:
So it turns out that the unattended install does not setup measured launch so that is not a new issue. I tried a master build, unattended install. I was able to use tpm-setup and then reseal and ML worked fine. I think there is a problem with this PR.

rossphilipson:
Retested this again. It seems to be working. I am not sure what the problem was earlier - could be HW related on the system I tested. I also validated that the new keys are installed. Merging soon.

PR #327, opened 2016-06-16
Qemu hardening

eric-ch:
http://docs.openstack.org/security-guide/compute/hardening-the-virtualization-layers.html

Also use pixman from the QEMU tree since it is built.

OXT-604


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/92

jean-edouard:
LGTM, +1.

jean-edouard:
This is going into 6.0, and was smoke tested last week.
Merging soon.

PR #330, opened 2016-06-17
OXT-496: remove ca-certificates package from dom0

dozylynx:
using the BAD_RECOMMENDATIONS and PACKAGE_REMOVE macros when
building the dom0 rootfs.

libcurl and wget have RRECOMMENDS which were likely pulling it in.

refs: OXT-496

### DIFF
--- a/recipes-core/images/xenclient-dom0-image.bb
+++ b/recipes-core/images/xenclient-dom0-image.bb
@@ -9,7 +9,8 @@ IMAGE_FSTYPES = "xc.ext3.gz"
 
 # No thanks, we provide our own xorg.conf with the hacked Intel driver
 # And we don't need Avahi
-BAD_RECOMMENDATIONS += "xserver-xorg avahi-daemon avahi-autoipd"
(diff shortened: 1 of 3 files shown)


jean-edouard:
Adding ca-certs to `BAD_RECOMMENDATIONS` makes sense to me.
However, the `PACKAGE_REMOVE` thing is an ugly hack that we added after not managing to remove all the dependencies on avahi.
Does ca-certs get installed even when it's just in `BAD_RECOMMENDATIONS`?

dozylynx:
OK, I checked dropping the PACKAGE_REMOVE from both dom0 and installer from the cset and just using BAD_RECOMMENDATIONS does the right thing: the CA certs are not present, so I've pushed an revised single patch.

jean-edouard:
LGTM now, thank you :)

jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/92

jean-edouard:
LGTM, merging soon.

PR #331, opened 2016-06-20
OXT-575 repackage iwlwifi firmware

dozylynx:
Create a virtual package: linux-firmware-iwlwifi
that depends upon all of the individually packaged iwlwifi firmware packages.

Change the image package lists to use the new virtual package.

Add packaging for the 8000c series device firmwares.

### DIFF
--- a/recipes-core/images/xenclient-installer-image.bb
+++ b/recipes-core/images/xenclient-installer-image.bb
@@ -37,24 +37,7 @@ IMAGE_INSTALL = "\
     packagegroup-xenclient-common \
     packagegroup-xenclient-installer \
     kernel-module-e1000e \
-    linux-firmware-iwlwifi-135-6 \
-    linux-firmware-iwlwifi-3160-7 \
-    linux-firmware-iwlwifi-3160-8 \
-    linux-firmware-iwlwifi-3160-9 \
-    linux-firmware-iwlwifi-6000-4 \
-    linux-firmware-iwlwifi-6000g2a-5 \
-    linux-firmware-iwlwifi-6000g2a-6 \
-    linux-firmware-iwlwifi-6000g2b-5 \
-    linux-firmware-iwlwifi-6000g2b-6 \
-    linux-firmware-iwlwifi-6050-4 \
-    linux-firmware-iwlwifi-6050-5 \
-    linux-firmware-iwlwifi-7260-7 \
-    linux-firmware-iwlwifi-7260-8 \
-    linux-firmware-iwlwifi-7260-9 \
-    linux-firmware-iwlwifi-7260-12 \
-    linux-firmware-iwlwifi-7260-13 \
-    linux-firmware-iwlwifi-7265-8 \
-    linux-firmware-iwlwifi-7265-9 \
+    linux-firmware-iwlwifi \
     linux-firmware-bnx2 \
     ${ANGSTROM_EXTRA_INSTALL}"
 
--- a/recipes-core/images/xenclient-ndvm-image.bb
+++ b/recipes-core/images/xenclient-ndvm-image.bb
@@ -33,24 +33,7 @@ IMAGE_INSTALL = "\
     xenclient-dbusbouncer \
     networkmanager \
     xenclient-toolstack \
-    linux-firmware-iwlwifi-135-6 \
-    linux-firmware-iwlwifi-3160-7 \
-    linux-firmware-iwlwifi-3160-8 \
-    linux-firmware-iwlwifi-3160-9 \
-    linux-firmware-iwlwifi-6000-4 \
-    linux-firmware-iwlwifi-6000g2a-5 \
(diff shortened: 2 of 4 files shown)


jean-edouard:
How about adding `FILES_${PN}-iwlwifi = "/lib/firmware/iwlwifi-*.ucode"` to catch what the individual packages left out?

cjp256:
Why are we still floating our own downstream linux-firmware recipe and not leveraging upstream?

dozylynx:
@cjp256 : I have submitted the packaging change upstream, as I indicated in the comment on OXT-575. In the meantime, we're closing down quickly on the OXT 6.0 release and we should get the firmware into the build to unblock testing on the hardware that needs these firmwares.

I'd like to handle the two tasks of getting the firmware packaged upstream and getting it into current OpenXT builds as separate items.

@jean-edouard : I have a draft patch that creates "linux-firmware-iwlwifi-misc" to catch all the remaining unpackaged firmwares. It's somewhat nicer than stuffing them all into "linux-firmware-iwlwifi" since it allows those misc firmwares to be installed without forcing everything else to also be installed due to the package dependencies.

However, it's basically 22MB of extra stuff, most of which is questionable whether it is of use or benefit. Below is the list. Maybe there are one or two in there we could use? If so, it might make more sense to just add packaging for those rather than adopt iwlwifi-misc.

iwlwifi-1000-3.ucode
iwlwifi-1000-5.ucode
iwlwifi-100-5.ucode
iwlwifi-105-6.ucode
iwlwifi-135-6.ucode
iwlwifi-2000-6.ucode
iwlwifi-2030-6.ucode
iwlwifi-3160-10.ucode
iwlwifi-3160-12.ucode
iwlwifi-3160-13.ucode
iwlwifi-3160-7.ucode
iwlwifi-3160-8.ucode
iwlwifi-3160-9.ucode
iwlwifi-3945-2.ucode
iwlwifi-4965-2.ucode
iwlwifi-5000-1.ucode
iwlwifi-5000-2.ucode
iwlwifi-5000-5.ucode
iwlwifi-5150-2.ucode
iwlwifi-6000-4.ucode
iwlwifi-6000g2a-5.ucode
iwlwifi-6000g2a-6.ucode
iwlwifi-6000g2b-5.ucode
iwlwifi-6000g2b-6.ucode
iwlwifi-6050-4.ucode
iwlwifi-6050-5.ucode
iwlwifi-7260-10.ucode
iwlwifi-7260-7.ucode
iwlwifi-7260-8.ucode
iwlwifi-7260-9.ucode
iwlwifi-7265-10.ucode
iwlwifi-7265-12.ucode
iwlwifi-7265-13.ucode
iwlwifi-7265-8.ucode
iwlwifi-7265-9.ucode
iwlwifi-7265D-10.ucode
iwlwifi-7265D-12.ucode
iwlwifi-7265D-13.ucode

@jean-edouard : let me know what you need.

jean-edouard:
As usual with Intel, figuring out which version belongs to which generation(s) is close to impossible.
I think we need at least all the 6xxx and 7xxx.
I still believe we should include everything, deal with the 22MB disk usage increase and be done with it...

richpersaud:
I vote for including everything.  We may want a wiki page or sample config that advises derivatives how to use bitbake to trim this list to support a narrower HCL.

dozylynx:
OK, have just pushed one more commit into this PR to package all the rest into a -misc package and the virtual package linux-firmware-iwlwifi will pick it up.

cjp256:
@dozylynx I think it's great that you are working on the upstream.  My concern is that bbappends such as this makes rebasing branches more difficult as it is not a "merge compatible" approach.  If OpenXT is looking to be a full blown managed distro, perhaps it makes more sense to maintain OE-core, etc. and cherry-pick upstream commits as desired.  No point in doing the same work twice in different formats?

If maintaining a downstream fork of OE-core is harder than maintaining bbappends, it would be nice to organize these changes in a manner is easy to handle during uprev.  i.e. maintain interfaces with the upstream (which I think you are trying to achieve, but can only do so once accepted upstream) and allow for simply dropping the "backport" bbappends when the branch is upreved.

As far as the content of this PR, I think pulling it all in is fine.  I'm +1, but noting that we figure out a better way in the future.  We should be working in upstream first, where possible, and to that extent, I think a git cherry-pick/merge/rebase strategy makes more sense than the approach used here.  :)

Thank you for doing this work.

rossphilipson:
Cause a build to fail:
http://158.69.127.133:8020/builders/openxt-custom/builds/93

dozylynx:
Fix pushed.
Needed to correctly locate the linux-firmware-iwlwifi-misc package name within the PACKAGES variable.

rossphilipson:
Built here:
http://158.69.127.133:8020/builders/openxt-custom/builds/99
Testing now...

rossphilipson:
All 41 files are there. Looks good, will merge soon.

PR #335, opened 2016-06-29
OXT-637: Package libreSSL for OpenXT

dozylynx:
LibreSSL is packaged as a provider of the openssl interface as an
alternative to the openssl package.

The certificate file cert.pem is packaged independently to allow
installation of the binaries without certificates.

Refs: OXT-637

Upstream: yes, this will also be rebased for submission upstream, independently of this PR.
If it is accepted and made available from upstream, the OpenXT copy should then be removed.

### DIFF
--- a/recipes-connectivity/libressl/libressl.inc
+++ b/recipes-connectivity/libressl/libressl.inc
@@ -1,0 +1,92 @@
+SUMMARY = "Libre Secure Socket Layer"
+DESCRIPTION = "Libre Secure Socket Layer (SSL) binary and related cryptographic tools."
+HOMEPAGE = "http://www.libressl.org/"
+BUGTRACKER = "https://github.com/libressl-portable/openbsd/issues"
+SECTION = "libs/network"
+
+# openssl | SSLeay dual license + ISC for LibreSSL modifications
+LICENSE = "openssl & ISC"
+LIC_FILES_CHKSUM = "file://COPYING;md5=01f9bb4d275f5eeea905377bef3de622"
+
+SRC_URI = "http://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-${PV}.tar.gz \
+          "
+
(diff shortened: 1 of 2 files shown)


jean-edouard:
Did you write the recipes from scratch?
A Google search found some discussions about this, might be worth having a look.

jean-edouard:
Also, what's the motivation for adding this recipe? Especially if we don't build it...

dozylynx:
@jean-edouard :

> what's the motivation for adding this recipe? Especially if we don't build it...

I ended up looking into this when working on OXT-496, to get the pile of new unnecessary CA certs off the filesystem after the Jethro recipe change. Before I figured out that the certs came in via their own package, I was following the hint on the ticket that they were from openssl's libssl, and that package and its recipe are just awful. It was not obvious that it wasn't pulling in a pile of certs from somewhere, and it looked like it might be faster to just swap it out entirely for an alternative, so that's where this recipe came from.

The motivation to add it is to support OpenXT's mission of providing a toolkit for building secure systems. We can't ditch OpenSSL as an option since the FIPS compliance is likely essential for some important derivative projects, but it is quite possible that LibreSSL will suit other derivative projects better than OpenSSL does.

As OpenSSL has been more tested in OpenXT builds so far, it's a sensible default.

I think this recipe ultimately belongs upstream and it's on my list to submit a revision there.

> Did you write the recipes from scratch?

Yes, I did, in order to make the alternative SSL library a build-time preference setting, so you don't have to go and change package dependency definitions to swap between the SSL library options.

> A Google search found some discussions about this, might be worth having a look.

My googling turned up a few attempts to get it included but I couldn't find a recipe that had been accepted in the repos. Any pointers?

jean-edouard:
Thanks for the detailed explanation @dozylynx 
This makes sense, I suggest we merge it in master after branching stable-6.

jean-edouard:
Merging soon.

PR #336, opened 2016-06-30
OXT-576: remove wifi linux-firmware-* from dom0 packagegroup

dozylynx:
No known reason for wifi firmware to be present on the dom0
filesystem so remove it.


rossphilipson:
Built here:
http://158.69.127.133:8020/builders/openxt-custom/builds/102

The firmware is gone from dom0. Will merge soon.

PR #337, opened 2016-07-01
Remove guest rules from rpc-proxy.rules.

stephensmalley:
rpc-proxy.rules still contains a set of a rules permitting obsolete and
removed functionality such as the guest switcher functionality. These
rules were also allowed for all domains, not just guests, so they were
accessible from e.g. the ndvm. We need to remove these rules.

Tested by trying to send one of these messages from the ndvm before
and after the change and confirming that it failed after the change,
with a rpc-proxy: (1->0) DENY message in /var/log/messages.

Saw no other rpc-proxy DENY messages in /var/log/messages, so does not
appear to break any functionality but admittedly did not extensively test.

OXT-639

### DIFF
--- a/recipes-openxt/xenclient/rpc-proxy/rpc-proxy.rules
+++ b/recipes-openxt/xenclient/rpc-proxy/rpc-proxy.rules
@@ -6,21 +6,6 @@ allow stubdom true destination com.citrix.xenclient.surfman
 allow stubdom true destination com.citrix.xenclient.xenmgr
 allow stubdom true destination org.freedesktop.DBus interface org.freedesktop.DBus
 
-# allow guests to call 'gather' on diagnostics interface (required by xc-diag)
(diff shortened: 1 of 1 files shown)


jean-edouard:
This makes sense for Windows, since I don't think it can't even do dbus anymore.
However, won't that break some of the xc-switcher features? We should probably just drop xc-switcher too, but maybe it's worth asking on the mailing list to make sure nobody cares.

stephensmalley:
Isn't that already broken by v4v firewall tightening?  Guests can't send to any dom0 v4v ports anymore...

jean-edouard:
Hah, good point, I already broke it :)
I guess I'll post to the mailing list then!

jean-edouard:
No objection on the mailing list, I created a ticket to track the removal of xc-switcher: https://openxt.atlassian.net/browse/OXT-641

Will merge soon.

PR #338, opened 2016-07-01
[acpi] Move BAT2 status port to prevent AMLI errors in Windows 10

rossphilipson:
The original guest battery ACPI code had used port 0x90 as the BAT2 status
register. For years this was causing harmless messages in the system log but
nothing was done about it and it was forgotten. With Windows 10, use of this
port started causing endless AMLI debugger errors to trace out from the
guest. This moves the port to someplace that is not problematic to Windows.

OXT-606


jean-edouard:
Code LGTM, worked fine on a couple of laptops.
Couldn't test the dual-battery Dell E6430ATG because of OXT-640 though...
https://openxt.atlassian.net/browse/OXT-640

jean-edouard:
Works fine, merging soon.

PR #340, opened 2016-07-01
OXT-630 harden linux kernel configs

dozylynx:
Same config changes applied to all six Linux kernels.
All kernels build successfully. Not currently tested beyond that point.


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/104

This seems to prevent NDVM from creating /dev/tty1, leading to errors in the logs.
It makes sense for NDVM to miss tty nodes and we should fix its /etc/inittab to stop using them.

However, it would be nice to know what caused this, to make sure nothing more important got removed. Maybe CONFIG_STRICT_DEVMEM?

@dozylynx, please advise

jean-edouard:
Interestingly, UIVM still has the tty nodes

jean-edouard:
Here: https://github.com/OpenXT/xenclient-oe/pull/342

dozylynx:
@jean-edouard : honestly not sure what triggered that.
Could be CONFIG_STRICT_DEVMEM as you indicated, or maybe CONFIG_SECURITY_DMESG_RESTRICT, but I think you're right to just drop the tty.

jean-edouard:
@dozylynx, I think we need a better idea of what these options change.
It seems dangerous to merge this without knowing or even understanding the impact of the modifications...

rossphilipson:
I am worried about this one too...

dozylynx:
ok, I'm going to add a comment on OXT-630 with explanation for each change in the config. Each of them seem pretty reasonable; the open question is not so much what the settings do, but why the tty creation was sensitive to those changes.

rossphilipson:
I think the open question is a bit broader than that. If we have an idea what each of the changes does and touches we can make some better estimates about what other things might be affected.

dozylynx:
As commented on OXT-630: I've yanked MPX and DEBUG_STACKOVERFLOW out of the changes here so hopefully the other modifications have consensus to go in.

jean-edouard:
The /dev/tty* nodes issues is actually unrelated.
Everything else looks OK.
Merging soon.

PR #345, opened 2016-07-20
OXT-664: Update OpenSSL to 1.0.1t

dozylynx:
Fixes CVEs.


rossphilipson:
Any thoughts on testing this?

dozylynx:
Fair question.
I think we need to identify what components of ours are using SSL, which would be useful to know anyway. I think the browser in the UIVM is has a dependency on libssl, but I don't know whether we use https transport for our UI, so we might have to coerce it into connecting to an external https site to test it.

OpenSSL can be driven as a command line tool. That's a reasonable way to smoke test that it has at least built a plausible binary.

eric-ch:
It seems reasonable to assume that upgrading openssl micro should be safe.
I recommend we merge this one and have it tested during the RC process since we have no specific test. Testing the RC should provide a good enough coverage regarding the changes involved until someone figures out a better way.

rossphilipson:
@eric-ch I like you suggestion, +1

jean-edouard:
+1, @dozylynx we probably want that in stable-6 too?

rossphilipson:
Yea we are ready to do the RC build, we need to get this in.

jean-edouard:
I created the stable-6 PR, hope you don't mind @dozylynx!

rossphilipson:
Will merge soon.

PR #348, opened 2016-07-20
xen/v4v: add a XSM/Flask v4v use permission check

stephensmalley:
Add a XSM/Flask v4v use permission check to control
the ability to invoke any of the v4v hypercalls.
The existing v4v send permission check only controls the
ability to send via v4v, but still allows use of other v4v
hypercalls, and a subset of the v4v send code path is still
reachable prior to the send check.  The use permission check can
be used to remove the v4v hypercall interface entirely from the
attack surface of the hypervisor for domains that have no legitimate
reason to use v4v.

This change depends on a separate change to xsm-policy to allow v4v
use by authorized domains.

Also update the xsm_v4v_send hook to follow the upstream convention
of including a XSM_DEFAULT_ARG parameter and returning the result of
xsm_default_action() when XSM is disabled; this convention was introduced
in xen commit 0d7f18b01f69c6b89aa3654bd2b11e24f41aaf71 but the v4v XSM
send hook was apparently not updated for it.  This has no impact in
OpenXT since XSM is always enabled, but makes it consistent with upstream.

OXT-666

### DIFF
--- a/recipes-extended/xen/files/openxt-xen-xsmv4vuse.patch
+++ b/recipes-extended/xen/files/openxt-xen-xsmv4vuse.patch
@@ -1,0 +1,140 @@
+Index: xen-4.3.4/xen/common/v4v.c
+===================================================================
+--- xen-4.3.4.orig/xen/common/v4v.c
++++ xen-4.3.4/xen/common/v4v.c
+@@ -1687,7 +1687,7 @@ v4v_send (struct domain *src_d, v4v_addr
+     }
+ 
+   /* XSM: verify if src is allowed to send to dst */
+-  if (xsm_v4v_send(src_d, dst_d) != 0)
++  if (xsm_v4v_send(XSM_HOOK, src_d, dst_d) != 0)
+     {
+       printk(KERN_ERR "V4V: XSM REJECTED %i -> %i\n",
+              src_addr->domain, dst_addr->domain);
+@@ -1798,7 +1798,7 @@ v4v_sendv (struct domain *src_d, v4v_add
+     }
+ 
+   /* XSM: verify if src is allowed to send to dst */
+-  if (xsm_v4v_send(src_d, dst_d) != 0)
++  if (xsm_v4v_send(XSM_HOOK, src_d, dst_d) != 0)
+     {
+       printk(KERN_ERR "V4V: XSM REJECTED %i -> %i\n",
+              src_addr->domain, dst_addr->domain);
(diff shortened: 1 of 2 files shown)


stephensmalley:
Depends on https://github.com/OpenXT/xsm-policy/pull/5

rossphilipson:
At one point there was some discussion about this, maybe an email thread. I have some concerns about changing behaviors on OpenXT and blocking domU access to v4v since there may very well be valid use cases. I am convinced that making this the default (along with shutting down the v4v ports to domU's) is a good approach. I would like to see it documented somewhere that this is the new behavior and if you want to use v4v in a domU you need to create a new XSM policy and open the ports you want. Since we don't have any real "product" docs at the moment, a small wiki would do.

stephensmalley:
Do we want this in 6.0?  If so, it could go in the release notes.
Do we need/want to discuss this on openxt-discuss?

rossphilipson:
6.0 is basically locked down - I am not in favor of putting it in. 6.1 might be OK. I really just want some small bit of docs that indicate this is different and giving a basic idea what you need to do to enable domU v4v access (not down to specific build details, just high level).

stephensmalley:
Is this what you want:
https://openxt.atlassian.net/wiki/pages/viewpage.action?pageId=46989314
Pardon any defects as I really have no experience with Confluence and friends.

rossphilipson:
Yea that is great, thanks. +1

jean-edouard:
LGTM, merging soon.

jean-edouard:
Couple of questions asked in https://github.com/OpenXT/xsm-policy/pull/5, will wait to merge until xsm-policy is ready.

eric-ch:
A small comment in the patch would have been nice for the next person to port the patch-queue in case the history happens to be lost or hard to find, if things get moved, etc. Even 1-line just pointing at some appropriate documentation.

stephensmalley:
Is there a convention or quilt workflow for that?  Don't really see one in other patches there except for the XSAs. The patch itself is just being generated via quilt refresh as per https://openxt.atlassian.net/wiki/display/DC/Dealing+with+quilt ; then I prepare a proper description for the git commit message.  I suppose I could paste the same text into quilt header -e that I use for the git commit message.

eric-ch:
I thought there was a documentation mentioning that somewhere I cannot find it either. And you are right, most of the Xen patches do not have comments either. That was probably discussed when porting the patch-queue to QEMU 1.4.

PR #350, opened 2016-07-21
OXT-665: OpenSSH upgrade to 7.1p2

eric-ch:
Use Jethro's current revision instead of having our own and bbappend it
to get the v4v wrappers and our patches applied and ship.

7.1p2 addresses some CVE.

There is one change in the sshd_config_v4v shipped with UIVM and NDVM that changes
```Match Host dom0
        PermitRootLogin yes```
to
```PermitRootLogin yes```
Otherwise sshv4v to UIVM or NDVM will fail with the server logging:
```error: PAM auth succeeded when it should have failed```

OXT-665


jean-edouard:
- sshd_config_v4v in dom0 still has a `Match Host` on uivm
- each successfull sshv4v connection starts with an X11 forwarding error. It's harmless but doesn't look great...

eric-ch:
@jean-edouard Indeed. It should now be fixed.

jean-edouard:
Merging soon.

PR #351, opened 2016-07-21
init.root-ro: default to "reseal & reboot" instead of "continue"

jean-edouard:
OXT-662


jean-edouard:
Note: this may not be such a good idea, see ticket

PR #352, opened 2016-07-21
xen/v4v: add a XSM/Flask v4v use permission check

stephensmalley:
Add a XSM/Flask v4v use permission check to control
the ability to invoke any of the v4v hypercalls.
The existing v4v send permission check only controls the
ability to send via v4v, but still allows use of other v4v
hypercalls, and a subset of the v4v send code path is still
reachable prior to the send check.  The use permission check can
be used to remove the v4v hypercall interface entirely from the
attack surface of the hypervisor for domains that have no legitimate
reason to use v4v.

This change depends on a separate change to xsm-policy to allow v4v
use by authorized domains.

Also update the xsm_v4v_send hook to follow the upstream convention
of including a XSM_DEFAULT_ARG parameter and returning the result of
xsm_default_action() when XSM is disabled; this convention was introduced
in xen commit 0d7f18b01f69c6b89aa3654bd2b11e24f41aaf71 but the v4v XSM
send hook was apparently not updated for it.  This has no impact in
OpenXT since XSM is always enabled, but makes it consistent with upstream.

OXT-666

(cherry picked from commit 304d2cfed33e175ca1299975c422cddafe18035f)

### DIFF
--- a/recipes-extended/xen/files/openxt-xen-xsmv4vuse.patch
+++ b/recipes-extended/xen/files/openxt-xen-xsmv4vuse.patch
@@ -1,0 +1,140 @@
+Index: xen-4.3.4/xen/common/v4v.c
+===================================================================
+--- xen-4.3.4.orig/xen/common/v4v.c
++++ xen-4.3.4/xen/common/v4v.c
+@@ -1687,7 +1687,7 @@ v4v_send (struct domain *src_d, v4v_addr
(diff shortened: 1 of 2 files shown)


jean-edouard:
This is in the latest master build, works fine there.
The ticket has a fix version of 6.0, so merging soon.

PR #354, opened 2016-07-21
Stable 6: OXT-668: update Intel microcode

dozylynx:
Note: the OpenXT mirror needs to be updated before this PR can be accepted.    

OXT-668: intel microcode update to 20160714
    
    This recipe should be replaced by a dependency on the meta-intel
    layer, but unfortunately at the moment the microcode version in
    the upstream layer is not from 2016. An upstream patch has been
    submitted to correct that and we update the OpenXT version here in
    the meantime.
    
    refs: OXT-668
    
    (cherry picked from commit b6a172f18d2b28d4632b1b9b671607ed8fd44e40)


dozylynx:
Hmm. This looks wrong. Will try again.

PR #356, opened 2016-07-22
OXT-670: rsyslog upgrade to follow OE revision

eric-ch:
Upgrading addresses a couple of CVEs.
Also cleanup the configuration files.

OXT-670


jean-edouard:
LGTM, smoke tested, merging soon.

PR #361, opened 2016-07-25
Stable-6: OXT-665: OpenSSH upgrade to 7.1p2

eric-ch:
Use Jethro's current revision instead of having our own and bbappend it
to get the v4v wrappers and our patches applied and ship.

7.1p2 addresses some CVE.

OXT-665


rossphilipson:
Discussed on ML, tested, ready to go. Merging soon.

PR #362, opened 2016-07-25
Stable-6: OXT-670: rsyslog upgrade to follow OE revision

eric-ch:
Upgrading addresses a couple of CVEs.
Also cleanup the configuration files.

OXT-670


jean-edouard:
This is in the latest master build, works fine there.

rossphilipson:
Discussed on ML, tested, ready to go. Merging soon.

PR #371, opened 2016-07-28
refpolicy: Do not allow sending to dbus unconfined domains by default.

stephensmalley:
Upstream refpolicy commit 6bef7a14757124c56fadc08c255e9dd6c29a15f9
added a reverse rule for the existing rule that allowed dbus unconfined
domains to send to any dbus client in order to allow those dbus clients to
answer.  However, this exposes dbus unconfined domains (i.e. those domains
which are allowed to send to any dbus destination) to any regular
dbus client.  Remove this rule.  Explicit rules on individual domain
pairs can be added if needed.

In the specific case of OpenXT, this opened up D-BUS send from staff_t
or any other dbus client domain to xenmgr (xend_t), since the OpenXT policy
patches make xend_t an unconfined domain.  The user-visible effect was
that xec and xec-vm worked from a staff_t shell rather than requiring
a newrole.  This change restores the prior behavior of denying use of
xec or xec-vm by staff_t.

The longer term fix is to remove unconfined from xend_t and
ultimately remove the unconfined module entirely from the OpenXT
policy and fully confine all domains, but that is likely to have
wider side effects and require substantially more testing.

OXT-686

### DIFF
--- a/recipes-security/refpolicy/refpolicy-mcs-2.20141203/patches/openxt-dbus-deny-send-unconfined.patch
+++ b/recipes-security/refpolicy/refpolicy-mcs-2.20141203/patches/openxt-dbus-deny-send-unconfined.patch
@@ -1,0 +1,18 @@
+Index: refpolicy/policy/modules/contrib/dbus.te
+===================================================================
+--- refpolicy.orig/policy/modules/contrib/dbus.te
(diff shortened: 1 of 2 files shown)


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/124
+1, merging soon.

PR #375, opened 2016-07-29
qemu: fix the iso media change code fix

jean-edouard:
OXT-372
OXT-632


jean-edouard:
The previous fix "re-inserts" the first iso early on, which confuses seabios.

eric-ch:
+1, tested with and without pv-tools. Installing a guest from an iso image worked as well. Will merge soon.

PR #379, opened 2016-08-01
[DO NOT MERGE - this is a fallback PR] - OXT-689 : return EEXIST when reregistering an existing ring

dozylynx:
Removes the code that replaced the MFNs of the existing ring.
It was not holding the necessary write locks to invoke
v4v_ring_remove_mfns so could allow excess decrements to
the page reference counts via concurrent accesses.

### DIFF
--- a/recipes-extended/xen/files/xc-xt-v4v.patch
+++ b/recipes-extended/xen/files/xc-xt-v4v.patch
@@ -135,10 +135,10 @@ index e22a41b..7ad5db0 100644
      int port;
 diff --git xen-4.3.4.orig/xen/common/v4v.c xen-4.3.4/xen/common/v4v.c
 new file mode 100644
-index 0000000..1549ac6
+index 0000000..20af7e8
 --- /dev/null
 +++ xen-4.3.4/xen/common/v4v.c
-@@ -0,0 +1,1976 @@
+@@ -0,0 +1,1974 @@
 +/******************************************************************************
 + * v4v.c
 + * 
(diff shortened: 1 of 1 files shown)


rossphilipson:
What is the effect on the existing drivers (Linux/Windows) when they try to register a ring in these cases and now get an error back?

dozylynx:
ah no, that refresh_pfn_list is not good.

rossphilipson:
It looks like the Windows code just assumes the rings are still intact and goes on during the S4 resume. It does reconnect the event channel but not much else. I guess this change won't cause a problem then.

dozylynx:
an alternative option would be to go and take the write locks in the hypervisor v4v_ring_add and do the MFN replacement that the original version intended to do.

ring (re)registration is more control path than data path, so slowing it down a bit by taking the write locks, in order to preserve the ring content, seems like a reasonable thing to do.

rossphilipson:
Well I think what you have is fine. It is simpler than messing around with locks and it looks like both drivers deal with it.

dozylynx:
I think that in the Linux driver, the guest and the hypervisor end up with a different view of which MFNs the ring occupies after resume, so I would expect the connections to be broken.

That said, if we don't need resume to work because none of the Linux helper VMs ever use S4, then it's not a reason to block this change.

rossphilipson:
Urrg this is tough either way you slice it. I am not sure what to do but knowingly breaking the Linux driver is not desirable. We are not saying you can't use v4v in domU's; our explicit stance is you can but you have to modify policy.

richpersaud:
What is the long-term plan to support non-public Linux Service VMs which need V4V and suspend/resume?  Would that be fixed in 7.0 and could it be backported to 6.1?

Does SyncXT use V4V?

dozylynx:
PR #381 is an alternative to this PR, where the locking has been changed to take the necessary L2 write lock and still refresh the MFNs.

Discussed this work with @rossphilipson and Windows and HVM Linux guests are actually fine with this PR, since guest MFNs will be preserved across S4 and the error logic pasted above in this thread will result in correct behaviour, but PV Linux guests would experience the broken rings after resume.

dozylynx:
We don't want this patch. Closing.

PR #390, opened 2016-08-09
OXT-708: Remove Linux unsupported versions

eric-ch:
Linux 4.4 is currently the main supported kernel version for OpenXT. No need to keep the former ones around, especially without supporting them.

OXT-708


rossphilipson:
Merging soon

PR #392, opened 2016-08-11
OXT-711: Linux upgrade to 4.4.17

eric-ch:
Update recipe, refresh patches and configuration files.
Also changed the source of linux tarball to tar.xz, they are usually
smaller and match with linux-libc-headers recipe. Although that could be
configured to point to gz, OE seems to prefer xz, might as well
follow...

OXT-711


eric-ch:
Requires https://github.com/OpenXT/xenclient-oe/pull/390

eric-ch:
Currently broken (PV guests/service-vms cannot access Xenstore).
Will reopen once fixed.

PR #397, opened 2016-08-17
OXT-709: Use Xen and iucode-tool to load CPU microcode

eric-ch:
Build intel microcode using meta-intel layer as much as possible, ship it in `/boot` and use Xen to load it from the module list using `ucode=` cmdline.
Doing this makes konrad-microcode.patch Linux patch redundant.

Using meta-intel Jethro's branch will not provide the latest version of iucode. So intel-microcode recipe is still cloned until moving to a more recent OE branch.

OXT-709


eric-ch:
Requires https://github.com/OpenXT/openxt/pull/176

rossphilipson:
Probably want to toss this recipe too:
https://github.com/OpenXT/xenclient-oe/tree/master/recipes-core/microcode-ctl

rossphilipson:
My guess is the change to grub is the one causing the merge conflict.

rossphilipson:
Also just a thought but we could probably load the ucode during the install too if we fixup the pxe/isolinux files.

rossphilipson:
@eric-ch  - I installed this build but the microcode changes seem to be missing...
http://158.69.127.133:8020/builders/openxt-custom/builds/135

Not sure what went wrong - maybe the stacked PR thing did not work right. Dunno.

eric-ch:
http://158.69.127.133:8020/builders/openxt-custom/builds/135 only has the Linux upgrade (up to 4.4.18).
I will start a new build with your suggestions and the the conflict taken care of.

rossphilipson:
ACK, the 4.4.18 stuff looks fine, will merge it when I get in.

eric-ch:
Will re-open when fixed.

PR #399, opened 2016-08-18
OXT-720: Upgrade Linux to 4.4.18

eric-ch:
OXT-720
Requires #396 (and former).


rossphilipson:
@eric-ch is building all these here: http://158.69.127.133:8020/builders/openxt-custom/builds/135

rossphilipson:
Will merge this soon with the other two.

PR #401, opened 2016-08-19
xen: sanitize patchqueue

jean-edouard:


rossphilipson:
Confirmed - this is a fuzz buster fix so very important - merging soon.

PR #403, opened 2016-08-24
OXT-709: Use Xen and iucode-tool for intel microcode

eric-ch:
Use the layer introduced by https://github.com/OpenXT/openxt/pull/177 prepare Intel microcode for deployment. As OpenXT aligns on OE Jethro branch, there is still an intel-microcode recipe to get the latest microcode version, this should go away in due time with later upgrades.

Remove the former microcode-ctl deprecated tools used to update the microcode and Konrad's patch to have dom0 capable to pass the microcode to Xen. Instead give the microcode to the bootloader (GRUB, isolinux, pxelinux) and let Xen perform the upgrade at boot using cmdline argument.

OXT-709


MagOXT:
Installation does not work from an usb stick
Loading microcode_intel.bin... failed!
No files found!

eric-ch:
Indeed, tools like Rufus will, by default (and recommend to) re-format the key and not use the isohybrid as it is (like dumping the ISO with dd on the stick). So microcode-intel.bin will not always be renamed microcode_intel.bin. Might as well give it a name that will not get mangled by the ISO9660 norm.

rossphilipson:
Started a build for retesting this:
http://158.69.127.133:8020/builders/openxt-custom/builds/149

MagOXT:
Installation from a usb stick succeed this step.
Loading microcode_intel.bin... ok

rossphilipson:
No error messages about microcode failures during the install. Magali, if you think it is ready I will merge it?

MagOXT:
You can merge it.

rossphilipson:
Ok so testing shows this is working. Note there is some confusion because in Xen 4.3.4, messages were traced out even when the microcode was not updated. Those messages are now gone (Xen 4.6 microcode update for Intel disables those messages with #define pr_debug(x...) ((void)0)). So on systems where there are no ucode updates you see nothing. One ones where there are you see:

 (XEN) microcode: CPU2 updated from revision 0x1c to 0x1f, date = 2016-04-01

Anyway, all working, wil merge soon...

PR #406, opened 2016-08-29
STABLE-5: refpolicy: fix OTA-related denials

stephensmalley:
Allow dmsetup and mount to statfs sysfs and proc filesystems.
Allow updatemgr to statfs ext4 filesystems.
Allow updatemgr to read /config/repo-cert.conf.

The last one should be allowed already but several /config files
are mislabeled in 5.x installs.  We aren't likely to fix that
now so we might as well just allow it here for OTA purposes.

### DIFF
--- a/recipes-security/selinux/refpolicy-mcs/openxt-ota-fixes.patch
+++ b/recipes-security/selinux/refpolicy-mcs/openxt-ota-fixes.patch
@@ -1,0 +1,75 @@
+Index: refpolicy/policy/modules/services/updatemgr.te
+===================================================================
+--- refpolicy.orig/policy/modules/services/updatemgr.te	2016-09-01 16:06:27.000000000 -0400
++++ refpolicy/policy/modules/services/updatemgr.te	2016-09-01 16:06:28.000000000 -0400
+@@ -81,3 +81,7 @@
+ allow updatemgr_t updatemgr_tmp_t:file manage_file_perms;
+ allow updatemgr_t self:capability { dac_override chown fowner fsetid };
+ 
++# OTA upgrades
++fs_getattr_xattr_fs(updatemgr_t)
++xc_read_config_files(updatemgr_t)
++selinux_get_fs_mount(updatemgr_t)
+Index: refpolicy/policy/modules/system/lvm.te
+===================================================================
+--- refpolicy.orig/policy/modules/system/lvm.te	2016-09-01 16:06:25.000000000 -0400
++++ refpolicy/policy/modules/system/lvm.te	2016-09-01 16:06:28.000000000 -0400
+@@ -411,3 +411,9 @@
+ 	updatemgr_dontaudit_rw_stream_sockets(lvm_t)
+ 	xc_installer_read_tmp_files(lvm_t)
+ ')
++
++# OTA upgrades
++kernel_getattr_proc(lvm_t)
++allow lvm_t sysfs_t:filesystem getattr;
++# For reading /run/udev/queue.bin
++udev_read_pid_files(lvm_t)
+Index: refpolicy/policy/modules/system/mount.te
+===================================================================
+--- refpolicy.orig/policy/modules/system/mount.te	2016-09-01 16:06:25.000000000 -0400
++++ refpolicy/policy/modules/system/mount.te	2016-09-01 16:06:28.000000000 -0400
+@@ -267,3 +267,7 @@
+ optional_policy(`
+ 	xc_dontaudit_rw_v4v_chr(mount_t)
+ ')
++
++# OTA upgrades
++kernel_getattr_proc(mount_t)
++allow mount_t sysfs_t:filesystem getattr;
+Index: refpolicy/policy/modules/system/xc-installer.if
+===================================================================
+--- refpolicy.orig/policy/modules/system/xc-installer.if	2016-09-01 16:06:28.000000000 -0400
++++ refpolicy/policy/modules/system/xc-installer.if	2016-09-01 16:06:28.000000000 -0400
+@@ -48,6 +48,7 @@
+ 	corecmd_search_bin($1)
+ 	domtrans_pattern($1, updatemgr_storage_t, xc_installer_t)
+ 	domtrans_pattern($1, xc_installer_exec_t, xc_installer_t)
++	allow $1 xc_installer_t:process { noatsecure siginh rlimitinh };
+ ')
+ ########################################
+ ## <summary>
+@@ -61,10 +62,11 @@
+ #
+ interface(`xc_installer_delete',`
+ 	gen_require(`
+-		type xc_installer_t;
++		type xc_installer_storage_t;
+ 	')
+ 
+-	allow $1 xc_installer_t:file delete_file_perms;
++	allow $1 xc_installer_storage_t:dir manage_dir_perms;
++	allow $1 xc_installer_storage_t:file delete_file_perms;
+ ')
+ ########################################
+ ## <summary>
+Index: refpolicy/policy/modules/system/xc-installer.fc
+===================================================================
+--- refpolicy.orig/policy/modules/system/xc-installer.fc	2016-09-01 16:06:28.000000000 -0400
++++ refpolicy/policy/modules/system/xc-installer.fc	2016-09-01 16:06:28.000000000 -0400
+@@ -18,5 +18,5 @@
+ #
+ #############################################################################
+ 
+-/storage/update/upgrade		--	gen_context(system_u:object_r:xc_installer_t,s0)
++/storage/update/upgrade		--	gen_context(system_u:object_r:xc_installer_storage_t,s0)
+ /usr/share/xenclient/post-upgrade.sh	--	gen_context(system_u:object_r:xc_installer_exec_t,s0)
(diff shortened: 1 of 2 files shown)


rossphilipson:
OXT-734

stephensmalley:
FYI, I wasn't able to test an actual OTA yet.  For some reason my stable-5 build doesn't bring up the ndvm successfully.  So this is just based on the denials reported by Jed.
Also, I'm not clear on how one kicks off an OTA in the first place.

stephensmalley:
Tried testing a 6.0 -> master upgrade by using the Settings Software Update dialog and pointing it to my master repo on my build server.  It downloaded 100%, then said download successful, restart your computer, and then immediately showed another download progress indicator and said download failed.  Doesn't seem to be SELinux-related since it happens even in permissive.  What am I doing wrong?

rossphilipson:
I was trying to compile the ref policy in my stable-5 tree and it fails to compile. It looks like dev_getattr_sysfs is not defined in that older version.

stephensmalley:
Based on the denials he posted, you want:
allow lvm_t sysfs_t:filesystem getattr;
allow mount_t sysfs_t:filesystem getattr;

It was tclass=filesystem not dir.

The weird thing is I did build it on stable-5.  But apparently because I didn't do a cleansstate in between it didn't truly rebuild it.

rossphilipson:
Ack, I just tried it and it failed with AVCs. I will try with your new ones.

stephensmalley:
Pushed an updated commit.
Were any of the AVCs new/different?

rossphilipson:
I just grabbed your updated patch and built with it. These were the AVCs:

[  353.514159] audit: type=1400 audit(1472498357.050:1458): avc:  denied  { getattr } for  pid=2254 comm="dmsetup" name="/" dev="sysfs" ino=1 scontext=system_u:system_r:l0
[  353.514669] audit: type=1400 audit(1472498357.050:1459): avc:  denied  { getattr } for  pid=2254 comm="dmsetup" name="/" dev="sysfs" ino=1 scontext=system_u:system_r:l0
[  353.522492] audit: type=1400 audit(1472498357.060:1460): avc:  denied  { getattr } for  pid=2255 comm="mount" name="/" dev="sysfs" ino=1 scontext=system_u:system_r:mou0
[  353.522716] audit: type=1400 audit(1472498357.060:1461): avc:  denied  { getattr } for  pid=2255 comm="mount" name="/" dev="sysfs" ino=1 scontext=system_u:system_r:mou0
[  353.527817] audit: type=1400 audit(1472498357.060:1462): avc:  denied  { getattr } for  pid=2256 comm="dmsetup" name="/" dev="sysfs" ino=1 scontext=system_u:system_r:l0
[  353.527931] audit: type=1400 audit(1472498357.060:1463): avc:  denied  { getattr } for  pid=2256 comm="dmsetup" name="/" dev="sysfs" ino=1 scontext=system_u:system_r:l0
[  353.750132] audit: type=1400 audit(1472498357.280:1464): avc:  denied  { getattr } for  pid=2301 comm="dmsetup" name="/" dev="sysfs" ino=1 scontext=system_u:system_r:l0
[  353.750261] audit: type=1400 audit(1472498357.280:1465): avc:  denied  { getattr } for  pid=2301 comm="dmsetup" name="/" dev="sysfs" ino=1 scontext=system_u:system_r:l0
[  353.772642] audit: type=1400 audit(1472498357.310:1466): avc:  denied  { getattr } for  pid=2313 comm="vgs" name="/" dev="sysfs" ino=1 scontext=system_u:system_r:lvm_t0
[  353.772805] audit: type=1400 audit(1472498357.310:1467): avc:  denied  { getattr } for  pid=2313 comm="vgs" name="/" dev="sysfs" ino=1 scontext=system_u:system_r:lvm_t0

stephensmalley:
Your logs are corrupted.  Do you see the same in dmesg output?
Those messages are truncated.

rossphilipson:
BTW this build should be used, it allows upgrades from 5.0.1:

http://openxt.ainfosec.com/builds/stable-6/custom-dev-141-stable-6/repository/

Until we sort out the build dev certs, you also have to short circuit this file:

/usr/bin/verify-repo-metadata

On line 89 I just add: VALID=1

stephensmalley:
Ok, how are you testing?  Via the Settings Software Update GUI or is there some command-line way to invoke the OTA upgrade?

rossphilipson:
Using the UI. Literally using that URL above. That part worked - it was happy to upgrade from 5.0.1. It died later on the AVCs. Trying again with your fixed patch.

rossphilipson:
Ok I think I got farther but something made udev very unhappy while trying to create lv's. I suspect there are still SELinux issues. Things were so hosed I could not really get a clear picture of what happened. Trying again in permissive mode.

rossphilipson:
I switched to using the ticket since formatting is easier there:

https://openxt.atlassian.net/browse/OXT-734

Noted some AVCs we are still getting.

stephensmalley:
Does the updated commit resolve it?
What needs to be removed/rebuilt if you change the certs in your build?
Tried replacing the ones in my stable-5 tree with the ones from my stable-6 tree so they would match and now I can't even successfully install my stable-5 build; signature verification fails during the install.

stephensmalley:
Does the updated commit resolve it?
What needs to be removed/rebuilt if you change the certs in your build?
Tried replacing the ones in my stable-5 tree with the ones from my stable-6 tree so they would match and now I can't even successfully install my stable-5 build; signature verification fails during the install.

rossphilipson:
we are trying to rectify the certificate issue - at least for the dev certs. We have updated the old builder with the new one's certs and Jed is working on doing an "official" 4.0.0 build (and a 5.0.1 later). That way all release versions will have the same dev cert.

I will try your updated patch - get back to you on that, thanks.

rossphilipson:
Still not working though those AVC's are gone. We found a Xen AVC that we fixed in 6.0.0+

https://github.com/OpenXT/xsm-policy/pull/8

rossphilipson:
Updated the ticket with a bit more information. There is one other bunch of AVCs related to acpid when the platform starts. I don't see how that could be related but may be worth investigating (since we have nothing else at this point).

stephensmalley:
There are two things we can do:
- Patch the kernel to not drop audit messages (it ratelimits them unless running auditd),
- rebuild the policy to strip dontaudit rules (normally silence noisy messages that you don't want to see but could be hiding actual issues).

The latter will produce lots of denials, which is where the former becomes importance since otherwise you might lose the ones you care about due to the noisy ones.  The latter can be done on an existing system by running:
nr
rw
semodule -DB
And can be reverted by running semodule -B later.

rossphilipson:
Point 1:
I actually started out today by trying to disable rate limit in the kernel and got side tracked looking at using auditd. I did use it and saw a bit more activity. I did not see anything happen (permissive or enforcing) at or around the time the upgrade failed. Maybe I need to configure auditd better.

Point 2:
Sounds like that is worth trying - I will look at that tomorrow. Thanks.

rossphilipson:
I suggest we take this PR now since it clearly fixes most of the SELinux problems. There are some still outstanding but if we get a build with this in it, it will make things easier. Merging in a while if there are no objections.

stephensmalley:
Up to you - I can either keep refreshing this one with new denial fixes or add those as a separate PR.  Currently have a build going with auditd added and switching several dontaudits to allows for mount_t and lvm_t to updatemgr_t since I saw those denials with semodule -DB.  Not sure though if those are the culprits since they were explicitly dontaudited by XenClient/OpenXT policy.

rossphilipson:
Ok well since you have something in the works I will keep it open for now. Note that this build ran last night with Jed's XSM fix so it is probably a better one to use for testing:

http://openxt.ainfosec.com/builds/stable-6/oxt-dev-158-stable-6/

rossphilipson:
Sorry, was confused - the XSM fix went into stable-5. We don't have a newer version of that yet. The build mentioned above is newer so it should be used anyway.

rossphilipson:
Commented on ticket and added a log file

stephensmalley:
Let's try this one.  Presently updatemgr policy always dontaudit's attempts to access open fds, pipes, and sockets it creates.  This is odd since I'd expect it to collect error output from any programs it execs for display to the user, but maybe it just ignores them anyway.  So switch those dontaudits to allows and at least get rid of that noise.  Running auditd too will help but that isn't part of this change.

rossphilipson:
Ok cool. I have a half arsed version of what you are doing there but I will use yours. It looks like it will take care of more cases than my hokey one.

rossphilipson:
That did not work either - still fails in the same place :(

After lunch I will disable rate limit in the kernel.

stephensmalley:
https://github.com/stephensmalley/xenclient-oe/commit/e9fcd2f8f54cd31621829650f3cc55aa9eb97c3d

stephensmalley:
Try this one.

rossphilipson:
Looks to be working, great, thanks! We should probably rename those dontallow macros in policy/modules/services/updatemgr.if. Something like:

updatemgr_dontaudit_rw_stream_sockets -> updatemgr_allow_rw_stream_sockets
updatemgr_dontaudit_rw_fifo_files -> updatemgr_allow_rw_fifo_files
updatemgr_dontaudit_use_fd -> updatemgr_allow_use_fd

stephensmalley:
First I'll try reverting/dropping that part of the change, as it was never clear if that was needed.
If needed, then I'll rename the interfaces but then I also have to update all the callers.

rossphilipson:
Ack, that makes sense.

rossphilipson:
I just tried it without the changes to policy/modules/services/updatemgr.if and the upgrade was fine. After the reboot there was only one avc in the log from the NDVM.

stephensmalley:
Ok, so hopefully the updated commit is good to go.

rossphilipson:
Cool. I will give it one more run after lunch and if that looks good I will merge it.

rossphilipson:
Tested latest, upgrade worked (at least no SELinux issues). Merging soon.

PR #412, opened 2016-08-30
[xdotool] Amend source address.

eric-ch:
http://158.69.127.133:8010/builders/openxt/builds/156


eric-ch:
Duplicates https://github.com/OpenXT/xenclient-oe/pull/407

PR #415, opened 2016-08-31
linux: add a couple of USB NIC drivers for the installer

jean-edouard:


eric-ch:
Will merge soon

dozylynx:
Is there a JIRA ticket relevant to this change?

If the Installer needs USB network device drivers in the kernel then it would make sense to have distinct separate kernels for the Installer and Dom0, to enable reduction of the Dom0 attack surface by omitting the unnecessary drivers from the Dom0 kernel.

jean-edouard:
dom0 doesn't include any of the NIC drivers.
They are all built as modules and included only in the installer.

dozylynx:
thanks, @jean-edouard -- excellent.

PR #417, opened 2016-09-06
STABLE-6: refpolicy: fix OTA-related denials

stephensmalley:
Allow updatemgr to trigger module auto-load if needed.

Allow updatemgr to manage the /storage/update/upgrade-data directory;
this was supposed to already be allowed but the wrong type was used in
the xc-installer.if and xc-installer.fc files.

Allow updatemgr to convey environment variables, signal
masks, and resource limits to the installer across a domain transition.

OXT-734

### DIFF
--- a/recipes-security/refpolicy/refpolicy-mcs-2.20141203/policy/modules/services/updatemgr.te
+++ b/recipes-security/refpolicy/refpolicy-mcs-2.20141203/policy/modules/services/updatemgr.te
@@ -82,3 +82,5 @@ allow updatemgr_t updatemgr_sync_client_storage_t:file manage_file_perms;
 allow updatemgr_t updatemgr_tmp_t:file manage_file_perms;
 allow updatemgr_t self:capability { dac_override chown fowner fsetid };
 
+kernel_request_load_module(updatemgr_t)
+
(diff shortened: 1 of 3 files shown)


stephensmalley:
With just these changes, I was able to successfully update from one build of stable-6 to another build of stable-6.  I still cannot update from stable-6 to master (fails on second download) but that doesn't appear to be SELinux-induced.

rossphilipson:
How are you upgrading from one version of stable-6 to another? You must be fiddling with the up-gradable version value somewhere? I am curious.

Also I will fix the title of the ticket to cover more than 5.0.1 to 6.0

stephensmalley:
I didn't change the version string, so they are both the same version.  Also, I am doing incremental builds rather than from-scratch builds every time, so I don't think it updates the build string (not sure if that matters). The software update dialog says that I am already running the latest version but may refresh it by downloading and applying it again, so I did that, and that seemed to succeed.

rossphilipson:
This build has the commit to allow it to upgrade from stable-6:
http://158.69.127.133:8010/builders/openxt/builds/162

rossphilipson:
I think this is valid on its own. The download failure seems to be unrelated to SELinux but without this patch you get nowhere. I am going to merge this.

PR #418, opened 2016-09-07
[upgrade] STABLE-4: Fix updatemgr to allow access to /config files.

rossphilipson:
OXT-734


rossphilipson:
Tested with this build and the SELinux change:
http://158.69.127.133:8040/builders/openxt-legacy2/builds/39

Upgrade works without problems or signature verification failures.

jean-edouard:
Yup, this is a subset of https://github.com/OpenXT/xenclient-oe/pull/406 which was successfully tested and merged.

Note: this is a workaround, the ideal way to fix this would be to fix the SELinux label on `/config/repo-cert.conf`.

Merging soon.

PR #419, opened 2016-09-08
XSA 183, 185 and 187

eric-ch:
XSA-184 is not in the scope of OpenXT (https://xenbits.xen.org/xsa/advisory-184.html).
XSA-186 does not affect Xen 4.6.1 (https://xenbits.xen.org/xsa/advisory-186.html)

OXT-742


rossphilipson:
Merging soon.

PR #420, opened 2016-09-08
STABLE-6: Xen XSA-185 and 187.

eric-ch:
XSA-183 does not affect Xen 4.3.4 (https://xenbits.xen.org/xsa/advisory-183.html)
XSA-184 is not in the scope of OpenXT (https://xenbits.xen.org/xsa/advisory-184.html).
XSA-186 does not affect Xen 4.3.4 (https://xenbits.xen.org/xsa/advisory-186.html)

OXT-742


rossphilipson:
Built here:
http://158.69.127.133:8020/builders/openxt-custom/builds/153

rossphilipson:
Smoke tested, looks good. Merging soon.

PR #421, opened 2016-09-08
refpolicy: fix OTA-related denials

stephensmalley:
Allow updatemgr to trigger module auto-load if needed.

Allow updatemgr to manage the /storage/update/upgrade-data directory;
this was supposed to already be allowed but the wrong type was used in
the xc-installer.if and xc-installer.fc files.

Allow updatemgr to convey environment variables, signal
masks, and resource limits to the installer across a domain transition.

OXT-734


stephensmalley:
This is the same fix for master

rossphilipson:
Yes that makes sense - same issues would be there. Will merge soon.

PR #423, opened 2016-09-19
[xen] Remove xentrace format patch.

rossphilipson:
After further consideration, this patch doesn't seem to make much sense and the
tool it patches is not used.

OXT-577


jean-edouard:
LGTM, merging soon.

PR #425, opened 2016-09-20
[xen] Comment SEABIOS patch, remove dead VGA target.

rossphilipson:
The force pop issue cannot be resolved cleanly and the reason
why is in the comments.

OXT-746


jean-edouard:
LGTM, merging soon.

PR #428, opened 2016-09-28
initscripts: convert do_install to do_install_append

dpsmith:
This is to resolve OXT-757

### DIFF
--- a/recipes-core/initscripts/initscripts_1.0.bbappend
+++ b/recipes-core/initscripts/initscripts_1.0.bbappend
@@ -27,6 +27,7 @@ do_install () {
 
 	install -m 0644    ${WORKDIR}/functions		${D}${sysconfdir}/init.d
 	install -m 0755    ${WORKDIR}/bootmisc.sh	${D}${sysconfdir}/init.d
+	install -m 0755    ${WORKDIR}/checkroot.sh	${D}${sysconfdir}/init.d
 	install -m 0755    ${WORKDIR}/halt		${D}${sysconfdir}/init.d
 	install -m 0755    ${WORKDIR}/hostname.sh	${D}${sysconfdir}/init.d
 	install -m 0755    ${WORKDIR}/mountall.sh	${D}${sysconfdir}/init.d
@@ -55,6 +56,7 @@ do_install () {
 	update-rc.d -r ${D} umountfs start 40 0 6 .
 	update-rc.d -r ${D} reboot start 90 6 .
 	update-rc.d -r ${D} halt start 90 0 .
+	update-rc.d -r ${D} checkroot.sh start 06 S .
 	update-rc.d -r ${D} save-rtc.sh start 25 0 6 .
 	update-rc.d -r ${D} mountall.sh start 35 S .
 	update-rc.d -r ${D} hostname.sh start 39 S .
--- a/recipes-core/base-files/base-files/fstab.xenclient-dom0
+++ b/recipes-core/base-files/base-files/fstab.xenclient-dom0
@@ -1,4 +1,4 @@
-rootfs                        /             auto   defaults,noatime                                             1  1
+rootfs                        /             auto   ro,defaults,noatime                                          1  1
 proc                          /proc         proc   defaults                                                     0  0
 devpts                        /dev/pts      devpts mode=0620,gid=5                                              0  0
 tmpfs                         /var/volatile tmpfs  defaults,rootcontext=system_u:object_r:var_t:s0,size=2M      0  0


jean-edouard:
That comment is not true anymore, it should probably be removed:
`# Override to reduce the number of scripts installed`

The comment brings up a good point, a bunch of "new" initscripts will get installed, we have to make sure it doesn't break anything. I'll start a custom build.

dpsmith:
Did a comparison and the additional init scripts are,
- checkroot.sh (this is the one meta-selinux is looking for)
- mountnfs.sh (while its benign, don't think we want it in the environment)
- umountnfs.sh (same as above)
- devpts.sh (setups /dev/pts if devpts filesystem is available)
- read-only-rootfs-hook.sh (setups up volatile tmpfs and bind mounts to /var/lib)
- sysfs.sh (setups proc, sysfs, debugfs,devtmpfs)
- dmesg.sh (rotates dmesg log which is unnecessary)
- banner.sh (prints booting message to console)


Updates/changes that are lost from xenclient-oe maintaining install
- moved file that holds urandom info
- fix to urandom startup script for new location
- a environment fix for when selinux is being included

Options
- Move back to overriding install and ensure we keep it up to date
- Follow the systemd approach and have a MASKED_SCRIPTS that during post_install removes the scripts we do not want

Both options require maintenance that entails tracking oe-core's initscripts recipe

jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/170

jean-edouard:
This PR seems to break things... See previous build, networking doesn't work.

jean-edouard:
LGTM, will build to test.
Why do you add `read-only-rootfs-hook.sh` to `SRC_URI`? It should already have it...

dpsmith:
Summary of the approach is in the commit comment but will provide a few more details here. In the end, `read-only-rootfs-hook.sh` is the script that is creating the conflict by it trying to move `/var/lib` onto `/var/volatile/lib` by copying everything in `/var/lib/` into `/var/volatile/lib` and bind mounting. Instead of writing tasks that tried to copy what we want or subtract what we don't, I made a copy of the script that disables the copy logic and because we prepend the `FILESEXTRAPATH` the OpenXT version should be the one that gets installed instead of the OE version.

jean-edouard:
Yup, so no need to re-add it to SRC_URI no?

dpsmith:
Not sure, when double checking to make sure I did it right the advice said to add it to SRC_URI in your bbappend.

jean-edouard:
This is weird... It's just a variable with some string content, that obviously already has that line...
But if it's advised to do it that way and doesn't warn/break, I guess it doesn't matter.
Will test next week. Thanks.

dpsmith:
Agreed. I may test with and without just to see if it does make a difference.

jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/176
Doesn't boot at all now...

dpsmith:
I will pull your build and debug to see why.

dpsmith:
This kind of snowballed, so pulling back for now and fixing to be compatible with meta-selinux.

recipes-core/initscripts/initscripts_1.0.bbappend
```diff
@@ -55,6 +56,7 @@ do_install () {
 	update-rc.d -r ${D} umountfs start 40 0 6 .
 	update-rc.d -r ${D} reboot start 90 6 .
 	update-rc.d -r ${D} halt start 90 0 .
+        update-rc.d -r ${D} checkroot.sh start 06 S .
```
jean-edouard: Indent should be tab

dpsmith: I replaced with tab to align with neighbors.

jean-edouard:
Didn't pass basic testing, `/` gets remounted in read-write mode.

At this point I recommend you close this PR, add more information to the ticket, start from scratch, test your work and submit a nice fresh PR.

dpsmith:
This was a simple issue of a reliance on bad behavior by the system and does not need any serious rethink, it just needs to be fixed. Most sysv init set ups will do a `mount -a` to ensure all mounts are mounted according to the fstab. The dom0 fstab has never explicitly specified that / is ro and relied on the fact that kernel command line initially setup rootfs as ro. This has been corrected in revised commit, testing will be addressed on mailing list.

dozylynx:
Latest version of this PR applied locally to master, built, installed and tested.

********************************************************************************
Tests executed on the system, and passed:

- System installs

- System boots
    - Measured Launch indicator shows success

- Can open a terminal (Ctrl-H and Ctrl-T)

- System can be configured with network access

- File system configuration:
    - / is ro
    - /etc/init.d/checkroot.sh is present and has executable permission
    - /etc/rcS.d/S06checkroot.sh symlink is present and points to init.d copy

- Swap is active
    - cat /proc/swaps
    /dev/mapper/swap\040(deleted)   partition   262140  0   -1

- Daemons are running in dom0:
    - xenmgr
    - xenvm (one per defined vm)
    - dbd
    - xenstored
    - uid
    - xcpmd
    - input-server
    - rpc-proxy (3 instances)
    - vusb-daemon
    - network-daemon

- SELinux enforcing mode in both dom0 and ndvm:
    * getenforce : Enforcing
    * cat /sys/fs/selinux/enforce : 1

- XSM/Flask enforcing mode
    * xenops getenforce : Enforcing

 - Create new VM
    - Used Linux template; boot to SeaBIOS without install media succeeded.

- Domains have the expected XSM/Flask label
    * xec-vm -n Network -g flask-label
        system_u:system_r:ndvm_t

    * xec-vm -n uivm -g flask-label
        system_u:system_r:uivm_t

    * xec-vm -n TestVM -g flask-label
        system_u:system_r:hvm_guest_t
- No kernel classes/permissions left undefined in the SELinux policy
    * grep "SELinux.*not defined" /var/log/messages

- Processes have the expected SELinux label in dom0:
    * ps -eZ
      * only kernel threads have kernel_t,
      * only init has init_t,
      * only logger is left running in initrc_t,

      * every daemon has its own domain
        - [majority pass, see exceptions below]

- Processes have the expected SELinux label in ndvm:
    * ps -eZ
      * only kernel threads have kernel_t,
      * only init has init_t,
      * only logger is left running in initrc_t,
        -> logger is not running in the ndvm
      * every daemon has its own domain

********************************************************************************
Tests executed on the system with tolerable failure, unlikely caused by this PR:

- SELinux label inspection:
`restorecon -Rnv /config /storage /boot`
"-n means do not make any changes, -v means report what would be changed to make the file labels match the file_contexts configuration.'

Result:
`restorecon reset /config/xenclient.conf context system_u:object_r:db_backup_t:s0->system_u:object_r:xc_config_t:s0`

- No unexpected denials from SELinux or XSM/Flask (grep avc:
/var/log/messages - not currently clean on master, which needs to be
fixed, but at least don't introduce any new ones).

Result:
Single denial: avc denied for sys_resource for xenconsoled

- Processes have the expected SELinux label in dom0:
      * every daemon has its own domain

Result:
observation: dbus-monitor is running with system_u:system_r:language_sync_t

observation: xenmgr is running with system_u:system_r:xend_t
             and so are xenvm processes

********************************************************************************
Tests not executed:

None

********************************************************************************

**Verdict:** This PR looks correct and safe to apply.

rossphilipson:
OK that is quite a bit of testing and validated. I think this is good to go, merging soon.

Those other two issues should probably be tickets right?

dozylynx:
@rossphilipson : yes, I'll add tickets in now.

PR #431, opened 2016-10-04
[xen] Remove unused hvm-info.patch and move XSM patch before XSAs.

rossphilipson:
XSAs and other backports should always be at the end of the patch Q to allow
for easy removal during uprevs.

OXT-597


rossphilipson:
Building it here:

http://158.69.127.133:8020/builders/openxt-custom/builds/177

rossphilipson:
Built and installed fine. Should be good to go.

PR #433, opened 2016-10-05
[xen] XSA-190 patch applied to Xen 4.3.4 for OpenXT stable-6

dozylynx:
http://xenbits.xen.org/xsa/advisory-190.html

### DIFF
--- a/recipes-extended/xen/files/xsa-190-x86-hvm-instr-emul-guest-user-mode-missing-exception.patch
+++ b/recipes-extended/xen/files/xsa-190-x86-hvm-instr-emul-guest-user-mode-missing-exception.patch
@@ -1,0 +1,189 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+XSA-190 http://xenbits.xen.org/xsa/advisory-190.html
+x86 HVM : instruction emulator is missing a required exception.
+
+################################################################################
+LONG DESCRIPTION:
+################################################################################
+Source: http://xenbits.xen.org/xsa/advisory-190.html
+Patch: xsa-190-x86-hvm-instr-emul-guest-user-mode-missing-exception.patch
+
+x86emul: honor guest CR0.TS and CR0.EM
+
+We must not emulate any instructions accessing respective registers
+when either of these flags is set in the guest view of the register, or
+else we may do so on data not belonging to the guest's current task.
+
+Being architecturally required behavior, the logic gets placed in the
+instruction emulator instead of hvmemul_get_fpu(). It should be noted,
+though, that hvmemul_get_fpu() being the only current handler for the
+get_fpu() callback, we don't have an active problem with CR4: Both
+CR4.OSFXSR and CR4.OSXSAVE get handled as necessary by that function.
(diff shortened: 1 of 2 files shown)


rossphilipson:
Can this go into the 6.1.0 release?

dozylynx:
Yes, it should go in 6.1.0. Thanks.

richpersaud:
Maybe it can go into a 6.0.1 security-only release, since 6.1.0 will have a functional change (ML).

pearsonk:
I think 6.1.0 is fine. Don't need to cut a formal release for every XSA

richpersaud:
We definitely don't want a formal release per XSA, but there may be future security fixes to the stable-6 branch, which could affect any component including Linux kernel or Xen.   If such fixes appear, they can be added to the stable-6 branch, or be implicitly included via upstream OE Jethro.  

After Jethro becomes unmaintained (https://wiki.yoctoproject.org/wiki/Stable_branch_maintenance) with the release of OE Morty at the end of this month, we may decide to backport some security fixes from upstream Linux or Xen, at least until OpenXT 7.0 is released.  After 7.0 is released, we can optionally cut a formal relese of 6.0.1.

The open question is whether we force people who want stable-6 security fixes to also accept a large functional change (Measured Launch is a sizable changeset) that affects upgrade, or whether ML/6.1.0 can go into a different stable-6 branch.  Since ML was delayed and is probably a one-off situation, a separate branch would provide a safe space for this large change to be stabilized and tested on a range of hardware.  ML code is unlikely to overlap with Linux and Xen security fixes, so it would not be difficult for either OpenXT or derivatives to merge security fixes from 6.0.1 into 6.1.0.

eric-ch:
LGTM, either for 6.0.x or for 6.x.0, that should be merged in stable-6 branch. Is anyone against that?

rossphilipson:
@eric-ch - it is fine to merge into the 6.x branch now - regardless of what the next build ends up being.

@richpersaud - I think it comes down to having the resources to maintain numerous branches. The short answer is we really don't. Also every release means a more complex upgrade path. I would say if a serious security problem comes out before 6.1.0 is ready we could do a 6.0.1 but otherwise we should focus on 6.1.0. Remember we are only doing 6.1.0 because ML did not land on time for 6.0.0 and we it seems to me we should consider the ML work a security fix itself.

richpersaud:
Ok, a better story on security updates will have to wait until someone volunteers to maintain and test OpenXT branches that contain only security fixes, where security fixes are defined by the security process of upstream Linux, Xen and OpenEmbedded.

richpersaud:
For 6.1.0 release notes, we need to document that the new ML implementation removes functionality available in 6.0.0, e.g. recovery keys.  This will help developers to make an informed choice about the pro/con of migrating from 6.0.0 to 6.1.0, or manually backporting security-only fixes from 6.1.0 to 6.0.0.

rossphilipson:
There is already a ticket for 6.1.0. You can add that information there:
https://openxt.atlassian.net/browse/OXT-714

PR #435, opened 2016-10-11
Llinux upgrade to 4.4.24

eric-ch:
OXT-773


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/186
LGTM, merging soon.

PR #436, opened 2016-10-11
[xen] Increase AP startup timeout when TBOOT launches Xen.

rossphilipson:
OXT-747


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/186
LGTM, merging soon.

PR #438, opened 2016-10-25
refpolicy: allow xenconsoled to increase fd limit

stephensmalley:
xenconsoled attempts to increase its fd limit, triggering the following
SELinux denial:
  avc:  denied  { sys_resource } for  pid=919 comm="xenconsoled" capability=24  scontext=system_u:system_r:xenconsoled_t:s0 tcontext=system_u:system_r:xenconsoled_t:s0 tclass=capability permissive=0
and the following log message from xenconsoled:
  xenconsoled: Unable to increase fd limit from {1024, 4096} to {132008, 132008}: (Operation not permitted) - May run out with lots of domains

Allow this capability for xenconsoled.

OXT-784


rossphilipson:
Built here:
http://158.69.127.133:8020/builders/openxt-custom/builds/193

eric-ch:
LGTM.

PR #443, opened 2016-10-26
refpolicy: move changes for OpenXT-added policy files to those files

stephensmalley:
OpenXT-added policy files are managed directly rather than as patches.
Move the changes from xen4.6-uprev.patch for OpenXT-added policy files
to those files.

This should yield no change in the final policy.


stephensmalley:
This is admittedly a nit but I was going to make changes to dbusbouncer and noticed this discrepancy, so I wanted to fix it before making other changes.

stephensmalley:
Something wrong here, will resubmit later

PR #447, opened 2016-10-28
Fix the battery op region PRT4 to use correct port 0x84

rossphilipson:
I also reverted a couple of other miscellaneous changes that were really not
needed. The presumably came from upstream. It is better to keep ssdt_pm.asl the
way it has been tested and working for years. Since this stuff got completely
removed the latest Xen it doesn't really matter.

OXT-793

Signed-off-by: Ross Philipson philipsonr@ainfosec.com


eric-ch:
LGTM, will merge soon as things are broken without it.

PR #450, opened 2016-11-03
refpolicy: Split rpcproxy_t domain into per-instance domains.

stephensmalley:
OpenXT launches three instances of rpc-proxy, running with
different arguments that affect their behavior, listening on
different v4v ports, and accessible by different client VMs.
These different instances have different permission requirements.
Split the single rpcproxy_t domain into three separate domains,
one per running instance, and specialize policy for each instance.
At present, the domains are named to match the names used in the
init script, i.e. a base rpc_proxy_t domain, a rpc_proxy_guest_t
domain, and a rpc_proxy_websockets_t domain.  rpc_proxy_guest_t
is a bit of a misnomer now that guest VMs are no longer allowed to
connect to it, so we could perhaps rename to rpc_proxy_syncvm_t since
that appears to be its only allowed client.

Testing was performed to identify all required permissions for each
of the three domains, including use of auditallow statements to
collect audit messages whenever a given permission was used.  This was
also used to prune dead permissions that no longer seem to be in use.

Some improvements we gain from this split:
- Each rpc-proxy instance is now limited by SELinux to only connecting
to backend services that are legitimate for its clients, allowing the
rpc firewall rules to be reinforced by SELinux even in the event of a
flaw in rpc-proxy. rpc-proxy is still trusted to enforce the finer-grained
rpc-firewall rules, but cannot exceed the boundaries specified by SELinux.
We could improve further upon this if we were to more cleanly separate which
rpc-proxy instances were used by each client VM (some are shared by
more than one kind of client).

- Only the websockets rpc-proxy instance needs to be allowed to execute
external helper programs (openssl), thereby preventing an exploit of one
of the other instances from executing external binaries.

- Only the base rpc-proxy instance needs to be allowed to register a
service name on the system bus; the other instances do not currently
use this functionality.  This could be refined to only allowing registering
a specific name through the use of dbus_contexts configuration.

- Each instance is isolated from the others; a compromise of one
no longer allows access to any other.

OXT-799


richpersaud:
I would suggest duplicating this valuable PR comment on the JIRA ticket (with a link to the PR), especially if the JIRA ticket will be used for multiple PRs/improvements.

stephensmalley:
Ping?

jean-edouard:
Starting a custom build, will test tomorrow.

jean-edouard:
Builds fine, works fine, checked domains of files and processes, will merge soon, sorry for the delay.

PR #453, opened 2016-11-08
Linux change tarball source.

eric-ch:
Use cdn.kernel.org (https://www.kernel.org/introducing-fastly-cdn.html).


jean-edouard:
LGTM, builds fine, merging soon.

PR #454, opened 2016-11-10
OXT-814 : fix populate-volatile.sh link creation logic

dozylynx:


jean-edouard:
Wow, the way this script is written is... interesting!

Anyway, it took a while but I now understand your fix, code LGTM :)

However, I'd be interested to know how you discovered the issue.
This script should not even exist, since we shouldn't be using volatiles...

dozylynx:
Thanks, @jean-edouard . I found this working on local infrastructure, where the software expects a config file to be present in one of the dom0 filesystem locations where we currently have a tmpfs and it made most sense to protect the confidentiality of the file contents by storing it in /config. This script appears to have been designed to assist in just such a scenario by performing the symlink creation during boot. Works fine with this fix applied.

jean-edouard:
Thanks, merging soon.

PR #457, opened 2016-11-10
STABLE-6: Linux recipes cleanup and mirror change

eric-ch:
Use KERNELORG_MIRROR variable instead of static path.

Remove defunct and unmaintained linux versions.

Use MAJOR.MINOR linux-libc-headers tarball to make micro updates easier.

### DIFF
--- a/recipes-kernel/linux-libc-headers/linux-libc-headers_4.4.27.bb
+++ b/recipes-kernel/linux-libc-headers/linux-libc-headers_4.4.27.bb
@@ -2,7 +2,7 @@ require linux-libc-headers.inc
 
 PV_MAJOR = "${@"${PV}".split('.', 3)[0]}"
 
-SRC_URI = "https://www.kernel.org/pub/linux/kernel/v${PV_MAJOR}.x/linux-${PV}.tar.gz;name=kernel"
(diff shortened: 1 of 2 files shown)


rossphilipson:
I am not sure cleaning up unused files like this falls under what should go in a micro release that should be only security and bug fixes. I guess it doesn't hurt but seems out of scope. Other thoughts on this?

jean-edouard:
The mirror change is needed, but I agree that the major cleanup doesn't really belong here, unless it serves a higher purpose than just "being cleaner"...

eric-ch:
@rossphilipson @jean-edouard ack, I will submit another PR. These recipes for Linux are stale and have not been maintained since a long time (so are their patches). Keeping them in the layer could imply OpenXT still support them, or so I thought.

rossphilipson:
@eric-ch i really just brought it up for discussion - i could go either way - but we should probably do a write-up on exactly what can and can't go in a micro release. i would leave it open for now...

eric-ch:
@rossphilipson @jean-edouard The 2 commits with the cleanups are now gone.

rossphilipson:
Ok LGTM now, merging soon.

PR #458, opened 2016-11-10
Use KERNELORG_MIRROR source.

eric-ch:
This is what OE does, so where we fetch linux-libc-headers already.
It makes it easier to change the mirror from local.conf if necessary.


rossphilipson:
Built here:
No problems, merging soon.

PR #462, opened 2016-11-15
Move installer patch-queue to git repo

eric-ch:
No reason to keep patches against OpenXT owned repositories.

https://github.com/OpenXT/installer/pull/25


jean-edouard:
See https://github.com/OpenXT/installer/pull/25#issuecomment-260716132

rossphilipson:
OXT-724

PR #464, opened 2016-11-15
OXT-805: remove OpenXT zlib recipe in favour of using upstream version.

dozylynx:
Patches to zlib that were being applied are believed to no longer be required for OpenXT.

openembedded-core/meta/recipes-core/zlib in jethro is currently providing the same zlib version, 1.2.8, so the immediate effect of this change is to drop the patches that were being applied to provide the save_inflate_checkpoint and load_inflate_checkpoint functions.


jean-edouard:
LGTM.
Just for clarity, this PR is removing a bbappend, not a whole recipe.
The patches are useless indeed.

Just to be on the safe side, I'll include this in the next custom build before merging.

rossphilipson:
Built here:
No problems, merging soon.

PR #466, opened 2016-11-16
Remove idldir conf from surfman EXTRA_OECONF.

eric-ch:
Surfman does not use the idl stubs, silence a warning.


rossphilipson:
Built here:
No problems, merging soon.

PR #467, opened 2016-11-16
dom0-tweaks RDEPENDS on bash

eric-ch:
Scripts #!/bin/bash require bash installed.
Silence QA warning.


jean-edouard:
You're sure it's safe to assume that RDEPENDS is empty prior to the assignment? I would have used `+=`, just in case...

eric-ch:
@jean-edouard fixed.

rossphilipson:
Built here:

No problems, merging soon.

PR #470, opened 2016-11-17
Adding the new Measured Launch architecture

dpsmith:
This refactors the measured launch capability to enable a forward seal capability. This pull request is reliant on pull request #26 for the installer repository

OXT-536

### DIFF
--- a/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
+++ b/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
@@ -169,10 +169,11 @@ tpm_setup() {
     insmod /lib/modules/$(uname -r)/kernel/drivers/char/tpm/tpm.ko
     insmod /lib/modules/$(uname -r)/kernel/drivers/char/tpm/tpm_tis.ko
     mknod /dev/tpm0 c 10 224
-    echo -n "initramfs measuring $ROOT_DEVICE: "
+    echo -n "initramfs measuring $ROOT_DEVICE ...  "
     s=$(sha1sum $ROOT_DEVICE)
-    echo $s
+    echo "done"
     echo -n ${s:0:40} | TCSD_LOG_OFF=yes tpm_extendpcr_sa -p 15
+    [ $? -ne 0 ] && fatal "PCR-15 extend failed"
 }
 
 
@@ -199,10 +200,8 @@ mount_root
 
 maybe_break tpm
 
-[ -e /root/boot/system/tpm/enabled ] && {
-    echo "Setting up TPM..."
-    tpm_setup
-}
+echo "Setting up TPM..."
+tpm_setup
 
 maybe_break pivot
 boot_root
--- a/recipes-openxt/openxt/openxt-keymanagement/key-functions
+++ b/recipes-openxt/openxt/openxt-keymanagement/key-functions
@@ -1,0 +1,551 @@
+#!/bin/sh
+#
+# Copyright (c) 2016 Daniel P. Smith, Apertus Solutions, LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+#
+
+# LUKS Key Slots 
+RSLOT=0
+ESLOT=1
+PSLOT=2
+
+KEYMNT="${KEYMNT:-/media/ram}"
+KEYDIR="${KEYDIR:-${KEYMNT}/keys}"
+
+# Setup keystore
+setup_keystore() {
+    if ! grep -q ${KEYMNT} /proc/mounts; then
+        mount -t ramfs -o defaults,relatime,size=5M ramfs ${KEYMNT} 2>/dev/null
+        if [ $? -ne 0 ]; then return 1; fi
+    fi
+
+    # Need to absolutely make sure its a ramfs under us
+    if [ "$(grep ${KEYMNT} /proc/mounts|cut -f3 -d' ')" != "ramfs" ]; then
+        mount -t ramfs -o defaults,relatime,size=5M ramfs ${KEYMNT} 2>/dev/null
+        if [ $? -ne 0 ]; then return 1; fi
+    fi
+
+    if [ ! -d ${KEYDIR} ]; then
+        mkdir ${KEYDIR} 2>/dev/null
+        if [ $? -ne 0 ]; then return 1; fi
+    fi
+}
+
+keystore_ready() {
+    if grep -q ${KEYMNT} /proc/mounts && [ -d ${KEYDIR} ]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+close_keystore() {
+    if grep -q ${KEYMNT} /proc/mounts; then
+        umount ${KEYMNT}
+    fi
+}
+
+password_prompt() {
+    local request=$1
+    local key=$2
+
+    exec 3>/dev/tty
+
+    if ! keystore_ready; then
+        dialog --ok-label Continue --msgbox \
+            "ERROR: Keystore does not appear to be available!" 0 0 1>&3
+        return 1;
+    fi
+
+    while [ 0 ] ; do
+        local capture="${KEYDIR}/${key}"
+        dialog --cancel-label Previous --insecure --passwordbox \
+            "Please enter a new password for ${request}" 0 50 "" 2>${capture} 1>&3
+
+        local ret=$?
+        if [ ${ret} -ne 0 ]; then
+            rm -f ${capture}
+            exec 3>&-
+            return ${ret}
+        fi
+
+        local repeat=$(mktemp -p ${KEYDIR})
+        dialog --cancel-label Previous --insecure --passwordbox \
+            "Please enter the same password to confirm" 0 50 "" 2>${repeat} 1>&3
+
+        local ret=$?
+        if [ ${ret} -ne 0 ]; then
+            rm -f ${capture}
+            rm -f ${repeat}
+            exec 3>&-
+            return ${ret}
+        fi
+
+        local password=$(cat ${capture})
+
+        if [ -z "${password}" ]; then
+            rm -f ${capture}
+            rm -f ${repeat}
+            dialog --colors --ok-label Continue --msgbox \
+                '  \ZbERROR:\ZB Empty password!\n\n    Please try again.' 7 30 1>&3
+            continue
+        fi
+
+        if diff -q ${capture} ${repeat} 2>/dev/null ; then
+            rm -f ${repeat}
+            break
+        fi
+        rm -f ${repeat}
+
+        dialog --ok-label Continue --msgbox \
+            "ERROR: Passwords did not match. Please try again." 0 0 1>&3
+    done
+
+    exec 3>&-
+    echo ${capture}
+}
+
+recovery_prompt() {
+    local try=""
+
+    # Explicit interact on tty1
+    exec 3>/dev/tty1
+
+    if ! keystore_ready; then
+        dialog --ok-label Continue --msgbox \
+            "ERROR: Keystore does not appear to be available!" 0 0 1>&3
+        return 2;
+    fi
+
+    case ${1} in
+        1) try="first" ;;
+        2) try="second" ;;
+        3) try="third" ;;
+        *) try="more than you should have" ;;
+    esac
+
+    local capture="${KEYDIR}/recovery"
+    dialog --no-cancel --insecure --colors --passwordbox "
+    This is the ${try} attempt.
+
+    Please enter the recovery password " 0 0 2>${capture} 1>&3
+
+    local ret=$?
+    if [ ${ret} -ne 0 ]; then
+        rm -f ${capture}
+        exec 3>&-
+        return ${ret}
+    fi
+
+    exec 3>&-
+    echo ${capture}
+}
+
+# Seed kernel's CPRNG with some entropy
+# This is a cheap attempt to get the entropy pool to an acceptable
+#   level. If there is a concern then override this function to call
+#   out to an external entropy source.
+seed_entropy() {
+    local seeks="5 72 96 128 51 83 112 12 31 205"
+
+    if [ ! -e /dev/sda ]; then
+        return 1
+    fi
+
+    for seek in $seeks; do
+        local entropy=$(cat /proc/sys/kernel/random/entropy_avail)
+        if [[ $entropy -gt 128 ]]; then
+            return 0;
+        fi
+
+        dd if=/dev/sda of=/dev/null bs=1M count=1 seek=${seek} 2>/dev/null
+        sync
+        sleep 2
+    done
+
+    if [[ $entropy -gt 128 ]]; then
+        return 0;
+    else
+        return 1;
+    fi
+}
+
+# Generates the TPM own key/password
+gen_own_key() {
+    local key_file=""
+    local well_known="0000000000000000000000000000000000000000"
+
+    if ! keystore_ready; then
+        setup_keystore
+
+        # If function 'answerfile_specifies' exist, then inside the installer
+        type answerfile_specifies 2>/dev/null | head -1 | grep -q "function$"
+        if [[ $? -eq 0 ]]; then
+            if interactive; then
+                show_cursor
+                key_file=$(password_prompt "TPM Own" "own")
+                hide_cursor
+            else
+                key_file=${key_file:-"${KEYDIR}/own"}
+                echo -n ${well_known} > ${key_file}
+            fi
+        else
+            key_file=$(password_prompt "TPM Own" "own")
+        fi
+    fi
+
+    key_file=${key_file:-"${KEYDIR}/own"}
+
+    if [ ! -e ${key_file} ]; then
+        echo -n ${well_known} > ${key_file}
+    fi
+
+    echo ${key_file}
+}
+
+# Generates the recovery key/password
+gen_recovery_key() {
+    local key_file=""
+    local well_known="0000000000000000000000000000000000000000"
+
+    if ! keystore_ready; then
+        setup_keystore
+
+        # If function 'answerfile_specifies' exist, then inside the installer
+        type answerfile_specifies 2>/dev/null | head -1 | grep -q "function$"
+        if [[ $? -eq 0 ]]; then
+            if interactive; then
+                show_cursor
+                key_file=$(password_prompt "device recovery" "recovery")
+                hide_cursor
+            else
+                key_file=${key_file:-"${KEYDIR}/recovery"}
+                echo -n ${well_known} > ${key_file}
+            fi
+        else
+            key_file=$(password_prompt "device recovery" "recovery")
+        fi
+    fi
+
+    key_file=${key_file:-"${KEYDIR}/recovery"}
+
+    if [ ! -e ${key_file} ]; then
+        echo -n ${well_known} > ${key_file}
+    fi
+
+    [ "$(getenforce)" == "Enforcing" ] &&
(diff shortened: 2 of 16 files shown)


rossphilipson:
Built here:

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
```diff
@@ -0,0 +1,375 @@
...
+
+hex2bin() {
+    read hex
+    echo -n $hex | sed 's/\([0-9A-F]\{2\}\)/\\\\\\x\1/gI' | xargs printf
+}
+
+# Function to parse grub.cfg and return all multiboot modules and their
+#   full command line
+# parameter 0: path to grub.conf
+parse_grub() {
```
jean-edouard: Aren't we trying to do tboot's work here? Can't we just do a `txt-stat`? Parsing the grub config file really doesn't seem ideal...

recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
```diff
@@ -169,10 +169,10 @@ tpm_setup() {
     insmod /lib/modules/$(uname -r)/kernel/drivers/char/tpm/tpm.ko
     insmod /lib/modules/$(uname -r)/kernel/drivers/char/tpm/tpm_tis.ko
     mknod /dev/tpm0 c 10 224
-    echo -n "initramfs measuring $ROOT_DEVICE: "
```
dpsmith: Never understood the case and really don't like that it was being done. If that is in fact the reason, it should be dropped all together. do you disagree?

recipes-openxt/openxt/openxt-keymanagement/key-functions
```diff
@@ -0,0 +1,551 @@
...
+        mkdir ${KEYDIR} 2>/dev/null
+        if [ $? -ne 0 ]; then return 1; fi
+    fi
+
+    return 0
+}
+
+keystore_ready() {
+    if grep -q ${KEYMNT} /proc/mounts && [ -d ${KEYDIR} ]; then
+        return 0;
```
dpsmith: likely an artifact from one-liner being expanded into a block, will clean up

recipes-openxt/openxt/openxt-keymanagement/key-functions
```diff
@@ -0,0 +1,551 @@
...
+}
+
+# Seed kernel's CPRNG with some entropy
+# This is a cheap attempt to get the entropy pool to an acceptable
+#   level. If there is a concern then override this function to call
+#   out to an external entropy source.
+seed_entropy() {
+    local seeks="5 72 96 128 51 83 112 12 31 205"
+
+    if [[ ! -e /dev/sda ]]; then
```
dpsmith: which would you prefer, the more portable [ or the least surprising/more forgiving [[?

recipes-openxt/openxt/openxt-keymanagement/key-functions
```diff
@@ -0,0 +1,551 @@
...
+
+    for seek in $seeks; do
+        local entropy=$(cat /proc/sys/kernel/random/entropy_avail)
+        if [[ $entropy -gt 128 ]]; then
+            return 0;
+        fi
+
+        dd if=/dev/sda of=/dev/null bs=1M count=1 seek=${seek} 2>/dev/null
+        sync
+        sleep 2
```
dpsmith: nope, we are sleeping to generate entropy, again as the said in the comment very poorly. sleeps cause interrupts to be fired which is a kernel entropy source but at the same time other sources could be depleting it which can have a negative effect.

recipes-openxt/openxt/openxt-keymanagement/key-functions
```diff
@@ -0,0 +1,551 @@
...
+        fi
+    fi
+
+    key_file=${key_file:-"${KEYDIR}/recovery"}
+
+    if [ ! -e ${key_file} ]; then
+        echo -n ${well_known} > ${key_file}
+    fi
+
+    chcon -t lvm_tmp_t ${key_file}
```
dpsmith: for chcon, because the previous is for the tpm owner key, not used by cryptsetup for lvm luks key. I would really like to get away from it because it's not a good practice, open to suggestions. for refactoring, this was already an abstraction from what was done in the existing approach to produce a clean key management api.

recipes-openxt/openxt/openxt-keymanagement/key-functions
```diff
@@ -0,0 +1,551 @@
...
+    echo $key_file
+}
+
+# Generates the log partition key
+gen_log_key() {
+    local root="${1:-/}"
+    local key_file="${root}/config/keys/log.key"
+
+    seed_entropy 2>/dev/null
+    dd if=/dev/urandom bs=1 count=128 of="${key_file}" 2>/dev/null || return 1
```
dpsmith: https://gist.github.com/tarcieri/6347417

recipes-openxt/openxt/openxt-keymanagement/key-functions
```diff
@@ -0,0 +1,551 @@
...
+    if [ -e ${key_file} ]; then
+        echo $key_file
+        return 0
+    else
+        echo ""
+        return 1
+    fi
+}
+
+# Gets the device binding key
```
dpsmith: why?

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
dpsmith: You can't use txt-stat as it will only tell you what was used for the current boot, it will not tell you what the grub file looks like from the upgrade. Relying on them to be the same will break if upgrade changes, e.g. like when intel-microcode was added to grub config

recipes-openxt/openxt/openxt-keymanagement/key-functions
dpsmith: Something being considered in the RSA implementation is to turn on tpm-rng kernel module and defer to rng-tool to seed the kernel's CPRNG. Perhaps this can be added in this go instead.

recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
jean-edouard: Running sha1sum on the whole root partition can be a lengthy process, especially on spinning hard drives. We should let the user/dev know what's happening.
If you don't like the `-n`, maybe just do an `echo Measuring rootfs...` before, and `echo Done. Hash = $s` after or something.

recipes-openxt/openxt/openxt-keymanagement/key-functions
jean-edouard: I don't have an opinion, I just think code should be consistent, especially in new files.
`[[` might be a tiny bit faster, since it's a builtin and not an external binary IIRC.

recipes-openxt/openxt/openxt-keymanagement/key-functions
jean-edouard: AFAIK, by itself, sleeping doesn't generate entropy. It just waits and hopes interrupts and stuff will happen during the sleep. Since 2 seconds * x can add up to a pretty long time, I'm just suggesting we tell the user to generate interrupts during the sleeps by using the keyboard/mouse.
Ultimately, I agree, we should have a proper mechanism like the one you're suggesting, but that can be a separate PR.

recipes-openxt/openxt/openxt-keymanagement/key-functions
jean-edouard: To reduce code duplication.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
jean-edouard: Is there more information on that somewhere? I don't understand the implications of grub changes re: upgrades.

recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
dpsmith: i am good with letting them know, just prefer we didn't send the hash. This is how i queued up the next iteration.

jean-edouard: Oh ok, yeah dropping the hash is fine.

recipes-openxt/openxt/openxt-keymanagement/key-functions
dpsmith: I dropped back to [ for consistency but if someone comes along with a reason to switch, can do so

jean-edouard: :+1:

recipes-openxt/openxt/openxt-keymanagement/key-functions
jean-edouard: Don't do it if you think it doesn't make sense. I don't know much about entropy generation, I just don't like sleeps :)

recipes-openxt/openxt/openxt-keymanagement/key-functions
dpsmith: A simple if logic that is 7 lines in total is not much duplication, it would be one thing if the refactor increased readability/logic reasoning but this refactor would result in nothing more than a does_file_exist function. Now if the logic was more complex like make sure the file existed, was readable and hand a fingerprint of xyz, then IMHO that would be worthy of a refactor.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
dpsmith: See [Phil's writeup](http://twobit.us/2013/06/calculating-pcr18-and-pcr19/) on PCRs 18 and 19 to understand what goes into the measurement. Other items about txt-stat is it does not tell you the filepath for the module it measured and it maps /dev/mem which is a poor security practice. The addition of a module is not the only change that would affect the measurement. Others include a module's commandline args changing or the order they are listed could change. The measurements are sensitive to all of these changes, any of which could happen during an upgrade. The whole point of this work is to predict the measurements of what will be not what currently is, which is what txt-stat is giving you.

dpsmith: For the record, this was least favored and last path chosen. Preferablly would like a solution such as [grubby](https://linux.die.net/man/8/grubby) that is able to properly parse grub config and print out in a parsable format the default boot entry. If you are aware of an existing tool or willing to write one, I would glad use it over a psuedo-parser written in bash.

recipes-openxt/openxt/openxt-keymanagement/key-functions
dpsmith: I don't like any of it at all. It's all a cheap way to do it and half tempted to write a full RFC that advocates tpm-rng and when TPM is not available, then havaged.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
jean-edouard: Grubby sounds good, anything that can avoid us maintaining grub parsing code would really be preferable...

dpsmith: Grubby doesn't actual print the full multiboot entry for the default boot (which is what we need), it will tell you the kernel path, entry number, and the title of the default entry.

rossphilipson:
The C code in 0001-Add-a-PCR-calculator-to-predict-measurements.patch looks pretty much the same as the result of the last review I did.

recipes-openxt/xenclient/xenclient-root-ro/init.root-ro
```diff
@@ -122,48 +103,196 @@ is_launch_measured()
...
+    local sealerr=$?
+
+    if [ $sealerr -eq 0 ]; then
+        dialog --colors --backtitle "${BACK_TITLE}" --mixedgauge \
+            "  Sealing: Done" 0 0 100
+
+        # store PCR hints for recovery
+        #  do PCR quote
+        if [ "${PCR_DEV}" ]; then
+        [ -f "${PCRS_BAD}" ] && rm "${PCRS_BAD}"
```
jean-edouard: Missing indent

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
jean-edouard: I still don't understand the link between upgrades and having to parse the grub config file.
Please give me more information, ideally not just a link to someone's writeup...

recipes-openxt/openxt/openxt-keymanagement/key-functions
dpsmith: but not during upgrade

jean-edouard: Do we have a way to test which case we're in? Could we at the very least redirect the error to /dev/null?

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
```diff
@@ -0,0 +1,375 @@
...
+    modules=$(parse_grub ${root}/$grub_conf)
+
+    if [[ $? -eq 0 ]]; then
+        local IFS=$'\n'
+        mod_num=0
+        for  modline in $modules; do
+            args=""
+            module=$(echo ${modline} | awk '{ print $1 }')
+            if [[ $(echo ${modline} | wc -w) -gt 2 ]]; then
+                cmdline=$(echo ${modline} | awk '{ print substr($0, index($0,$3)) }')
```
dpsmith: In the current utilization, correct, but if it was possible to read the acm heap, then pcr17 can be calculated instead of re-used.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
```diff
@@ -0,0 +1,375 @@
...
+# Function to calculate PCR 18
+# parameter 0: path to grub.cfg
+# return 0 on success
+# return 1 on failure
+pcr18() {
+    local root="${1:-/}"
+    local grub_conf=${2}
+    local pcr18=''
+
+    modules=$(parse_grub ${root}/$grub_conf)
```
dpsmith: yep, will refactor.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
```diff
@@ -0,0 +1,375 @@
...
+    local tboot_idx='0x20000001'
+
+    [ -f $policy ] && rm -f $policy
+
+    tb_polgen --create --type nonfatal --ctrl 1 $policy 2>&1 1>/dev/null && \
+    tb_polgen --add --num 0 --pcr none --hash any $policy 2>&1 1>/dev/null && \
+    tb_polgen --add --num any --pcr 19 --hash any $policy 2>&1 1>/dev/null
+    if [ $? -ne 0 ]; then
+        echo "failed to generate tboot Verified Launch policy" >&2
+        return 1
```
dpsmith: Its only a temp file with no sensitive information. The reason it removed at beginning is that tb_polgen --create fails if the file exist, IIRC.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
```diff
@@ -0,0 +1,375 @@
...
+    fi
+
+    local polsize=$(cat $policy | wc -c)
+    tpmnv_defindex -i $tboot_idx -s $polsize -pv 0x02 -p "$password" 2>&1 1>/dev/null && \
+    lcp_writepol -i $tboot_idx -f $policy -p "$password" 2>&1 1>/dev/null
+    if [ $? -ne 0 ]; then
+        echo "failed to write tboot Verified Launch policy to TPM" >&2
+        return 2
```
dpsmith: Same as above.

recipes-openxt/xenclient/xenclient-config-access/config-access.initscript
```diff
@@ -32,8 +32,7 @@ sig=$(dd if="/dev/mapper/${CONFIG_LV}" bs=4 count=1 2>/dev/null)
 case "$sig" in
 LUKS)
 	keycmd="get-config-key"
-	slot=0
-	eval $keycmd | cryptsetup -q -d - -S $slot luksOpen \
+	eval $keycmd | cryptsetup -q -d - luksOpen \
```
dpsmith: If no slot is specified, it will try every defined slot. If you specify a slot, it will only attempt to use key to unlock the specified slot or fail. Taking this approach so that alternate key schemes can place the keys in the slots they prefer and this script will not have to be changed.

recipes-openxt/openxt/openxt-keymanagement/key-functions
dpsmith: yes and yes.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
jean-edouard: Then maybe comment out the code and write that as an explanation?

recipes-openxt/xenclient/xenclient-config-access/config-access.initscript
jean-edouard: That makes sense, thank you.

recipes-openxt/openxt/openxt-keymanagement/key-functions
```diff
@@ -0,0 +1,549 @@
...
+
+# Generates the TPM own key/password
+gen_own_key() {
+    local key_file=""
+    local well_known="0000000000000000000000000000000000000000"
+
+    if ! keystore_ready; then
+        setup_keystore
+
+        # If function 'answerfile_specifies' exist, then inside the installer
```
jean-edouard: Just ran into this, seems better:
https://github.com/OpenXT/xenclient-oe/blob/master/recipes-core/images/xenclient-installer-image.bb#L90

dpsmith: Sorry but not understanding what you are comparing?

dpsmith: Okay, definitely. Though I believe I got this function elsewhere so perhaps we should clean it up everywhere.

dpsmith:
Resolved the installer issue. Please try again.

rossphilipson:
Latest built here: http://158.69.127.133:8020/builders/openxt-custom/builds/234

rossphilipson:
Latest build rebased on XL merge:

http://158.69.127.133:8020/builders/openxt-custom/builds/235

rossphilipson:
Testing done on the latest using:
http://openxt.ainfosec.com/builds/master/custom-dev-237-master/
Looks good. Tested:
Fresh install - PXE
Breaking measurements and resealing.
Upgrading from 6.0.0 to 7.0.0 w/ ML
Upgrading from build with ML pretending to be 6.0.0 to 7.0.0 w/ ML
All forward sealing operations seem to be working.

Merging later today if there are no objections...

rossphilipson:
Merging soon

recipes-openxt/openxt/openxt-keymanagement/key-functions
```diff
@@ -0,0 +1,551 @@
...
+        fi
+    fi
+
+    key_file=${key_file:-"${KEYDIR}/recovery"}
+
+    if [ ! -e ${key_file} ]; then
+        echo -n ${well_known} > ${key_file}
+    fi
+
+    [ "$(getenforce)" == "Enforcing" ] &&
```
jean-edouard: @dpsmith, this line breaks on the installer. (`[ == Enforcing]`).

dpsmith: Is this really "breaking" the installer or just creating an error that needs to be cleaned up?

jean-edouard: Not sure

dpsmith: Jason, says there are a handful of bashism that he saw when redoing the password capture. Probably should create a ticket to have all these cleaned up.

eric-ch: The output of ```getenforce``` will be empty and fail the ```==``` check. But the installer-part 1 does not have the ```getenforce``` binary and should check for its presence before (e.g. https://github.com/eric-ch/xenclient-oe/commit/98594359ba2f9952d10b1dc7241d6ab5553bad1b)

dpsmith: Yes, that will deal with it. Through discussion here, it probably would be better to have the policycoreutils-setfiles package added to xenclient-installer-image recipe and then replace drop the getenforce check and just call setfiles.

eric-ch: @dpsmith That works. Can you put a ticket in for that?

PR #471, opened 2016-11-17
Add qmp-helper recipe

jean-edouard:


rossphilipson:
Built here w/o problems:

Merging soon.

PR #472, opened 2016-11-18
Disable TMEM stats in xenstat tools.

rossphilipson:
We block TMEM with XSM so the tool should not try to get TMEM information
from the hypervisor.

OXT-798


eric-ch:
+1, LGTM.

PR #478, opened 2016-11-23
Remove icbinn-fuse, fuse and kernel fuse support in dom0.

rossphilipson:
OXT-669


rossphilipson:
Nothing is using fuse so the best thing to do is remove it. It built here yesterday:
http://158.69.127.133:8020/builders/openxt-custom/builds/218

rossphilipson:
@stephensmalley there are bits in the SELinux policy that reference fuse (see ticket). Is it simply a matter of removing them since fuse is being removed?

stephensmalley:
Doesn't look like those are adding or removing any policy related to fuse; those lines are just context lines in the diffs.  And there isn't any straightforward way to exclude all fuse-related policy, so I don't think there is anything to do here policy-wise.

rossphilipson:
Ack, thanks.

jean-edouard:
Tested, LGTM.
@rossphilipson Did you make sure icbinn doesn't use fuse?

rossphilipson:
As far as I can tell, only the icbinn-fuse sub-project uses it and it is not used in openxt in any way I can find. Looking at history, it sees icbinn-fuse was some WIP that never got finished.

jean-edouard:
Cool, merging soon then.

PR #479, opened 2016-11-23
STABLE-6: XSAs, november 2016

eric-ch:
XSA-194: Xen 4.4 is not affected.
XSA-196: Xen 4.4 is not affected.
XSA-198: OpenXT does not use pygrub.

OXT-843


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/221
LGTM, seems to work fine, merging soon.

PR #481, opened 2016-11-30
[xen] Comment HVM RTC patch as best as possible.

rossphilipson:
OXT-585


eric-ch:
+1, merging.

PR #482, opened 2016-12-05
Update the snapshot and scan functions to use the vhd realpath

rneilturner:
The snapshot and scan functions needed to use vhd_realpath to use icbinn when necessary. Without this commit the synchronizer will never create the snapshot disk after downloading the base vhd.

See OXT-785

To test this, run vhd-util for a dom0 snapshot from the synchronizer with the correct LIBVHD_ICBINN_VHD_SERVER and LIBVHD_ICBINN_KEY_SERVER set. Before this change, you will receive an error code ENOENT. After this change the vhd-util snapshot function should recognize the existence of the icbinn server set and execute the function using icbinn.

### DIFF
--- a/recipes-extended/xen/files/blktap2-vhd-icbinn-support.patch
+++ b/recipes-extended/xen/files/blktap2-vhd-icbinn-support.patch
@@ -1091,3 +1091,29 @@ Index: xen-4.6.1/tools/blktap2/drivers/Makefile
  
  install: all
  	$(INSTALL_DIR) -p $(DESTDIR)$(INST_DIR)
+Index: xen-4.6.1/tools/blktap2/vhd/lib/vhd-util-snapshot.c
+===================================================================
+--- xen-4.6.1.orig/tools/blktap2/vhd/lib/vhd-util-snapshot.c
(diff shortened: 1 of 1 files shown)


rossphilipson:
That makes perfect sense since vhd_realpath was introduced by icbinn to differentiate between local and icbinn paths. Looking back at the bkltap.git repo, it looks like that was the original code. I must have missed that when i extricated and rebased the patches on in-tree blktap2. We don't really have the means to test syncvm bits at this point. I am inclined to take this commit as it stands unless there are objections.

richpersaud:
Thanks @rneilturner for identifying and fixing this regression.

If this affects stable-6, we can add a 6.0.1 release note based on the second sentence of the PR and details in the JIRA ticket.

Hopefully there are no other patches that we need from the blktap.git repo.  It would be good to have a few BATS CLI regression tests that could be run locally within a SyncXT client VM, without needing to stand up a central management server.

rossphilipson:
This does not affect stable 6.0 since we are still using the old blktap repo there.

Extracting our "useful" code from that old blktap repo was a nightmare. Luckily there have been few regressions.

rossphilipson:
So there doesn't seem to be any disagreement on merging it. Will merge soon...

PR #483, opened 2016-12-05
Turn off switcher debug logging.

eric-ch:
"XenInput:direct event handler"
"XenInput:send led keycode 0x0"
... are very noisy otherwise.

OXT-857


jean-edouard:
LGTM, merging tomorrow.

PR #485, opened 2016-12-06
[OXT-580] Updated acpid to use bbappend to pull from openembedded as …

cdkell3:
[OXT-580] Changed acpid to use bbappend files instead of bb. This allows us to get updates from upstream as requested.


jean-edouard:

LGTM, merging soon.

PR #486, opened 2016-12-06
STABLE-6: OXT-859: XSA-199

eric-ch:
https://xenbits.xen.org/xsa/advisory-199.html

OXT-859


jean-edouard:
Built here: http://158.69.127.133:8020/builders/openxt-custom/builds/224
LGTM, merging soon.

PR #487, opened 2016-12-06
OXT-860: Upgrade to 4.4.36

eric-ch:
xenstore-from-pv-guest.patch removed: Upstream commit: 9a035a4

OXT-860


jean-edouard:

Works fine and code LGTM, merging soon.

PR #492, opened 2016-12-14
Fix the readonly-ide patch to write out the config option to the make file

rneilturner:
Without this fix you can't have readonly ide disks with the QEmu 2.6.2 uprev


rossphilipson:
Guess I missed that bit during the port. LGTM, merging soon.

PR #494, opened 2016-12-19
Replace midori in favor of a minimalist browser

rneilturner:
See OXT-875


jean-edouard:
Looks like Midori scripts were in charge of doing stuff related to Japanese keyboards (using uim), that's gone in the PR and could break stuff for potential Japanese users.
Also the scripts you removed close the nm-applet menu on browser start and stop, to maintain the illusion that the menu is part of the webpage, that seems to be gone too.

richpersaud:
Unless we have Japanese testers, or an automated Japanese testing framework, or one developer willing to maintain Japanese localization, I suggest that we create a ticket to track the removed keyboard support, and move on.

rneilturner:
Sorry for the late response, been very busy. The script midori report was removed because I couldn't replicate the issue it was supposedly resolving. I can add it back in easy enough if thats what everyone wants. Just let me know either way. I can't see any reason to add back in the midori login script as I believe this is no longer applicable.

rossphilipson:
This was built and minimally tested. I think the best way to get more test coverage on this is to just merge it and have people start using it. That would be more likely to ferret out hard to find issues. Thoughts?

richpersaud:
Agreed on merge to master.

rossphilipson:
I will merge it tomorrow morning if there are no objections.

PR #499, opened 2017-01-03
refpolicy: adjust for qtdbd

dpsmith:
Adjust the policy for accesses qtdbd require.

OXT-787

### DIFF
--- a/recipes-security/refpolicy/refpolicy-mcs-2.20141203/patches/policy.booleans.diff
+++ b/recipes-security/refpolicy/refpolicy-mcs-2.20141203/patches/policy.booleans.diff
@@ -1,0 +1,13 @@
+Index: refpolicy/policy/booleans.conf
+===================================================================
+--- refpolicy.orig/policy/booleans.conf
++++ refpolicy/policy/booleans.conf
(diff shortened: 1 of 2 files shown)


stephensmalley:
Looks good as far as content.  I did notice however that it was diff -Naur generated rather than quilt refresh generated so the diff style is inconsistent; might want to fix that. Note that the labeling of /sys/devices/system/cpu/online is now upstream in refpolicy too so eventually that will drop out upon a rebase.

rossphilipson:
This was discussed and accepted as reasonable. We can merge it once the patch is reformatted.

dpsmith:
Would it be better to break the labeling of /sys/devices/system/cpu/online into a separate patch so it can just be dropped when rebase occurs?

dpsmith:
Patches have been refreshed using quilt.

rossphilipson:
Ack, merging soon.

rossphilipson:
Built here:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/554
Testing...

rossphilipson:
Looks good, no issues. Merging soon.

PR #501, opened 2017-01-04
STABLE-6: qemu: xenmou: accept all scrolls, not just 1 and -1

jean-edouard:


rossphilipson:
What ticket does this go with? I assume we want it in 6.0.1 since we are trying to fix the scroll issues there.

jean-edouard:
No ticket

PR #502, opened 2017-01-05
OXT-884: fixed correct dom0_max_vcpus param for xen 4.3+

gamorg02:
For xen 4.3.4+ the param changed from max_dom0_vcpus to dom0_max_vcpus so the installer correctly only loads one vcpu for installation.

[Xen Performance Tuning Reference](https://wiki.xen.org/wiki/Tuning_Xen_for_Performance)

There may be additional potential changes for the grub.cfg as well to increase stability


richpersaud:
Without this fix, does dom0 get all VCPUs, then lose them as VMs are started?  Or do the VCPUs continue to be shared between dom0 and VMs?  If the latter, then we should include this fix in 6.0.1, as it would reduce scheduling of VCPUs between guests and VMs.

jean-edouard:
@richpersaud I think the fix is already in dom0, this PR is just to fix the installer

richpersaud:
How was this problem found, e.g were there visible negative side-effects on the installer?

rossphilipson:
Actually we don't use that parameter in the grub config at all. We just use the default behavior. I think it has always been that way. I am not sure why we explicitly set it in the installer configs or how it was found.

I do think we should fix it for 6.0.1 since it is simply wrong.

gamorg02:
@richpersaud , this was found as the parameter name changed since 4.1.3 and was never updated for the installer only.  It should improve installations on systems with many cores (32+) so not all CPUs need to be in dom0 to start installation.  For the general 2-8 vcpu system there should be no noticeable impact.

@jean-edouard, the runtime dom0 does _not_ have 1 vcpu as the parameter is not set:

https://github.com/OpenXT/xenclient-oe/blob/63b35679f5a5d712b88a804fdc8bae720c6e2fa7/recipes-openxt/xenclient/xenclient-dom0-tweaks/grub.cfg

And running cat /proc/cpuinfo in dom0 shows as many vcpus there are on the system until that is set in grub.cfg (at least on stable6-184).  

I am still investigating the impact of changing the runtime configuration.

rossphilipson:
Any objection to pulling this fix into 6.0.1 before doing an RC1?

gamorg02:
One other thought could be to remove the parameter completely, which would clean up the boot params and keep behavior the same.

rossphilipson:
That is another possibility. Perhaps the best thing to do is:

Master/7.0.0: Take the change in your PR since you have identified a situation where having more than one VCPU in dom0 causes a performance problem. The installer does not need more than one VCPU.

6.0.1: Submit a PR to remove the incorrect parameter but not cause a functional change since it is a micro release.

gamorg02:
+1 to Ross's idea.

rossphilipson:
I am not sure max_dom0_vcpus was ever right. We may have just always had a bum parameter.

gamorg02:
That syntax works on Xen 4.1.3 oddly enough.

rossphilipson:
Since no one objected, I am going to submit a stable-6 PR that just removes those params.

rossphilipson:
Done: https://github.com/OpenXT/xenclient-oe/pull/503

rossphilipson:
No objections to taking this version in master?

eric-ch:
+1.

rossphilipson:
This one can go in. Merging later this morning.

PR #506, opened 2017-01-11
Add the KabyLake SINIT acm

rneilturner:

See OXT-887


eric-ch:
LGTM, +1.

rossphilipson:
@rneilturner could you post a version of this PR for stable-6?

rossphilipson:
LGTM merging soon. We don't have any systems that use KBL SINIT so we are not able to test directly.

PR #509, opened 2017-01-19
OXT-897: Add recent i915 arch firmware.

eric-ch:
https://01.org/linuxgraphics/intel-linux-graphics-firmwares

New generations of Intel Graphics for Linux requires Firmware components
to be integrated in the kernel-mode driver.

It is likely that an updated version of the meta OE recipe will package these at some point.

OXT-897


richpersaud:
Can we contribute this to upstream OE?

eric-ch:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/542

@richpersaud If this is not already done in versions that use a recent enough kernel.

eric-ch:
The recipe we have does not fetch a recent enough version with 4.9. Will re-open once fixed.

PR #511, opened 2017-01-20
OE config refactor

dpsmith:
This refactors the OpenXT OE configuration to better align with OE practices. This PR is inter-dependent on PR [#218](https://github.com/OpenXT/openxt/pull/218) from openxt.gt repository.

### DIFF
--- a/conf/distro/include/openxt-xen.inc
+++ b/conf/distro/include/openxt-xen.inc
@@ -1,0 +1,5 @@
+# xen version and source archive
+XEN_VERSION="4.9.0"
+XEN_SRC_URI="https://downloads.xenproject.org/release/xen/4.9.0/xen-4.9.0.tar.gz"
+XEN_SRC_MD5SUM="f0a753637630f982dfbdb64121fd71e1"
+XEN_SRC_SHA256SUM="cade643fe3310d4d6f97d0c215c6fa323bc1130d7e64d7e2043ffaa73a96f33b"
--- a/conf/distro/openxt-main.conf
+++ b/conf/distro/openxt-main.conf
@@ -16,51 +16,66 @@
 # require conf/sanity.conf
 #
 
-INHERIT += "openxt-images"
+DISTRO_NAME = "OpenXT"
+DISTRO_VERSION = "${@time.strftime('%Y.%m',time.gmtime())}"
 
-PREFERRED_PROVIDER_virtual/kernel = "linux-openxt"
-PREFERRED_VERSION_linux-openxt ?= "4.9%"
+BUILDNAME = "OpenXT ${DISTRO_VERSION}"
+
+XENCLIENT_BUILD ?= "${@time.strftime('%y%m%d',time.gmtime())}"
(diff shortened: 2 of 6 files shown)


conf/distro/openxt-main.conf
```diff
@@ -16,44 +16,65 @@
...
+# make hacks
+PARALLEL_MAKE_pn-gnumail = "-j 1"
+PARALLEL_MAKE_pn-gnumail-native = "-j 1"
+PARALLEL_MAKE_pn-shared-mime-info = "-j 1"
+PARALLEL_MAKE_pn-shared-mime-info-native = "-j 1"
+
+LICENSE_FLAGS_WHITELIST = "commercial_broadcom-wifi"
+LICENSE_FLAGS_WHITELIST += "commercial_microcode-ctl"
+
+XENCLIENT_PACKAGE_FEED_URI="file:///storage/ipk"
```
jean-edouard: Is there a way to override this value from the build-scripts in openxt.git?
The builder needs to be able to set itself as the feed URI.

dpsmith: Sure, it can be added to BB_ENV_EXTRAWHITE, and change to ?= in .conf file. Then build can override by setting it in the environment.

jean-edouard: Sounds good, I just don't know much about this extrawhite thing, I'll probably need a bit of help to fix the builder.

dpsmith: Sure, but could you clarify by what you are referring to when you say "builder". To clarify BB_ENV_EXTRAWHITE, bitbake erases all environment varialbes to ensure no taiting from host machine except for any variable names listed in BB_ENV_WHITE and BB_ENV_EXTRAWHITE. Convention is that in the envoking environment defines any variables it wants bitbake to inherit in the BB_ENV_EXTRAWHITE environment variable. See the shell script build/bb in openxt.git, which is where I would add XENCLIENT_PACKAGE_FEED_URI. Then the "builder" exports what it wants to have set in XENCLIENT_PACKAGE_FEED_URI and it will override what is set in openxt-main.conf

jean-edouard: I just about finished setting up a new build server for openxt.ainfosec.com.
I used setup.sh from master, so it's a pretty standard setup.
I just have "small" script for buildbot to kick the build, you can see it there:
https://github.com/OXTbuilders/builder2017/blob/master/home/shared/build.sh
At line 143 I set XENCLIENT_PACKAGE_FEED_URI to the correct value. That's what I need to still be able to do after this PR.

conf/distro/include/openxt-xen.inc
```diff
@@ -0,0 +1,5 @@
+# xen version and source archive
+XEN_VERSION="4.6.1"
```
eric-ch: That should be upgraded to 4.6.4, to match the upgraded version in master before it can be merged.

dpsmith:
Updated to 4.6.4.

dpsmith:
Reworked version of these changes will be coming in a new PR.

PR #518, opened 2017-01-26
STABLE-6: OXT-899: Silence TPM warnings.

eric-ch:
use modprobe instead of insmods, TPM modules should be reported in
modules.dep, and depmod should have been run at initramfs image
creation.

The tpm_tis module will create the dev node.

tpm_bios no longer exist in recent kernels.

OXT-899


MagOXT:
OK in /custom/stable-6/551/openxt-dev-551-stable-6/

rossphilipson:
Thanks Magali, I will merge these soon.

PR #519, opened 2017-01-26
ocaml: backport another segfault fix in the garbage collector

jean-edouard:
OXT-896


jean-edouard:
The commit doesn't really fix a segfault, the ticket number in the description has a typo (it doesn't address PR#7378 but PR#7368, which is a minor issue).
Closing.

PR #522, opened 2017-01-30
tboot: fix merge collision.

eric-ch:
The recent changes in Tboot did not integrate as expected, the new Measured Launch architecture & OXT-462 (Tboot uprev) collided leaving things behind.

This should add and build the pcr-calc tool that went missing.


eric-ch:
OXT-462, that pretty much falls under that umbrella since the patch was against 1.8.3 and was left behind by the upgrade.

eric-ch:
Built here: http://openxt.ainfosec.com/builds/custom/master/559/

rossphilipson:
There is still a problem with ML and OTA upgrades but I don't think it is related to this. Adding the patch back correctly resulted in the missing binaries being where they needed to be. My testing of this shows that. Merging soon.

PR #524, opened 2017-01-31
libxl: install and run xen-init-dom0 at boot time

jean-edouard:
OXT-916

### DIFF
--- a/recipes-openxt/xenclient/xenclient-dom0-tweaks/xenstore-init
+++ b/recipes-openxt/xenclient/xenclient-dom0-tweaks/xenstore-init
@@ -2,4 +2,3 @@
 
 # add xenstore entries here
 
-xenstore-write domid 0
--- a/recipes-extended/xen/files/xen-init-dom0.initscript
+++ b/recipes-extended/xen/files/xen-init-dom0.initscript
@@ -1,0 +1,46 @@
+#! /bin/sh
+#
+# Copyright (c) 2017 Assured Information Security
+# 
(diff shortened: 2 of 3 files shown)


crogers1:
+1.  What's libnl for?

jean-edouard:
Dunno but OE says we use it, and complains if we don't have it in DEPENDS :)

crogers1:
Ah, I see.  Also, we might need selinux rules for the init script and xen-init-dom0 binary itself.

richpersaud:
libnl dependency may be related to Remus and COLO 

https://wiki.xen.org/wiki/COLO_-_Coarse_Grain_Lock_Stepping

https://wiki.xenproject.org/wiki/Xen_Project_Release_Features

Would be nice if we could disable those features in LibXL and eliminate the libnl dependency, but maybe that requires Kconfig and Xen 4.7.

rossphilipson:
Build here (testing now):
http://openxt.ainfosec.com/builds/custom/master/584

recipes-openxt/xenclient/xenclient-dom0-tweaks/xenstore-init
```diff
@@ -2,4 +2,3 @@
 
 # add xenstore entries here
 
-xenstore-write domid 0
```
rossphilipson: Why is this removed? Does xen-init-dom0 do that now?

jean-edouard: Yes

PR #527, opened 2017-02-01
OXT-923: Do not store WPA PSK keys in plaintext in the dom-store.

troyc:
This patches network-manager-applet so that PSK passwords are only ever dealt with in their hashed form. A separate commit also disables the "show password" checkboxes.

### DIFF
--- a/recipes-connectivity/network-manager-applet/network-manager-applet/disable-show-password.patch
+++ b/recipes-connectivity/network-manager-applet/network-manager-applet/disable-show-password.patch
@@ -1,0 +1,39 @@
+--- network-manager-applet-0.9.2.0/src/wireless-security/wireless-security.c	2017-01-13 10:50:54.830968023 -0500
++++ network-manager-applet-0.9.2.0/src/wireless-security/wireless-security.c	2017-01-16 15:23:41.424841517 -0500
+@@ -37,6 +37,20 @@
+ #include "wireless-security.h"
+ #include "eap-method.h"
+ 
++static void
++removeCheckboxes (GtkBuilder *builder)
++{
++	GSList *objects;
++	const char showCheckbuttonPrefix[] = "show_checkbutton_";
++
++	for (objects = gtk_builder_get_objects (builder); objects != NULL; objects = objects->next) {
++		const gchar *id = gtk_buildable_get_name (objects->data);
++		if (id && strncmp (id, showCheckbuttonPrefix, sizeof (showCheckbuttonPrefix) - 1) == 0)
++			gtk_widget_hide (GTK_WIDGET (objects->data));
++	}
++	g_slist_free(objects);
++}
++
+ GType
+ wireless_security_get_g_type (void)
+ {
+@@ -197,6 +211,7 @@
+ 	}
+ 	g_object_ref_sink (sec->ui_widget);
+ 
++	removeCheckboxes (sec->builder);
+ 	return sec;
+ }
+ 
+@@ -314,6 +329,7 @@
+ 		if (eap_default_widget)
+ 			gtk_widget_grab_focus (eap_default_widget);
+ 	}
++	removeCheckboxes (eap->builder);
+ 
+ 	eap_method_unref (eap);
+ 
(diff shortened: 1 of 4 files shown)


recipes-connectivity/network-manager-applet/network-manager-applet/disable-show-password.patch
```diff
@@ -0,0 +1,39 @@
+--- network-manager-applet-0.9.2.0/src/wireless-security/wireless-security.c	2017-01-13 10:50:54.830968023 -0500
++++ network-manager-applet-0.9.2.0/src/wireless-security/wireless-security.c	2017-01-16 15:23:41.424841517 -0500
+@@ -37,6 +37,20 @@
+ #include "wireless-security.h"
+ #include "eap-method.h"
+ 
++static void
++removeCheckboxes (GtkBuilder *builder)
```
jean-edouard: Why not a patch that removes the checkbox creation code, instead of adding code to hide them?

recipes-connectivity/network-manager-applet/network-manager-applet_0.9.2.0.bbappend
```diff
@@ -35,6 +35,8 @@ SRC_URI += " \
            file://xc-menus.patch;patch=1 \
            file://disable_available_to_all_users_checkbox.patch;patch=1 \
            file://default-certs-dir.patch;patch=1 \
+		   file://always-use-psk-hash.patch;patch=1 \
```
jean-edouard: Indent

recipes-connectivity/network-manager-applet/network-manager-applet/disable-show-password.patch
troyc: Hiding them is the simplest solution. Removing the checkboxes altogether would require you to also change the code that references them. The result would be a more complicated and bug-prone patch.

recipes-connectivity/network-manager-applet/network-manager-applet_0.9.2.0.bbappend
troyc: fixed

rossphilipson:
Can you explain in more detail exactly what you change does (it is hard to the the whole picture just from the PR)?

Also I don't think removing the show password checkbox is really necessary and that feature may be useful to people.

troyc:
There's two separate commits. The first commit causes WPA-PSK passwords to be stored as hashes rather than as plaintext (in the dom-store). It does this by patching network-manager-applet to hash the password as soon as it's read from the GTK text input. The second commit hides all "show password" checkboxes.

Removing the "show password" checkboxes is especially prudent with this approach because the user would only ever be able to see the hashed forms of passwords, which might not be what the user expected. (The best way to let the user be able to see the passwords without storing them in plaintext would be to use symmetric encryption, which would be more complicated and less secure.)

Ultimately, you have to decide whether this trade-off of usability for security is appropriate for OpenXT.

rossphilipson:
Thanks for the explanation. The first part makes a lot of sense - I guess I wonder why something like that wasn't done in the first place. Storing the passwords as text has always been an issue.

I thought the "show password" checkbox was just to allow you to see what you are typing as you are entering a new password. If it is trying to show the existing password, I can see how this would now not make sense. I guess I don't really have a problem with removing this given the way it works.

jean-edouard:
@troyc, if you really want to disable that checkbox to show the password (because it's confusing to see a hash there), please do so only when a hash is indeed present (the way you do it, by letting the thing create the checkbox and then hiding it, should make it easy to add that check).
Otherwise, just rename "Show password" to something like "Show password / hash".

troyc:
I updated the PR to only disable the show password checkbox when a hash is present. Note that I haven't been able to test the dialog that appears when you first try to connect to a wireless network because wireless connections aren't currently working right in OpenXT on my test box.

rossphilipson:
I will try to get this one in a build and test it now that wireless is working thought it may be blocked by thsi now: https://openxt.atlassian.net/browse/OXT-1066

Sorry this one got overlooked for so long.

rossphilipson:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/750
Will test.

rossphilipson:
This is looking good. I put the test procedure I did on the ticket. Will merge with RM signoff.

PR #530, opened 2017-02-03
[libxl] Fix tapdisk cleanup

crogers1:
This commit fixes two issues. First, we remove the call to
  clean the backend xenstore path in initiate_device_remove. When
  the devices are actually called for removal, the xs path no longer
  exists, so tapdisk destruction fails since it needs the tapdisk-params
  xs entry.  Second, the tapdisk_is_shared really should only apply to
  cdroms.  A domain and its stubdomain have the same value for
  tapdisk-params; therefore, the function will think the vhd is shared and
  the tap device will not be cleaned up.

  OXT-435

### DIFF
--- a/recipes-extended/xen/files/libxl-xenmgr-support.patch
+++ b/recipes-extended/xen/files/libxl-xenmgr-support.patch
@@ -396,22 +396,6 @@ Index: xen-4.6.4/tools/libxl/xl_cmdimpl.c
      rc = libxl_domain_destroy(ctx, domid, 0);
      if (rc) { fprintf(stderr,"destroy failed (rc=%d)\n",rc); exit(-1); }
      libxl_update_state_direct(ctx, uuid, "shutdown");
-Index: xen-4.6.4/tools/libxl/libxl_device.c
-===================================================================
---- xen-4.6.4.orig/tools/libxl/libxl_device.c
-+++ xen-4.6.4/tools/libxl/libxl_device.c
-@@ -904,6 +904,11 @@ void libxl__initiate_device_remove(libxl
-             }
-         }
- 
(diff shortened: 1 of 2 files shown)


crogers1:
Discovered while fixing issue addressed with this PR: https://github.com/OpenXT/manager/pull/56

They should be merged together.

recipes-extended/xen/files/libxl-xenmgr-support.patch
```diff
@@ -396,22 +396,6 @@ Index: xen-4.6.4/tools/libxl/xl_cmdimpl.c
...
      libxl_update_state_direct(ctx, uuid, "shutdown");
-Index: xen-4.6.4/tools/libxl/libxl_device.c
-===================================================================
---- xen-4.6.4.orig/tools/libxl/libxl_device.c
-+++ xen-4.6.4/tools/libxl/libxl_device.c
-@@ -904,6 +904,11 @@ void libxl__initiate_device_remove(libxl
-             }
-         }
- 
-+        if(!libxl__xs_read_checked(gc, t, libxl__device_backend_path(gc, aodev->dev), &state))
```
rossphilipson: So is this just not cleaned up anywhere?

rossphilipson:
Built here, testing now:
http://openxt.ainfosec.com/builds/custom/master/587/

recipes-extended/xen/files/libxl-xenmgr-support.patch
crogers1: It should be cleaned up in libxl_device.c:libxl__device_destroy().  I think this specific hunk was a mistake to have included in the first place.  I'm pretty sure it was something I added, but I don't have the history or any comments to know for sure.

rossphilipson:
Tested, no issues seen. Merging soon.

PR #533, opened 2017-02-07
[libxl] Support read-only vhds on domains with

crogers1:
linux-based stubdoms. There is an unnecessary restriction in xl
  on enforcing rw for vhds.  As long as the filesystem in the guest
  is ok with being mounted ro, there's no reason not to support it.

  OXT-435


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/588
Code LGTM.
Tested r/o disks before and after: crashes before and works just fine after.
Merging soon.

PR #544, opened 2017-02-16
STABLE-6: Add the KabyLake SINIT acm

rneilturner:


rossphilipson:
LGTM merging soon. We don't have any systems that use KBL SINIT so we are not able to test directly.

PR #549, opened 2017-02-22
[libxl] Fix tapdisk cleanup when 2 domains share

crogers1:
a read-only disk. The current code fails to consider the case
  of non-cdrom type read only disk. This should handle both cases
  now.

  OXT-961


jean-edouard:
Look like you're skipping all the stubdoms here, which might cause trouble.
I wrote a similar (the same?) function there, which only ignores our own stubdom:
https://github.com/OpenXT/cdrom-daemon/blob/master/src/blktap.c#L78

Maybe we should use the same here.

Ultimately, we should move all blktap management from xl to cdrom-daemon, that'll make things much easier to manage.

crogers1:
I don't think there's a difference between skipping just our stubdom and skipping all stubdoms.  Stubdom should have identical tap devices to the domain its stubbing, which makes it irrelevant when checking shared tap devices.  Is there a case where stubdom tap devices differ from their base domain?

jean-edouard:
That's calling for race conditions to happen.
If you shutdown a VM, I can see the main domain being gone and the stubdom still "using" the tapdev.

crogers1:
But that was the original problem.  It was counting stubdoms as 'shared' and not tearing down the tapdisk when the domain is shutdown.  If the main domain gets shutdown, the stubdom is shutdown too, and the tapdev should be closed, unless another 'main' domain is using it too.

jean-edouard:
Ok, domain A and B are running with stubdoms and sharing a tapdev.
You shutdown A and B.
When A enters your code, B is gone but its stubdom is still shutting down.
The code will ignore all stubdoms and nuke the tapdev.
That will either crash B's stubdom or freak out when trying to remove the tapdev that's already gone.

jean-edouard:
Also it's an easy fix, just check if the stubdom is ours, and only skip it in that case.

dpsmith:
Just 2 cents on an earlier comment,  I don't think you want cdrom-daemon to be responsible for tapdev management or change name to storage-daemon and make it responsible for managing all storage. Of course that would require that a proper interface for such a daemon be defined first.

jean-edouard:
I should have been more precise, I was just talking about cdrom tapdevs, which are the only one that can be shared between VMs. If those were managing them entirely in cdrom-daemon we wouldn't need this PR.

dpsmith:
No worries, thanks for the clarification.

crogers1:
Implemented @jean-edouard 's suggestion, instead of filtering every stubdom, it only filters the stubdom of the domain that's getting its devices cleaned up (ie, shutdown).

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/701

jean-edouard:
Merging soon.

PR #550, opened 2017-02-22
key-functions: make rsa key ref's relative

dpsmith:
The call into recovery_unlock in certain upgrade use cases from the
old xc scheme may have the target rootfs mounted on an alternative
location. Allow the passing of the alternative mount point as the
third argument.

OXT-955


jean-edouard:

rossphilipson:
Goes with:
https://github.com/OpenXT/installer/pull/29

Testing...

rossphilipson:
Tested, sorry doesn't work :(

rossphilipson:
Test failed, fix not working. See other PR for details.

PR #558, opened 2017-02-28
OXT-580: dmidecode: Use upstream recipe and upgrade to 3.0.

eric-ch:
Original recipe has a type in LICENSE field (s|files://|file://).

Anyway, might as well use the upstream recipe. Since it is an upgrade
and there is not .inc, it requires a duplicate unfortunately.

OXT-580

### DIFF
--- a/recipes-devtools/dmidecode/dmidecode_3.0.bb
+++ b/recipes-devtools/dmidecode/dmidecode_3.0.bb
@@ -1,17 +1,20 @@
-DESCRIPTION = "DMI (Desktop Management Interface) table related utilities"
-SECTION = "console/utils"
+SUMMARY = "DMI (Desktop Management Interface) table related utilities"
 HOMEPAGE = "http://www.nongnu.org/dmidecode/"
 LICENSE = "GPLv2"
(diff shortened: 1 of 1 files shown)


jean-edouard:

jean-edouard:
LGTM, builds fine works fine.
However, @eric-ch, I don't understand why we need a duplicate recipe.

eric-ch:
@jean-edouard The upstream recipes for dmidecode do not have a .inc file to put configuration in common across versions. Since this is an upgrade, it requires dmidecode_3.0.bb, which cannot be .bbappend in any sane way I could think of.

jean-edouard:
Oh, you mean OE upstream uses a version below 3.0?
Why do we need to upgrade then?

eric-ch:
@jean-edouard because we were already using dmidecode 3.0 in our layer and downgrading did not seem to be interesting anyway. Morty will have dmidecode 3.0, so it seems ok to have it here until we upgrade (again, this is dmidecode_3.0.bb from morty).

jean-edouard:
Oh now I get it! Thanks :)
Merging soon.

PR #561, opened 2017-02-28
OXT-580: ionice: Use util-linux recipe upstream.

eric-ch:
No need to split the package if it is deployed anyway in dom0.


jean-edouard:

jean-edouard:
LGTM, makes sense, merging soon.

PR #563, opened 2017-02-28
Generic cleanup of some recipes.

eric-ch:
Planing for future upgrades (Morty):
Some recipes should append upstream OE ones for various versions.
Also remove some dead code or added configuration that is upstream.


jean-edouard:

recipes-connectivity/iproute2/iproute2_4.%.bbappend
```diff
@@ -1,10 +1,7 @@
-PR .= ".1"
+PR .= ".2"
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
-EXTRA_OEMAKE += "-j 1"
```
jean-edouard: Are you sure this is ok to remove? Is it documented that they used to have trouble building in parallel and that they don't anymore?

jean-edouard:
Appart from the question above, LGTM.

recipes-connectivity/iproute2/iproute2_4.%.bbappend
eric-ch: I was doubtful it would be a problem so I tried in different trees and had no issue. Also, this is appending a recipe upstream which is assumed to build using multiple threads.

jean-edouard:
Fair enough, merging soon.

eric-ch:
OXT-976

PR #564, opened 2017-03-03
OXT-980: fix xenstat-disable-tmem-use patch to avoid compile warning

dozylynx:
about a variable declared but not used.


rossphilipson:
Makes sense, merging soon.

PR #565, opened 2017-03-03
OXT-972: initramfs: Setup TPM for measurement only.

eric-ch:
There is no need to setup TPM if it is not used.

OXT-972


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/684

jean-edouard:
Yup, thanks, merging soon.

PR #567, opened 2017-03-03
Prepare Xen for meta-virtualization

dozylynx:
[description here updated to reflect code revision]

This PR refactors the Xen recipes to adopt meta-virtualization as an upstream source. This PR requires a corresponding PR in openxt.git to introduce the meta-virtualization layer.

Two commits:
* refactor the OpenXT Xen recipes to require and extend the meta-virtualization Xen recipe.
* change the references to the Xen recipe across the rest of the build to match the new meta-virtualization compatible package and recipe naming.

A side-effect of this change is that it is much clearer how the 64-bit hypervisor build is currently being performed by using host tools rather than native tools.

It has been built, booted and had light testing which passed. The filesystem has been compared with a build without this change: there are differences, but are expected. The meta-virtualization packaging is finer-grained and some unnecessary scripts and binaries are now omitted from the dom0 filesystem.


eric-ch:
Although I am all for using meta-virtualization instead of OpenXT homebrewed recipes, this does not feel great: Why not move directly to meta-virtualization?

It looks like it would only require a [xen_4.6.4.bb](http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization/tree/recipes-extended/xen/xen_4.8.0.bb) in [xenclient-oe](https://github.com/OpenXT/xenclient-oe/) with a ```PREFERED_xen``` variable in the distro, and the hacks for the 64b hypervisor compilation by the host environment. Actually the meta-virtualization layer has good includes for that.

dozylynx:
Thanks, Eric - this is the kind of feedback I was hoping for.

Moving directly to meta-virtualization will involve selecting the target release branch which will need to include a recipe for the same major version of Xen that OpenXT uses, (currently 4.6, so krogoth) plus fixing up any recipe collisions (eg. ALSA), and selection of that branch will likely force the same release branch to be used in the upstream repos. It's not a simple single step to take, hence this smaller proposal as step one.

eric-ch:
Yes, we should select a release branch, but even so, there is no real requirement to use the xen_%.bb recipe version since most of the configuration is done in .inc files which we can use in OpenXT recipes, changing the versions and applying the patch-queue.

Following the rest of the current layers, using [meta-virtualization jethro branch](http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization/log/?h=jethro), [xen_4.6.1.bb](http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization/tree/recipes-extended/xen/xen_4.6.1.bb?h=jethro) should be easy enough to duplicate in xen_4.6.4.bb in xenclient-oe.git and apply the OpenXT patch-queue (even have the patch-queue in a .bbappend to be ready for futur upgrades).
... Actually we could even submit the recipe updated to xen 4.6.4 to meta-virtualization.

I must be missing the point as I do not understand how ALSA confict with this?

dozylynx:
ok, as it's been a couple of months since my last look at that, I'll take another pass at using a meta-virtualization layer directly again and see how it goes. Last time the ALSA recipe needed masking in xenclient-oe's layer.conf to prevent build errors.

dozylynx:
Building with current meta-virtualization jethro looks good, so I've refreshed the branch to use the layer directly and no longer include copies of upstream recipes, and rewritten the description in this PR. A corresponding PR in openxt.git introduces the layer submodule.

eric-ch:
(Build 664)[http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/664] passes smoke testing.

eric-ch:
LGTM, +1.

PR #568, opened 2017-03-07
Get the BlockBackend name directly from the device we have already.

rossphilipson:
OXT-853


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/672
Segfault fixed, merging soon.

PR #571, opened 2017-03-09
cdrom-daemon: adding recipes

jean-edouard:
OXT-847


jean-edouard:
Goes with https://github.com/OpenXT/xenclient-oe/pull/572 and https://github.com/OpenXT/manager/pull/69

PR #577, opened 2017-03-15
[libxl] Don't rm state node, causes xs race

crogers1:
After shutdown, immediately removing the state node caused a race
  with listeners using xs_watch, the watch would never fire, sending
  a state update notify. The xs_rm isn't really necessary here since
  xenmgr can just cleanup the state node when/if the domain is deleted.

  OXT-1003


jean-edouard:
Build here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/702
Fixed the issue for me.
Merging soon.

PR #578, opened 2017-03-15
Rewrite the linux-firmware recipe

jean-edouard:
OXT-977

### DIFF
--- a/recipes-kernel/linux-firmware/linux-firmware_git.bb
+++ b/recipes-kernel/linux-firmware/linux-firmware_git.bb
@@ -1,134 +1,22 @@
 SUMMARY = "Firmware files for use with Linux kernel"
 SECTION = "kernel"
 
-LICENSE = "\
-    Firmware-Abilis \
-    & Firmware-agere \
-    & Firmware-amd-ucode \
-    & Firmware-atheros_firmware \
-    & Firmware-broadcom_bcm43xx \
-    & Firmware-ca0132 \
-    & Firmware-chelsio_firmware \
-    & Firmware-cw1200 \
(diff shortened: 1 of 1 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/708

dozylynx:
@cjp256 : please could you review this PR?

eric-ch:
+1 LGTM.

Firmwares are installed in dom0 and NDVM using build 708.

The recipe might as well only deal with the firmware used in OpenXT since it is duplicated from usptream until the OE recipe is upgraded.

jean-edouard:
I don't have any, if you do and want to add it to the HCL, then just add your firmware to the recipe in a separate PR.

dpsmith:
Just because none of us can test and put on the HCL doesn't mean we should allow the firmware of common cards to be included. By doing so, you have stopped end-users who might want to attempt and report back from doing so.

jean-edouard:
Ok, sorry.

dpsmith:
You didn't need to close the PR, wasn't trying to reject it. Like I said, I am for trimming down some of the firmware just don't exclude the more common devices.

eric-ch:
In this case I believe the alternative is to send a PR upstream with the missing firmware and use the previous duplicated recipe with split packages in dom0 and NDVM in the mean time. This would radically modify the changes introduced by that PR.

jean-edouard:
This PR adds firmware to the target, it doesn't remove any, but I don't have time to argue endlessly, I'll just diverge my downstream a bit more.

PR #580, opened 2017-03-16
[libxl] Send shutdown signal before rebooted.

crogers1:
In testing, it seems much of the platform is dependent on seeing
  a "shutdown" state update sent across dbus. This commit makes
  sure we send that update before changing to the rebooted state.

OXT-1006
OXT-1005
OXT-990  

### DIFF
--- a/recipes-extended/xen/files/libxl-domain-state.patch
+++ b/recipes-extended/xen/files/libxl-domain-state.patch
@@ -216,15 +216,17 @@ Index: xen-4.6.4/tools/libxl/xl_cmdimpl.c
      if (!paused)
          libxl_domain_unpause(ctx, domid);
  
-@@ -2992,6 +2996,7 @@ start:
+@@ -2992,6 +2996,9 @@ start:
                   * re-creation fails sometimes.
                   */
                  LOG("Done. Rebooting now");
-+                libxl_update_state_direct(ctx, d_config.c_info.uuid, "rebooted");
++                libxl_update_state_direct(ctx, d_config.c_info.uuid, "shutdown"); //Sleep here because daemons with an xs_watch on this node
++                sleep(2);                                                         //won't see the "shutdown" event, just the "rebooted" one.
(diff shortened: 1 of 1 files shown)


jean-edouard:
Urg `sleep(2)`... I don't have anything better to offer though, so I'll test like that.

eric-ch:
If I understand correctly the usual behavior to accommodate with xenmgr when synchronizing on a Xenstore node is to sleep(2).
That would be a hack, but I would not want to be send in Haskell just yet to try and offer a way around that. This has to be documented and/or commented somewhere anyone reading or editing that code later will understand the reason behind it.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/713

jean-edouard:
This PR seems to fix OXT-1006, but not OXT-990.

jean-edouard:
My bad, forgot to "Always assign", OXT-990 might be fixed as well.

crogers1:
@eric-ch : commented on why the sleep is there. The problem ultimately lies in the xenstore libs and the behavior of xs_watch when multiple xs_writes occur in quick succession.  For example, an xs_watch("/some/node") in some daemon will not trigger if xl were to do something like 
```c
xs_write("/some/node", "somevalue");
xs_rm("/some/node");
```
The same can be said for
```c
xs_write("/some/node", "valueA");
xs_write("/some/node", "valueB");
```
The accompanying xs_watch only triggers with "valueB".

eric-ch:
@crogers1 That sounds familiar actually, and very likely to trigger races and errors. Thank you for documenting it.

rossphilipson:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/732

Testing

rossphilipson:
This LGTM w/r to the 3 tickets which I will list in the PR. Merging tomorrow morning.

PR #581, opened 2017-03-17
OXT-1019: busybox: Override configuration from other layers.

eric-ch:
Busybox is configured by merging .cfg files listed in SRC_URI. Different
layers use that (e.g. openembedded-core and meta-virtualization), then
the configurations in .cfg files are merged with the defconfig from
SRC_URI.

As some of the listed features are not required in OpenXT, create
matching .cfg files to override the settings in additional layers:
- lspci, provided by pciutils
- lsusb, provided by usbutils
- login-utilities, login & passwd provided by shadow (rest is not used)
- mdev, unused
- cifs, unused
- ps-extra, unused

OXT-1019


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/738

jean-edouard:
LGTM, merging soon.

PR #582, opened 2017-03-18
OXT-1020: Use upstream OCaml Xenstored

dozylynx:
Build, package and use the upstream OCaml Xenstored rather than the stale OpenXT forked OCaml Xenstored.

Another PR in toolstack.git deletes the forked code.


jean-edouard:
Fails to build: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/722/steps/Build/logs/stdio

dozylynx:
Update: have reproduced and confirmed the build problem, apparently due to the OCaml tool recipe missing dependencies. Tracking the fix for that at OXT-1022 , with a candidate fix for it undergoing testing now.

dozylynx:
This is ready with this Ocaml dependency fix: https://github.com/OpenXT/meta-openxt-ocaml-platform/pull/3

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/734

jean-edouard:
LGTM, smoke tested, merging tomorrow.

recipes-extended/xen/files/oxenstored.conf
```diff
@@ -0,0 +1,45 @@
+# OpenXT xenstored config
+
+# Where the pid file is stored
+pid-file = /var/run/xenstored.pid
+
+# Randomly failed a transaction with EAGAIN. Used for testing Xs user
+test-eagain = false
```
rossphilipson: This is the only param that is different. @jean-edouard and I are looking what this does and it seems like the logic in xenstored is backwards/broken. We need to look at this more closely but right now it looks like setting this to true is really disabling it.

jean-edouard:
We definitely want test-eagain to be false. Now the logic seems backwards in the oxenstored code, and false seems to really mean true, but that's a separate issue.

I suggest we file a ticket to investigate the logic in oxenstored.
See: http://xenbits.xenproject.org/gitweb/?p=xen.git;a=blob;f=tools/ocaml/xenstored/transaction.ml;h=6b37fc2ad0a459858aebddba450e91799f7d609e;hb=HEAD#l198

Merging soon.

PR #583, opened 2017-03-20
file: use release tarball instead of git

jean-edouard:
OXT-1021

### DIFF
--- a/recipes-devtools/file/file_5.24.bbappend
+++ b/recipes-devtools/file/file_5.24.bbappend
@@ -1,0 +1,10 @@
+SRC_URI = "ftp://ftp.astron.com/pub/file/file-5.24.tar.gz \
+        file://debian-742262.patch \
+        file://0001-Add-P-prompt-into-Usage-info.patch \
+        file://host-file.patch \
+        "
+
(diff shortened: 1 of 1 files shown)


rossphilipson:
Since the build is currently broken and this fixes it I am going to merge it very soon.

dpsmith:
Would it not be better to just update the hash like OE upstream has done?[1](http://git.openembedded.org/openembedded-core/commit/?h=krogoth&id=b35225c88ff681a4a903f7fb4612ac768214f539).

jean-edouard:
I'm not sure I see the point of using a hash, and it'll likely break again!

jean-edouard:
(This is how I found the release tarball: https://github.com/file/file -> http://www.darwinsys.com/file/ -> ftp://ftp.astron.com/pub/file/)

dpsmith:
As long as we have provenance and are willing to maintain.

rossphilipson:
So what is the final story on this? We need to unbreak the build.

dpsmith:
Pull it

rossphilipson:
Ack, merging.

PR #585, opened 2017-03-20
xen: Don't boot unless IOMMU is enabled

jean-edouard:
OXT-826


dozylynx:
The previous system behaviour, prior to the libxl toolstack change, was to boot even with the IOMMU disabled but to disable networking in that case.

In the installer, it would boot but present a message to the user that the system was incorrectly configured and what was wrong.

In a VM, it was possible to interactively install the OpenXT software, even without VTd available.

This PR is not a fix for OXT-826.

jean-edouard:
There's no reason for OpenXT to install/boot on systems that don't have vt-d.
Once this fix is in, OXT-826 won't be a problem anymore, aka fixed.

dpsmith:
And there is no reason OpenXT should install/boot on systems without TPMs but yet it was filed as a regression because that is how it worked before and it "might" be useful for testing. This is the same case. Prior behavior is that OpenXT/XenClientXT would install on a platform without VT-d, warn the user, and then not allow the NDVM to be started. Not implementing this exact behavior is a regression and any attempt to implement an alternate behavior is a feature/enhancement.

jean-edouard:
IMO it's not the same, if you can guarantee the physical safety of your machine you don't need TPM, but that's beyond the point.

This PR is my fix for a bug in OpenXT. If the community doesn't want it that's fine, but I'm not writing another fix for this.
Keep in mind that any rejected PR is a new potential divergence with a downstream project (our testers!).

dpsmith:
In terms of security gains, your right that its not the same, its worse to run without a TPM than without VT-d. Any gains of device isolation is negate by the fact that one has no idea of what was boot or is running. In terms of declaring what is a regression it is exactly the same. Prior to the addition of a feature there was a way the system behaved and then how the system's behavior change after the addition. This is what occurred in both cases. You can't say one is a regression that requires the restoration the original behavior and then decide that changing the behavior based on what you decided was a faulty behavior. Then you are no longer fixing a bug but proposing a new enhancement which should be a separate.
Your thinking is backwards with regards to downstream project. If they so choose to deviate from the upstream project, then that is their deviation. As the lead for another downstream project, with several deviations understand this and thus issues caused by this are our own and not that of the OpenXT community. We also are trying to work with the community to upstream our deviations where we can or feel the community will be receptive.

richpersaud:
Please discuss this topic on the mailing list so the community can define the requirement.

PR #586, opened 2017-03-22
OXT-1024: Disable xenstored fault injection

dozylynx:
The logic within xenstored for operating on the config file flag is
inverted, so set the config value to the right value to obtain the
intended disabled behaviour.

### DIFF
--- a/recipes-extended/xen/files/oxenstored.conf
+++ b/recipes-extended/xen/files/oxenstored.conf
@@ -4,7 +4,9 @@
 pid-file = /var/run/xenstored.pid
 
 # Randomly failed a transaction with EAGAIN. Used for testing Xs user
-test-eagain = false
+# Note: this flag is boneheadedly inversed: "true" is the correct
+# production setting that *disables* the fault injection behaviour.
+test-eagain = true
 
 # Activate transaction merge support
 merge-activate = true


jean-edouard:
-1, we should write a patch in ocaml to fix the broken code instead

rossphilipson:
Sorry it was actually my suggestion to do it that way. @jean-edouard way is better.

dozylynx:
There is more involved than just inverting the logic in the code: it's highly likely that other users of that version of xenstored, which is the upstream recommended version, have the flag set to true in their config file.

The config file may or may not get replaced when people upgrade their xen + tools, so I can't see upstream accepting a code change with a simple invert of the flag behaviour.

Stripping that functionality out entirely is might be an acceptable option but even that might not fly because some people may actually use it -- it does have a legitimate purpose, which is why it was implemented.

Inverting the flag with a comment is probably the least worst option for the OpenXT code for now, until upstream perspective has been determined.

rossphilipson:
I will post to xen-devel tomorrow and ask them how they would like it fixed. We can hold off on this PR until we get an answer.

richpersaud:
We are trying to avoid the cost of maintaining local fixes and forks.  A proper fix should be made in upstream Xen and and will picked up locally once accepted upstream. In the meantime, we can document the confusing configuration.

jean-edouard:
I did some actual work, wrote a test program and got interesting results.

With the new xenstore daemon, oxenstored, having test_eagain true or false does have any impact, and every (non-conflicting) transaction succeeds. I actually suspect that all the options in the config file (/etc/xen/oxenstored) are being ignored (see OXT-1026).

With the old xenstore daemon, test_eagain behaves as expected! It breaks 1/3rd of the transactions if set to true, and doesn't if set to false. And we've always had it set to true......

PR #587, opened 2017-03-23
key-functions: make rsa key ref's relative

dpsmith:
The call into recovery_unlock in certain upgrade use cases from the
old xc scheme may have the target rootfs mounted on an alternative
location. Allow the passing of the alternative mount point as the
third argument.

OXT-955


rossphilipson:
Tested, OTA upgrade to custom build worked. Merging soon.

http://openxt.ainfosec.com/builds/custom/master/741

MagOXT:
Tested with
http://openxt.ainfosec.com/builds/regular/master/743/
OK OTA upgrade
KO PXE upgrade
KO USB stick upgrade

PR #598, opened 2017-03-30
Allow writable CDs and fix guest crash on write.

rossphilipson:
This fixes CD writing to what it is in stable-6 currently. Some things fail
like formatting the CD or dragging files/burning. Other things work like
right clicking an ISO and burning it to a CD. A separate ticket will be
opened to track those failures but this fixes the RO problem and the crash.

OXT-931

### DIFF
--- a/recipes-openxt/qemu-dm/qemu-dm-2.6.2/atapi-pass-through.patch
+++ b/recipes-openxt/qemu-dm/qemu-dm-2.6.2/atapi-pass-through.patch
@@ -1722,7 +1722,17 @@ Index: qemu-2.6.2/blockdev.c
      int cyls, heads, secs, translation;
      int max_devs, bus_id, unit_id, index;
      const char *devaddr;
-@@ -985,6 +991,13 @@ DriveInfo *drive_new(QemuOpts *all_opts,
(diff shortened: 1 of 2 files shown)


jean-edouard:
Build here http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/746

eric-ch:
Tested build with win7 & win10 guests on a Dell 9020, with and without stub-domain:
- No crash, but...
- Trying to read the disk triggers "An error occured while ejecting 'DVD RW Driver (D:)'
Could not read or write to the disk.

rossphilipson:
Also built here with just the CD fixes:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/744

rossphilipson:
I just tested 744 on my HP EliteDesk 800 G1 SFF and I was able to burn an ISO to the disk (with SELinux on and off) then read it.

cjp256:
LGTM

rossphilipson:
Just burned and then read an ISO on a Dell E6540 laptop with 744 without problems.

cjp256:
LGTM, ATAPI PT has been known to be flaky at times anyhow :)

PR #599, opened 2017-03-30
libxl: atapi-pt: fix syntax and allow readwrite

jean-edouard:
OXT-1045


crogers1:
+1

jean-edouard:
Build here http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/746

eric-ch:
Tested build with win7 & win10 guests on a Dell 9020, with and without stub-domain:
- cmdline is: ```-drive file=atapi-pt-v4v:/dev/bsg/0:0:0:0,if=atapi-pt,media=cdrom,format=atapi-pt-fmt```, but...
- Trying to read the disk triggers "An error occured while ejecting 'DVD RW Driver (D:)'
Could not read or write to the disk.

PR #601, opened 2017-04-03
keymanagement: handle non-ml upgrade from legacy

dpsmith:
Under legacy key management different LUKS slots where used for key storage.
Since we are not remapping LUKS slots when upgrading from a system using the
legacy slots, make the new key managent attempt a general unlock if the unlock
using the expected slot fails.

OXT-1054


rossphilipson:
Was this tested and what would one do to test it?

dpsmith:
Its simple enough that I thought inspection might be enough, but if you would like to test, then upgrade a non-ml 6.0.0 with 7.0.0

rossphilipson:
Ok, I see. It LGTM. I will signoff and merge soon.

rossphilipson:
RM Signed-off-by: Ross Philipson philipsonr@ainfosec.com

PR #602, opened 2017-04-05
ACPI Fixes

jandryuk:
These two changes fix up the ACPI power button support in Xen hvmloader's ASL and QEMU's emulation.  Shutdown is more consistent with these changes.

Note, Windows guests must have "When I press the power button" set to "Shut down" to actually trigger a shutdown.  This is not always the case on install.  Specifically, Windows 10 can leave the option unset when installed and not respond to the button.

OXT-1025

### DIFF
--- a/recipes-openxt/qemu-dm/qemu-dm-2.6.2/acpi-pm-feature.patch
+++ b/recipes-openxt/qemu-dm/qemu-dm-2.6.2/acpi-pm-feature.patch
@@ -213,7 +213,7 @@ Index: qemu-2.6.2/hw/xen/xen_acpi_pm.c
 ===================================================================
 --- /dev/null
 +++ qemu-2.6.2/hw/xen/xen_acpi_pm.c
-@@ -0,0 +1,1057 @@
+@@ -0,0 +1,1127 @@
 +/*
 + * APCI PM feature for battery/AC/lid management for OpenXT guests.
 + *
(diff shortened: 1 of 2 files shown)


dpsmith:
Please amend your commits with the JIRA issue number as well.

recipes-openxt/qemu-dm/qemu-dm-2.6.2/acpi-pm-feature.patch
```diff
@@ -1029,17 +1035,79 @@ Index: qemu-2.6.2/hw/xen/xen_acpi_pm.c
...
++ * Track state of SLP/WAK bits for power and sleep buttons.  The fields
++ * are WriteAsZero, writing 1 clears the bit.
++ */
++static void acpi_button_sts_write(void * opaque, hwaddr addr,
++                                  uint64_t val, uint32_t size)
++{
++    XenACPIPMState *s = opaque;
++
++    if (addr == 0) {
++        s->button_status &= ~val;
```
rossphilipson: I think and'ing it with the ~3 to make sure no other bits are set makes sense here just to be on the safe side so we never get anything but bits 0 and 1 set.

rossphilipson:
Reviewed, one comment, rest LGTM. I will throw it in a custom build and we will test it. Can you put test instructions in the ticket please?

recipes-openxt/qemu-dm/qemu-dm-2.6.2/acpi-pm-feature.patch
jandryuk: So button_status is only ever set by other QEMU code for the the two bits.  The above "write" is really only clearing set bits, so we can never set additional bits.

I can mask down val, but I didn't think it was necessary.

rossphilipson: It is being written by the AML code there through the op region. An entire byte is coming through - it probably only has 2 bits set at most but since the rest of the byte is undefined...

jandryuk: Bits 2-7 of button_status are 0, so the bitwise & can never set them.

&= ~ can only clear bits from button_status - it cannot set bits in button_status.  Masking down val doesn't change that.

rossphilipson: How do you know bits 2 - 7 are always 0 since the byte comes from outside QEMU when the ACPI code in the guests writes that byte? it is probably true but are you 100% it is always true?

rossphilipson: Nothing every writes directly to button_status other than init to zero so I was wrong. Never mind.

rossphilipson:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/750
Will test.

rossphilipson:
Tested LGTM . I shutdown a Win7 x64 VM (no tools) multiple times w/o anything unexpected happening. Will merge after RM signoff.

PR #605, opened 2017-04-06
STABLE-6: ACPI Fixes

jandryuk:
These two changes fix up the ACPI power button support in Xen hvmloader's ASL and QEMU's emulation. Shutdown is more consistent with these changes.

Note, Windows guests must have "When I press the power button" set to "Shut down" to actually trigger a shutdown. This is not always the case on install. Specifically, Windows 10 can leave the option unset when installed and not respond to the button.

OXT-1025


rossphilipson:
Same code, same problem so I will merge this after making sure it builds.

rossphilipson:
This built: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/757

It is good to go in, it was tested in master. Merging soon.

PR #606, opened 2017-04-06
OXT-1021: file: use source tarball from tagged releases.

eric-ch:


dpsmith:
Were you able to successful build yourself? The recipe is straightforward enough that if you were able to build with it, then I would be willing to sign off.

jean-edouard:
LGTM, builds in my tree.

jean-edouard:
The two tarballs are vastly different, since the actual release tarball is autoconf-ed, when the GitHub one is not. Some seemingly generated code in the .c files is also different, but it should only be slight cosmetic differences.

dpsmith:
Are you still good with it then?

jean-edouard:
Yes, just thought it was worth mentioning.

rossphilipson:
RM Signed-off-by: Ross Philipson philipsonr@ainfosec.com

PR #608, opened 2017-04-07
STABLE-6: OXT-1021: file: use source tarball from tagged releases.

eric-ch:
(cherry picked from commit 6e6d40b62ba9b72f98cf0509a0cf54e9d926f999)


rossphilipson:
Same change, same stuffs, good to go into 6.0.1

PR #610, opened 2017-04-13
Remove dependency on xen-libxl for non-dom0 machines

jean-edouard:
OXT-1016

### DIFF
--- a/recipes-extended/xen/xen-ocaml-libs.bb
+++ b/recipes-extended/xen/xen-ocaml-libs.bb
@@ -1,0 +1,110 @@
+require recipes-extended/xen/xen.inc
+require xen-common.inc
+
+inherit findlib
+
+DESCRIPTION = "Xen hypervisor ocaml libs and xenstore components"
+
+# OpenXT packages both the C and OCaml versions of XenStored.
+# This recipe packages the OCaml daemon; xen.bb packages the C one.
+FILES_xen-xenstored-ocaml = " \
+    ${sbindir}/xenstored.xen-xenstored-ocaml \
+    ${localstatedir}/lib/xenstored \
+    ${sysconfdir}/init.d/xenstored.xen-xenstored-ocaml \
+    ${sysconfdir}/xen/oxenstored.conf \
+    "
+
+PROVIDES =+ "xen-xenstored xen-xenstored-ocaml"
+RPROVIDES_xen-xenstored-ocaml = "xen-xenstored xen-xenstored-ocaml"
+
+DEPENDS += " \
+    util-linux \
+    xen \
+    xen-blktap \
+    libnl \
+    ocaml-cross \
+    "
+EXTRA_OECONF_remove = "--disable-ocamltools"
+
+SRC_URI_append = " \
+    file://xenstored.initscript \
+    file://oxenstored.conf \
+    "
+
+PACKAGES = " \
+    xen-xenstored-ocaml \
+    ${PN}-dev \
+    ${PN}-dbg \
+    ${PN}-staticdev \
+    ${PN} \
+    "
+
+FILES_${PN}-dev = "${ocamllibdir}/*/*.so"
+FILES_${PN}-dbg += "${ocamllibdir}/*/.debug/*"
+FILES_${PN}-staticdev = "${ocamllibdir}/*/*.a"
+FILES_${PN} = "${ocamllibdir}/*"
+
+EXTRA_OEMAKE += "CROSS_SYS_ROOT=${STAGING_DIR_HOST} CROSS_COMPILE=${HOST_PREFIX}"
+EXTRA_OEMAKE += "CONFIG_IOEMU=n"
+EXTRA_OEMAKE += "DESTDIR=${D}"
+
+#Make sure we disable all compiler optimizations to avoid a nasty segfault in the 
+#reboot case.
+BUILD_LDFLAGS += " -Wl,-O0 -O0"
+BUILDSDK_LDFLAGS += " -Wl,-O0 -O0"
+TARGET_LDFLAGS += " -Wl,-O0 -O0"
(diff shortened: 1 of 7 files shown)


jean-edouard:
This PR cuts the dependency on xen-libxl for the stubdomain, uivm and ndvm rootfses.
It mitigates the mess caused by meta-virtualization:
- libxl is built 1 time instead of 4
- the xen ocaml libs are built 2 times instead of 4 (previously part of xen-libxl.bb for some reason)
- xen-blktap is built 3 times instead of 4

The situation is still not ideal, and this PR does *not* fix the ticket, but it's a good workaround, especially for libxl.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/785

jean-edouard:
That build fails to boot because xenstored can't create /var/lib/xenstored/tdb and fails to start...
Missing tmpfs?
Please do not merge until this is figured out.

recipes-extended/xen/xen-ocaml-libs.bb
```diff
@@ -0,0 +1,110 @@
...
+EXTRA_OEMAKE += "CONFIG_IOEMU=n"
+EXTRA_OEMAKE += "DESTDIR=${D}"
+
+#Make sure we disable all compiler optimizations to avoid a nasty segfault in the 
+#reboot case.
+BUILD_LDFLAGS += " -Wl,-O0 -O0"
+BUILDSDK_LDFLAGS += " -Wl,-O0 -O0"
+TARGET_LDFLAGS += " -Wl,-O0 -O0"
+BUILD_OPTIMIZATION = "-pipe"
+FULL_OPTIMIZATION = "-pipe ${DEBUG_FLAGS}"
```
dozylynx: The above lines -- starting at "Make sure we disable all compiler optimizations to avoid a nasty segfault in the reboot case." -- originated in the libxl recipe from the libxl work. If they are not required for the OCaml build, it would be good to remove them from this recipe.

recipes-openxt/xenclient/xenclient-toolstack_git.bb
```diff
@@ -1,8 +1,9 @@
 DESCRIPTION = "XenClient toolstack"
 LICENSE = "LGPLv2.1"
 LIC_FILES_CHKSUM = "file://COPYING;md5=321bf41f280cf805086dd5a720b37785"
-DEPENDS += "ocaml-cross ocaml-dbus ocaml-camomile xen xen-libxl xz"
+DEPENDS += "ocaml-cross ocaml-dbus ocaml-camomile xen xz"
 RDEPENDS_${PN} = "xen-xenstore xen-xenstored"
+RDEPENDS_${PN}_xenclient-dom0 += "xen-xl"
```
dozylynx: Switching the toolstack from a build-time to a run-time dependency on xl is a significant change worth mentioning in the commit message.

recipes-extended/xen/xen-ocaml-libs.bb
```diff
@@ -0,0 +1,110 @@
...
+pkg_prerm_xen-xenstored-ocaml () {
+    update-alternatives --remove xenstored xenstored.xen-xenstored-ocaml
+    update-alternatives --remove xenstored-initscript xenstored.xen-xenstored-ocaml
+}
+
+do_configure_prepend() {
+	#remove optimizations in the config files
+	sed -i 's/-O2//g' ${WORKDIR}/xen-${XEN_VERSION}/Config.mk
+	sed -i 's/-O2//g' ${WORKDIR}/xen-${XEN_VERSION}/config/StdGNU.mk
+}
```
dozylynx: The `do_configure_prepend` function again originated in the libxl recipe and should be removable from this recipe.

recipes-extended/xen/xen-ocaml-libs.bb
jean-edouard: Thanks, will do.

recipes-extended/xen/xen-ocaml-libs.bb
jean-edouard: Thanks, will do.

recipes-openxt/xenclient/xenclient-toolstack_git.bb
eric-ch: Apparently there is no longer a dependency from xenclient-toolstack to xen-xl, I can only see RDEPENDS on xen-xenstore from the scripts.
Why is there an RDEPENDS on xen-xl ?

eric-ch:
Build [803](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/803)

recipes-openxt/xenclient/xenclient-toolstack_git.bb
jean-edouard: Dunno, xl was in DEPENDS but didn't seem to be required to build, so I downgraded it to RDEPENDS, but I can completely nuke it if desired.

jean-edouard: Done

jean-edouard:
All comments addressed.
@eric-ch, @dozylynx, good to go?
@dpsmith, want this in stable-7?

dozylynx:
The motivation for this PR changes somewhat in light of the effect of PR #618, but there are a number of useful things in here that make it worth continuing with.

* The correction to list xen-xenstore-c in the dom0 packagegroup is important.

* Moving the OCaml libraries to their own recipe is sensible, even though it increases the number of Xen recipes, because it enables separate expression of build-time and run-time dependencies. It also makes it clearer that OCaml doesn't require the compile optimization changes that are in the libxl recipe.

* Removing libxl as a xenclient-toolstack dependency is good.

So yes, looks ok.

dpsmith:
Yes I believe this should go into stable-7 as well.

PR #614, opened 2017-04-27
[libxl] Fix segfault caused by strstr on domain

crogers1:
shutdown. Just check and make sure tmp != NULL before calling
  strstr.

  OXT-1086


jean-edouard:
The only real diff I see is an empty line, am I missing something?

crogers1:
It just moved the strstr after the block that checks if(!tmp)

jean-edouard:
Oh I see! Interesting, first time I see a commit that fixes a patch by just modifying context lines :)
Will test.

jean-edouard:
Good catch, thanks, merging soon.

PR #619, opened 2017-05-02
OXT-826 : XSM: Add distinct device use permissions with and without IOMMU

dozylynx:
[ PR for initial @dpsmith review ]

 * Enable policy to optionally express that device access is only allowed when an IOMMU is active.
 * Enable policy to require an IOMMU capable of interrupt remapping before granting device access that depends on an IOMMU.

### DIFF
--- a/recipes-extended/xen/files/xsm-for-vtd.patch
+++ b/recipes-extended/xen/files/xsm-for-vtd.patch
@@ -217,7 +217,7 @@ diff --git a/xen/xsm/flask/policy/access_vectors b/xen/xsm/flask/policy/access_v
 index b06cf16..92f4a0f 100644
 --- a/xen/xsm/flask/policy/access_vectors
 +++ b/xen/xsm/flask/policy/access_vectors
-@@ -430,6 +430,12 @@ class resource
+@@ -430,6 +430,11 @@ class resource
  #  target = resource's security label
  # also checked when using some core Xen devices (target xen_t)
      use
@@ -225,12 +225,11 @@ index b06cf16..92f4a0f 100644
 +# checked when adding a resource to a domain:
 +#  source = domain which will have access to the resource
 +#  target = resource's security label
-+# also checked when using some core Xen devices (target xen_t)
 +    use_with_iommu
  # PHYSDEVOP_map_pirq and ioapic writes for dom0, when acting on real IRQs
  #  For GSI interrupts, the IRQ's label is indexed by the IRQ number
  #  For MSI interrupts, the label of the PCI device is used
-@@ -462,6 +468,15 @@ class resource
+@@ -462,6 +467,15 @@ class resource
      setup
  }
  
--- a/recipes-extended/xen/files/xsm-for-vtd.patch
+++ b/recipes-extended/xen/files/xsm-for-vtd.patch
@@ -1,0 +1,259 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Enable XSM control over device passthrough with separate policy controls for:
+* with and without an IOMMU available
+* an IOMMU with more capable isolation, in this case via interrupt remapping
+
+################################################################################
+LONG DESCRIPTION:
+################################################################################
+
+################################################################################
+CHANGELOG
+################################################################################
+
+################################################################################
+REMOVAL
+################################################################################
+Via upstreaming.
+
+################################################################################
+UPSTREAM PLAN
+################################################################################
+Yes.
+
+################################################################################
+INTERNAL DEPENDENCIES
+################################################################################
+OpenXT's XSM policy in xsm-policy.git
+
+################################################################################
+PATCHES
+################################################################################
+diff --git a/xen/xsm/flask/hooks.c b/xen/xsm/flask/hooks.c
+index 59905e2..bac29ad 100644
+--- a/xen/xsm/flask/hooks.c
++++ b/xen/xsm/flask/hooks.c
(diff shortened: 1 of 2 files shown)


recipes-extended/xen/files/xsm-for-vtd.patch
```diff
@@ -0,0 +1,259 @@
...
++
++    /* Use the noaudit method for checking this permission because
++     * we're identifying which _other_ permission to check for.
++     * Failure here is sometimes expected and always acceptable so
++     * it is important not to log it to protect the logs.
++     */
++    if (iommu_enabled && (iommu_intremap ||
++                          !avc_has_perm_noaudit(csid, SECINITSID_XEN,
++                                                SECCLASS_IOMMU,
++                                                IOMMU__USE_INSECURE, &avd)))
```
stephensmalley: Does the iommu use_insecure permission check make sense as a per-domain permission?  Would we ever allow it for domain D1 but deny it for domain D2?

dgdegraaf: Assuming you want the permission per-domain, I think it would be clearer to just have three distinct use_with_* bits rather than using a "hidden" iommu__use_insecure bit:
- resource__use (possibly rename to resource__use_noiommu) - checked if (!iommu_enabled)
- resource__use_iommu_nointremap - checked if (iommu_enabled && !iommu_intremap)
- resource__use_iommu - checked if (iommu_enabled && iommu_intremap)

I also think that if resource__use is mapped to one of these, it should be mapped to resource__use_iommu (i.e. the only fully secure one), with the other two being "enable this to allow less-secure systems to still work" bits.

If this doesn't need to be a per-domain policy, the existing Xen "iommu=" command line argument can enforce the setting; systems that can't use intremap would already need to disable it in that parameter.

recipes-extended/xen/files/xsm-for-vtd.patch
```diff
@@ -0,0 +1,259 @@
...
++++ b/xen/xsm/flask/policy/access_vectors
+@@ -430,6 +430,12 @@ class resource
+ #  target = resource's security label
+ # also checked when using some core Xen devices (target xen_t)
+     use
++# IOMMU required before a resource will be added to a domain:
++# checked when adding a resource to a domain:
++#  source = domain which will have access to the resource
++#  target = resource's security label
++# also checked when using some core Xen devices (target xen_t)
```
dgdegraaf: This line (copied from use) should be removed. It is only true for flask_resource_use_core, which is a part of the page offline sysctl and (correctly) only checks the "use" bit.

recipes-extended/xen/files/xsm-for-vtd.patch
dozylynx: Stephen, Daniel, thanks very much for taking the time to review this and for your comments.

### Re: per-domain permissions:

> Does the iommu use_insecure permission check make sense as a per-domain permission?
> Would we ever allow it for domain D1 but deny it for domain D2?

It's a good question and I think possibly, or possibly not.
Here is a somewhat contrived example but perhaps:

D1: contains the graphics display device driver and a local display device is assigned.
Exposed to internal graphical output only.

IOMMU containment ensures that a buggy device driver is less able to impact system stability, or usability, or the integrity of other VMs.

D2: contains a NIC device driver and an external network interface is assigned.
Exposed to external network traffic.

IOMMU containment ensures that a compromised device driver domain cannot be used by an attacker to obtain access to other data on the system.

(ie. the less capable IOMMU is deemed acceptable for D1 but not D2)

The same host software (and same XSM policy) can be used on multiple machines with different IOMMU capabilities, with the behaviour on the hosts with weaker IOMMUs being that VMs are operable with working display output, but they are isolated from the external network by disallowing the NIC device assignment.

### Re: design change to use three distinct use_with_* bits :
I will do that. I'll prepare an alternative PR with that method for comparison with this one.

With this version of the code, three points that I like -- and will hopefully keep in the other version -- are:

1) The "allow use of the insecure IOMMU" capability is confined to only the domains expressly indicated in the policy, and all others denied, which removes a capability from domains that have no need of it.

2) The XSM policy alone, rather than a combination of XSM policy and the host command line, can set the isolation requirements.

Setting the IOMMU capability requirement via the boot command line makes the configuration setting more distant from the VM XSM policy which depends upon it, in a different format for a different tool.

In OpenXT, we want to support:

* Producing a single build of software that runs on diverse hardware -- so default to not requiring specific hardware capabilities.

* Allow derivative projects to produce specialized builds, to _require_ that specific hardware-enabled security features are always present and enabled, and accept a corresponding reduction in the hardware compatibility.

If separate builds are being produced to use with different hardware, swapping a XSM policy seems simpler to manage than rewriting the hypervisor command lines in the bootloader config files and then ensuring that changed settings are preserved correctly across software upgrades. It's still a change to a boot-time file though, so maybe it's not all that different in practice.

3) The XSM policy and iommu control variables are tested every time the device assignment action is performed, and not only relying on a one-time variable check at host boot.

Hope that explains some of the design, and I'll get a revised version prepared next. Thanks!

recipes-extended/xen/files/xsm-for-vtd.patch
dozylynx: ack - thanks for that catch, will fix it.

dozylynx:
The newer alternative version of this PR that follows the recommendations on the comments here is: #626 .

dozylynx:
Retiring this PR and replacing it with #626.

PR #621, opened 2017-05-02
[xl] On reboot, have xl reparse the config

crogers1:
Move the jump label for the domain reboot case to also include
  the config parse code, since xenmgr just writes a new config
  to the same filename.  This fixes the bug where config value
  changes such as changing the memory amount aren't taking effect
  on reboot

  OXT-1090


rossphilipson:
Built here:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/829
Testing now

rossphilipson:
Tested by adding memory while a VM was running and rebooting. The new setting was loaded and the guest did have the added memory after reboot. Merging soon.

PR #622, opened 2017-05-02
STABLE-7: [xl] On reboot, have xl reparse the config

crogers1:
Move the jump label for the domain reboot case to also include
  the config parse code, since xenmgr just writes a new config
  to the same filename.  This fixes the bug where config value
  changes such as changing the memory amount aren't taking effect
  on reboot

  OXT-1090


rossphilipson:
This one looks like it needs a rebase, can't merge.

crogers1:
Github is showing me the "This branch has no conflicts with the base branch" message.  You sure there are merge conflicts?

rossphilipson:
Hugh and now it merges. Before there was a screen that said cannot merge. Maybe github burped...

PR #624, opened 2017-05-04
  Support swapping isos in virtual cd drive

crogers1:
while guest is on. Make sure we set readonly flags for blockdev
  in qemu when set to 'preserved'. Attempting to open xvdc fails
  because open flags have writable set.

  OXT-1068

### DIFF
--- a/recipes-extended/xen/files/libxl-iso-hotswap.patch
+++ b/recipes-extended/xen/files/libxl-iso-hotswap.patch
@@ -1,0 +1,352 @@
+commit d04740f8f1dfc5c2149863aa3f896ac4664b5c88
+Author: Chris Rogers <rogersc@ainfosec.com>
+Date:   Tue Apr 18 14:43:21 2017 -0400
+
+    implement iso hotswap
+
+Index: xen-4.6.4/tools/libxl/libxl.c
+===================================================================
+--- xen-4.6.4.orig/tools/libxl/libxl.c
++++ xen-4.6.4/tools/libxl/libxl.c
+@@ -2940,6 +2940,126 @@ int libxl_device_disk_getinfo(libxl_ctx
+     return rc;
+ }
+ 
++int libxl_cdrom_change(libxl_ctx *ctx, uint32_t domid, char *iso, libxl_device_disk *olddisk, char *vdev,
++                       const libxl_asyncop_how *ao_how)
++{
++    AO_CREATE(ctx, domid, ao_how);
++    int rc = -1;
++    libxl__device device;
++    libxl_device_disk disk_empty;
++    xs_transaction_t t = XBT_NULL;
++    char *devpath = NULL;
++    char *be_path = NULL;
++    char *fe_path = NULL;
++    char *tmp = NULL;
++    char *be_state = NULL;
++    char *fe_state = NULL;
++    uint32_t stubdomid = libxl_get_stubdom_id(ctx, domid);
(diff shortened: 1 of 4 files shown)


rossphilipson:
Built here:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/829
Testing now

rossphilipson:
ISO appears in running guest when assigned in the UIVM. It is still present after reboot. Looks like it does as advertised. Merging soon.

rossphilipson:
Actually I have a couple of questions on the code first. Will comment in code.

recipes-extended/xen/files/libxl-iso-hotswap.patch
```diff
@@ -0,0 +1,346 @@
...
++    /* insert empty cdrom */
++    libxl__qmp_insert_cdrom(gc, domid, &disk_empty);
++
++    be_path = libxl__device_backend_path(gc, &device);
++    fe_path = libxl__device_frontend_path(gc, &device);
++
++    /* disconnect iso for now */
++    do {
++        t = xs_transaction_start(ctx->xsh);
++        tmp = libxl__xs_read(gc, t, libxl__sprintf(gc, "%s/tapdisk-params", be_path));
```
rossphilipson: So these return vals from these calls to libxl__xs_read are not cleaned up. I assume it is because the gc gets them?

recipes-extended/xen/files/libxl-iso-hotswap.patch
```diff
@@ -0,0 +1,346 @@
...
+ {
+Index: xen-4.6.4/tools/libxl/libxl_utils.c
+===================================================================
+--- xen-4.6.4.orig/tools/libxl/libxl_utils.c
++++ xen-4.6.4/tools/libxl/libxl_utils.c
+@@ -1384,6 +1384,12 @@ int libxl__random_bytes(libxl__gc *gc, u
+     return ret;
+ }
+ 
++int libxl_util_xs_read(libxl_ctx *ctx, char *path, char **out)
```
rossphilipson: I am not sure what the point of this function is?

recipes-extended/xen/files/libxl-iso-hotswap.patch
crogers1: Yes, libxl__xs_read has a call to libxl__add_ptr().  The pointer gets added to the gc->alloc_ptrs array for tracking.  libxl__ptr_add() also will resize the alloc_ptrs if it's full.

recipes-extended/xen/files/libxl-iso-hotswap.patch
crogers1: xl_cmdimpl.c doesn't include libxl_internal.h (where the libxl__xs_<op>() functions are declared) and it can't without significant refactoring due to header import restrictions and conflicts.  xl_cmdimpl.c _does_ include libxl_utils.h, which is not a bad place to implement a simple xs_read call.

However it doesn't look like anyone's cleaning up the pointer, so I'll need to add a free in there.

crogers1:
Pushed a fix to address a handful of leaks in the cd_insert() changes.

rossphilipson:
So is this fixed now - ready to go? Also please update the stable-7 version.

crogers1:
Yes, good to go, and I updated the stable-7 version as well.

PR #625, opened 2017-05-04
  STABLE-7: Support swapping isos in virtual cd drive

crogers1:
while guest is on. Make sure we set readonly flags for blockdev
  in qemu when set to 'preserved'. Attempting to open xvdc fails
  because open flags have writable set.

  OXT-1068


crogers1:
Updated stable-7 PR with changes in https://github.com/OpenXT/xenclient-oe/pull/624

PR #626, opened 2017-05-05
OXT-826  [ alternative ] : XSM: Add distinct device use permissions with and without IOMMU

dozylynx:
This is a revised implementation in response to the feedback from @stephensmalley and @dgdegraaf  on PR #619 . (Also mentioning @dpsmith here to trigger mail notification.) This version has the requested three distinct use_* bits:

* resource__use_iommu
* resource__use_iommu_nointremap
* resource__use_noiommu

==
 * Allow only when an IOMMU is active, or not.
 * Option to require an IOMMU capable of interrupt remapping.

### DIFF
--- a/recipes-extended/xen/files/xsm-for-vtd.patch
+++ b/recipes-extended/xen/files/xsm-for-vtd.patch
@@ -1,0 +1,240 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Enable XSM control over device passthrough with separate policy controls for:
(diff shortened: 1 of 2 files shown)


rossphilipson:
Built here (stable-7):
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/859
Validated stable-7 on (see ticket): HP EliteDesk 800 G1 SFF (HSW)
Code looks LGTM
Will merge soon.

PR #628, opened 2017-05-05
STABLE-7: [libxl] Add hypercall shutdown option.

crogers1:
This option uses xc_shutdown_domain to turn off the guest. Used
  when no pv drivers and acpi interface in qemu is not listening.

  OXT-1091


eric-ch:
To be merged with: https://github.com/OpenXT/manager/pull/91

PR #633, opened 2017-05-15
OXT-1084: tcsd: Disable ipv6

eric-ch:
Dom0 does not have ipv6 support and the SELinux policy does not give
tcsd the permission to try and load the module, which it will try to do.

OXT-1084


rossphilipson:
Tested on stable-7, see ticket. Merging soon.

PR #634, opened 2017-05-15
STABLE-7: OXT-1084: tcsd: Disable ipv6

eric-ch:
Dom0 does not have ipv6 support and the SELinux policy does not give
tcsd the permission to try and load the module, which it will try to do.

OXT-1084


rossphilipson:
Building here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/854

rossphilipson:
Disregard that other build, built here:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/856

rossphilipson:
Tested, see ticket. Merging soon.

PR #637, opened 2017-05-16
STABLE-7: OXT-826 : XSM: Add distinct device use permissions with and without IOMMU

dozylynx:
 * Allow only when an IOMMU is active, or not.
 * Option to require an IOMMU capable of interrupt remapping.


rossphilipson:
Built here:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/859
Validated on: HP EliteDesk 800 G1 SFF (HSW)

PR #641, opened 2017-05-19
OXT 782 QEMU minimization

jandryuk:
This minimizes the set of QEMU emulated devices.
1) configure --disables many options
2) deselects many devices in the device configuration
3) patches out some block format and network functionality.

Size shrank from ~8.1MB to 6.3MB

### DIFF
--- a/recipes-openxt/qemu-dm/qemu-dm-2.6.2/block-remove-unused-block-format-support.patch
+++ b/recipes-openxt/qemu-dm/qemu-dm-2.6.2/block-remove-unused-block-format-support.patch
@@ -1,0 +1,85 @@
+###############################################################################
+SHORT DESCRIPTION:
+###############################################################################
+QEMU remove unused block format support
+
+###############################################################################
+LONG DESCRIPTION:
+###############################################################################
+In an effort to minimize the size of QEMU, we remove unused features.  Here, we
+build-time disable many of the unused block formats.  These cannot be disabled
+in the configure script, so we set them to 'n' in Makefile.objs.
+
+block.c needed modification to remove references to the disabled qcow2
+format.
+
+###############################################################################
+PATCHES:
+###############################################################################
+diff --git a/block.c b/block.c
+index 951f5f4..3bad8ef 100644
+--- a/block.c
++++ b/block.c
+@@ -1405,7 +1405,7 @@ done:
+     qdict_del(options, bdref_key);
+     return c;
+ }
+-
++#if 0 /* OpenXT remove reference to Qcow2 */
+ static int bdrv_append_temp_snapshot(BlockDriverState *bs, int flags,
+                                      QDict *snapshot_options, Error **errp)
+ {
+@@ -1471,7 +1471,7 @@ out:
+     g_free(tmp_filename);
+     return ret;
+ }
+-
++#endif
+ /*
+  * Opens a disk image (raw, qcow2, vmdk, ...)
+  *
+@@ -1702,8 +1702,10 @@ static int bdrv_open_inherit(BlockDriverState **pbs, const char *filename,
+     /* For snapshot=on, create a temporary qcow2 overlay. bs points to the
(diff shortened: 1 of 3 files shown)


recipes-openxt/qemu-dm/qemu-dm-2.6.2/block-remove-unused-block-format-support.patch
```diff
@@ -0,0 +1,79 @@
+From 5e9813ea9c6794307937af2c8bf709b99e065468 Mon Sep 17 00:00:00 2001
```
rossphilipson: These are new patches, please use the standard patch comment block that all the other patches use.

recipes-openxt/qemu-dm/qemu-dm-2.6.2/block-remove-unused-block-format-support.patch
```diff
@@ -0,0 +1,79 @@
...
+diff --git a/block.c b/block.c
+index 951f5f4..3bad8ef 100644
+--- a/block.c
++++ b/block.c
+@@ -1405,7 +1405,7 @@ done:
+     qdict_del(options, bdref_key);
+     return c;
+ }
+-
++#if 0
```
rossphilipson: I assume these had to be ifdef'ed out due to something you turned off in the Makefile?

rossphilipson:
Other than the missing patch comments and a minor question this LGTM. @jean-edouard looked it over too and said it looked fine. Once you fix the patch comments, please do a stable-7 PR also.

recipes-openxt/qemu-dm/qemu-dm-2.6.2/block-remove-unused-block-format-support.patch
jandryuk: The patch description was supposed to explain this: "block.c need to be modified to remove references to the disabled qcow2 format."  I can add a comment at the ifdef location.

We don't use qemu for disk snapshots, right?

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/876

recipes-openxt/qemu-dm/qemu-dm-2.6.2/block-remove-unused-block-format-support.patch
rossphilipson: Ack on comment. It is fine to leave it in the patch comments. I just did not notice it there. We use blktap for snapshots so removing that should be fine.

jandryuk:
I have a small build improvement patch.  Instead of modifying the default configs, it generates and then modifies the dynamic target-specific config.  Should I add it to this PR, or create a new one?

rossphilipson:
@jandryuk it sounds related so I would say add it to this one.

rossphilipson:
Building here:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/882

recipes-openxt/qemu-dm/qemu-dm.inc
```diff
@@ -115,14 +115,15 @@ CFLAGS_append = " -Wno-unused-parameter -Wno-unused-local-typedefs --param ssp-b
...
+do_compile_prepend() {
+	rm -f i386-softmmu/config-devices.mak
+	oe_runmake i386-softmmu/config-devices.mak
 
 	dev_disable() {
-		sed -i -e 's/\('$1'=\).*$/\1n/' default-configs/pci.mak \
-						default-configs/usb.mak \
-						default-configs/sound.mak \
-						default-configs/i386-softmmu.mak
+		sed -i -e 's/\('$1'=\).*$/\1n/' i386-softmmu/config-devices.mak
```
rossphilipson: I like that approach much better. Good call.

rossphilipson:
Just tested, LGTM, no errors, VMs start.

@jandryuk we need a stable-7 cherry pick for this.

PR #643, opened 2017-05-19
STABLE-7 : OXT-1104: SELinux policy: remove storage permissions from xenstored

dozylynx:
Upstream refpolicy has already removed these permissions from xenstored.


rossphilipson:
These SELinux PRs look good to us, merging after RM sign off.

PR #644, opened 2017-05-23
SELinux: add coredump support

jean-edouard:
Label /var/cores and the files inside it.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/876

rossphilipson:
LGTM merging soon

PR #648, opened 2017-05-24
STABLE-7: SELinux: add coredump support

jean-edouard:
Label /var/cores and the files inside it.


rossphilipson:
LGTM merging with RM sign off

PR #592, opened 2017-03-27
OXT-580: Updated mobile-broadband-provider-info and syslinux to use u…

eric-ch:
…pstream OE.

syslinux was converted into a bbappend file that only has the couple of differences we have so we are mostly dependant on OE now.

mobile-broadband-provider-info was converted into a bbappend.

Initial PR: https://github.com/OpenXT/xenclient-oe/pull/546

PR #650, opened 2017-05-24
STABLE-7: OXT-580: syslinux mobile broadband

eric-ch:
Update mobile-broadband-provider-info.
Upgrade syslinux.


rossphilipson:
@eric-ch should there be a master version of this too?

eric-ch:
@rossphilipson there is: https://github.com/OpenXT/xenclient-oe/pull/592

rossphilipson:
Tested on master, LGTM. Will merge with RM sign off

richpersaud:
Is there a known-good version of pxelinux.0 that is compatible with this stable7 version of syslinux and the syslinux version in stable6?

eric-ch:
@richpersaud pxelinux.0 is built from the syslinux recipe, therefor it is 6.03 (stable-7 & master) or 4.04 (stable-6). So, you ca either PXE boot pxelinux.0 4.04 (packaged with a stable-6 build in the netboot dir) or a pxelinux.0 6.03 (packaged with a stable-7 or master build in the netboot dir).
If the goal is to test PXE installation of stable-6 vs stable-7 from one PXE menu, you should be able to chain-load from your pxe-server pxelinux.0 to the one that is packaged with the tested build.

richpersaud:
Could we add pxelinux.0 to the netboot directory?  Not seeing it there currently.

PR #655, opened 2017-05-26
STABLE-7: OXT-782: QEMU minimization

jandryuk:
This minimizes the set of QEMU emulated devices.

1. configure --disables many options
2. deselects many devices in the device configuration
3. patches out some block format and network functionality.

Size shrank from ~8.1MB to 6.3MB


rossphilipson:
Merging with RM sign off.

PR #660, opened 2017-05-31
STABLE-7: OXT-670: Remove old rsyslog recipe

jandryuk:
The old rsyslog 5.6.5 recipe was removed from the stable-6 branch in PR #362. However the master PR #356 did not remove the now-unused recipe. Do so now.


eric-ch:
LGTM, +1. The recipe is not used anyway. As @jandryuk said it is already merged in master.

rossphilipson:
RM signed off, merging

PR #661, opened 2017-05-31
OXT-545: vgabios: Remove unused LFB patch

jandryuk:
Commit 60b473baf1c2f804cce344fd004ef442f0508ff6 (OXT-118) dropped
vbe-hvmloader-lfb-addr.patch from SRC_URI but did not remove the file
from git.  The patch is undocumented and unused, so remove it.


eric-ch:
LGTM, +1. This having no effect on the build: will merge soon.

PR #663, opened 2017-06-01
[tpm2.0] Set the correct flag if detected tpm_own

crogers1:
Fix backwards logic for tpm_is_owned call from installer.

  OXT-1116


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/910

jean-edouard:
Code LGTM, is there an easy way to test this?

crogers1:
Don't clear the TPM before trying to install the build, you should get an error message early on in the installer saying that the TPM is already owned, rather than at the end of the install.

jean-edouard:
LGTM, merging soon.

PR #664, opened 2017-06-01
STABLE-7: [tpm2.0] Set the correct flag if detected tpm_own

crogers1:
Fix backwards logic for tpm_is_owned call from installer.

  OXT-1116


jean-edouard:
@dpsmith, the ticket has the label "7.0.0_RC2_Priority", does that mean this should be merged only after RC1?

dpsmith:
I went back and looked at the ticket, you are correct, it should probably be held for RC2.

PR #665, opened 2017-06-01
Updated forward seal for ML

dpsmith:
This updates forward seal to support TPM2 use case.

OXT-1004


rossphilipson:
Building here:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/905

recipes-extended/xen/files/tboot-xen-evtlog-support.patch
```diff
@@ -0,0 +1,236 @@
+Index: xen-4.6.4/xen/include/asm-x86/tboot.h
+===================================================================
+--- xen-4.6.4.orig/xen/include/asm-x86/tboot.h
++++ xen-4.6.4/xen/include/asm-x86/tboot.h
+@@ -102,6 +102,9 @@ typedef struct __packed {
+     uint32_t  flags;
+     uint64_t  ap_wake_addr;      /* phys addr of kernel/VMM SIPI vector */
+     uint32_t  ap_wake_trigger;   /* kernel/VMM writes APIC ID to wake AP */
++    /* TPM even log fields: */
```
crogers1: s/even/event/ ?

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
```diff
@@ -128,46 +111,29 @@ hash_extend() {
...
 # return 0 on success
 # return 1 on failure
 pcr17() {
-    is_tpm_2_0
-    local tpm2=$?
+    local root="${1:-/}"
+    local pcr17=''
 
-    if [ "${tpm2}" -eq 0 ]; then
-        pcr17=$(tpm2_listpcrs -g 0xb | grep PCR_17 | cut -f2 -d: | tr -d ' ' | tr -d '\n')
```
crogers1: I don't think you'll find the PCRs exposed in the sysfs for TPM 2.0 right now, so I'm not sure if you want remove the tpm2_listpcrs call here (and the surrounding context).

rossphilipson: Maybe we should do a patch to have the tpm2 driver expose them so there is some consistency.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
```diff
@@ -204,13 +167,7 @@ pcr18() {
         case ${mod_num} in
         0)
             cmdline=$(echo ${cmdline} | sed -e 's/FIXME\s*//')
-            if [ "${tpm2}" -eq 0 ]; then
```
crogers1: Same concern as above in pcr17() for tpm2 stuffs.  Also hash_modules seems to use the new style lcp2_mlehash syntax, but not this func.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
crogers1: There might be support in the 4.11 kernel we could backport, but if not, I'm not sure where to begin adding that.

crogers1: On top of that, tpm2.0 can have multiple pcr banks, so we still have to pick from which bank pcr17 is coming from.  I tried to default to sha256 (with a fallback to sha1) in most of my code, but in the future I can see vendors dropping sha1 support and only going with the longer bit length algs.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
```diff
@@ -384,6 +445,37 @@ generate_policy() {
...
+    local hash_alg="sha1" #default to sha1 for legacy support
+
+    is_tpm_2_0
+    tpm2=$?
+
+    if [ "${tpm2}" -eq 0 ]; then
+        hash_alg=$(get_preferred_alg)
+        [ $? -ne 0 ] && echo "No valid algs available" && return 1
+        #Two possible nv index locations on tpm2.0 that tboot checks
+        local tboot_idx='0x01200001'
```
crogers1: Missing second possible index, 0x01c10131

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
dpsmith: I was going to strip this out because the new approach does not need these functions. I added it back in for forward seal when coming up from a stable-6 box. If you are doing measured launch in a stable-6 environment, then have to be on a tpm1.2. If you are upgrading from a system that has since incorporated then tboot/xen/linux trio patch to export the log, this function is not used.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
dpsmith: I only need one and this relies on the fact that you have already stored it in both locations during generate-policy. I am not worried which index was used by tboot because we put the same policy in both.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
crogers1: Ah I see, alright then.

recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
crogers1: Yep, you're right, forgot I put that in.

dpsmith:
Too much has change to correct for TPM2 approach, closing and will resubmit when resolved

PR #668, opened 2017-06-01
STABLE-7: OXT-1125: blktap: Backport XSA-155

eric-ch:
http://xenbits.xen.org/xsa/advisory-155.html
While the rest of the XSA is upstream or patched in the kernel
(patches from QSB-23), the blktap part is missing in 4.6.5 upstream and
the OpenXT patch-queue.


jean-edouard:
Merging soon.

PR #670, opened 2017-06-06
STABLE-7: [tpm2.0] Only support single bank sealing sha256

crogers1:
for now.

  OXT-1132


rossphilipson:
See the master PR. Will merge with RM sign off.

PR #671, opened 2017-06-06
dbd: fix depends line

dpsmith:
This removes the xenclient-toolstack depend that is no longer needed and
adds xen-ocaml-libs.

OXT-1133

### DIFF
--- a/recipes-openxt/xenclient/dbd_git.bb
+++ b/recipes-openxt/xenclient/dbd_git.bb
@@ -4,7 +4,7 @@ require recipes-devtools/ghc/ghc-pkg.inc
 DESCRIPTION = "XenClient database daemon"
 LICENSE = "GPLv2"
 LIC_FILES_CHKSUM = "file://../COPYING;md5=4641e94ec96f98fabc56ff9cc48be14b"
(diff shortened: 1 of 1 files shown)


dozylynx:
looks right to me.

jean-edouard:
LGTM, merging soon.

rossphilipson:
Signed off on stable 7 one too.

jean-edouard:
This PR broke the build.

dpsmith:
I just used it for the ML build. Did the PR break or was because GH went down last night?

jean-edouard:
I think it breaks the dependency on ocaml-dbus.

jean-edouard:
Oh I know what's wrong, xenclient-toolstack is still the recipe that installs the ocaml standard libs, not xen-ocaml-libs! I'll fix that.

jean-edouard:
Actually that's a problem, since for some reason the libs are located in toolstack.git...
I guess we need yet another recipe.

dpsmith:
Would you mind attaching the build log to OXT-1333 so I can review and see if I can recreate the failure.

rossphilipson:
We should move those libs to somewhere else like xclibs.git. Then toolstack.git can go to the boneyard at some point.

jean-edouard:
https://github.com/OpenXT/xenclient-oe/pull/673

PR #672, opened 2017-06-06
[STABLE-7] dbd: fix depends line

dpsmith:
This removes the xenclient-toolstack depend that is no longer needed and
adds xen-ocaml-libs.

OXT-1133


rossphilipson:
Alternate RM sign off: philipsonr@ainfosec.com

PR #676, opened 2017-06-07
Silence warnings during installation.

eric-ch:
- ```xen-dom0-init```, run at startup, will log a warning if xl.conf is not present (xl binary seems to as well), so ship a default one.
- ```sshd``` configuration currently used contains deprecated options.


rossphilipson:
Tested on 942 (master custom), LGTM, merging soon.

rossphilipson:
This one is missing a stable 7 cherry pick.

PR #677, opened 2017-06-07
Allow surfman to coredump

jean-edouard:


eric-ch:
See https://openxt.atlassian.net/browse/OXT-1178, for further development.

PR #678, opened 2017-06-07
dbd: re-add xenclient-toolstack dependency

jean-edouard:
OXT-1133


eric-ch:
+1 LGTM, discussed on the community call.
Required to fixed the build, so merged quickly.

PR #680, opened 2017-06-12
OXT-1130: Password Rework

jandryuk:
These changes support OXT-1130's installer password querying re-work.  Some changes add flexibility for validating the recovery password.  Other changes support the TPM owner password checking.


eric-ch:
Building: [934](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/934)

jandryuk:
Sorry for not mentioning this earlier, but this patchset without the installer patchset breaks password prompting.  The two patchsets need to go in together.

recipes-openxt/xenclient/xenclient-tpm-scripts/tpm-functions
```diff
@@ -193,6 +193,34 @@ tpm_is_owner_wks() {
...
+tpm_check_owner_password() {
+    local passwd="$1"
+    local ret=0
+    local msg=""
+
+    is_tpm_2_0
+    local tpm2=$?
+    if [ "${tpm2}" -eq 0 ];
+    then
+        return $ret
```
jandryuk: We aren't actually checking the owner password, so we need to return failure.

eric-ch:
Depends on: https://github.com/OpenXT/installer/pull/38

recipes-openxt/xenclient/xenclient-tpm-scripts/tpm-functions
eric-ch: Indeed, I will re-trigger a build once that is attended.

recipes-openxt/openxt/openxt-keymanagement/key-functions
```diff
@@ -454,54 +446,88 @@ set_encrypted_key() {
...
 
     return 1
 }
 
+recovery_unlock_key() {
+    local part="${1}"
+    local name="${2}"
+    local boot="${3}"
+    local key="${4}"
+    local extra="${5:-}"
```
jean-edouard: Does this mean "set extra to $5, unless $5 is empty, in which case set extra to empty"?
That wouldn't be any different than `local extra="${5}"` would it?

jandryuk: Yes, you are correct.

The ${5:-} is a pattern from when I write (bash) scripts with `set -u`.  That flag generates errors when accessing unbound variables, and numbered parameters beyond the number passed in are just such a case.  The default value provided by ${5:-} bypasses the error.

jean-edouard: Interesting, I knew there had to be something behind it :)

jandryuk:
For TPM2, tpm_check_owner_password is hard coded to fail.  I'm still investigating how to support re-using a TPM2.

jandryuk:
This builds on the previous patchset.  In addition, it reworks and extends the keymanagement API to provide features needed by the installer modifications.

While doing that, I rebased and reordered commits to squashed things and provide a sensible ordering.  However, that means GitHub displays the commit order incorrectly (being based off dates) instead of the actual commit history ordering.

TPM2 ownership checking is still a work in progress... like the overall TPM2 support.  I wanted to get the bulk of this out there so the patch queue doesn't grow larger.

https://github.com/OpenXT/installer/pull/38 Depends on this.

jandryuk:
This depends on https://github.com/OpenXT/xenclient-oe/pull/685, so I included @crogers1's patch at the beginning of the series.

https://github.com/OpenXT/installer/pull/38 depends on this.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1012

jean-edouard:
Merging tomorrow.

jean-edouard:
#685 was just merged, please remove from this PR.

jandryuk:
This latest patchset adds https://github.com/OpenXT/xenclient-oe/pull/680/commits/c61de466437c1a84fdc54803563d9684157c2d13 to properly handle the tboot policy in NVRAM when re-using an already-owned TPM.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1027

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1034

jean-edouard:
Merging soon.

jean-edouard:
Is this PR expected to break TPM2.0 reseals? ...

jandryuk:
What do you mean by reseal?  OTA forward sealing, Non-OTA upgrade, or something else?

This doesn't really touch TPM2 code, so it shouldn't affect them in particular.

jean-edouard:
I mean when you have a sealed system and you break the measurement so you have to reseal.

jandryuk:
I don't think I broke the reseal.  Are you seeing breakage?  A quick test here shows recovery reseal working, but it's my build with some pending TPM2 changes.

For these changesets, I refactored the password collection for unlocking the config partition, but that is separate from the re-sealing which I didn't touch.

jean-edouard:
Yeah I know it's weird, that's why I didn't test that right away...
I'll start a new build with just this PR and the installer one to make sure.

jean-edouard:
This build contains just the password rework PRs, and TPM2 reseal works as expected: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1036
Something else in build 1034 must be causing the issue, I'll investigate tomorrow...

PR #681, opened 2017-06-14
STABLE-7: OXT-1124: updatemgr: SELinux perms to read ssl certs files.

eric-ch:
Authorize updatemgr to read certificates and configuration required for openssl.


rossphilipson:
Tested on 942 (master custom), LGTM, merging with RM sign off.

PR #688, opened 2017-06-20
init: adjusting for tpm2 extend

dpsmith:
This adjusts the root hash from ascii to binary to align with what the
tpm2_extendpcr utility is expecting.

OXT-1150

### DIFF
--- a/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
+++ b/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
@@ -47,6 +47,11 @@ pcr_bank_exists () {
     return 1
 }
 
+hex2bin() {
+    read hex
+    echo -n $hex | sed 's/\([0-9A-F]\{2\}\)/\\\\\\x\1/gI' | xargs printf
+}
+
 early_setup() {
     mkdir -p /proc /sys /mnt /tmp
     mount -t proc proc /proc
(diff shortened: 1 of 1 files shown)


crogers1:
+1, good catch.

jean-edouard:
@dpsmith, I played around with hex2bin, and it seems that 0x00 bytes get dropped, no?

dpsmith:
I am not certain what you mean, but here is the result of running in dash with leading and trailing 0x00 and they are not dropped (if that is what you meant).
```shell
$ echo -n "0002a3f21d"|hex2bin|od -tx1
0000000 00 02 a3 f2 1d
0000005
$ echo -n "0002a3f21d00"|hex2bin|od -tx1
0000000 00 02 a3 f2 1d 00
0000006
```

jean-edouard:
I did something like:
```
$ DIGEST=$(echo -n "0002a3f21d00"|hex2bin)
$ echo -n "$DIGEST" > /tmp/t
$ od -tx1 /tmp/t
0000000 02 a3 f2 1d
0000004
```

dpsmith:
I did several tests and determined that hex2bin is not dropping the NULLs, the shell is when storing into a variable. I am trying to think if there is any other way to programmatically pass a binary string in on the command line and preserve leading and trailing NULLs. Will do some googling as well to see if I can find an answer. If not, just raises credence that the utility itself needs to be fixed.

dpsmith:
As I was digging, I was reminded that it doesn't matter if I get bash to not drop the nulls. Null characters on the command line will get interpreted as parameter separators. So this fix would only work as long as the hash does not contain nulls.

PR #690, opened 2017-06-20
[OXT-1154] Fix buffer overflows in libxl patches

dozylynx:
OpenXT's libxl-domain-state.patch and libxl-fix-reboot.patch both
contained incorrectly sized arrays on the stack.

The arrays are used for manipulating xenstore strings and were overrun
by string operations. Testing with Xen 4.8 showed junk state in xenstore
and broken guest VM reboot function, revealing these preexisting
defects.

This commit also corrects edge case control flow in libxl-fix-reboot logic.

### DIFF
--- a/recipes-extended/xen/files/libxl-domain-state.patch
+++ b/recipes-extended/xen/files/libxl-domain-state.patch
@@ -120,7 +120,7 @@ Index: xen-4.6.4/tools/libxl/libxl_utils.c
  
 +int libxl_update_state_direct(libxl_ctx *ctx, libxl_uuid xl_uuid, const char * state)
 +{
-+    char path[48];
++    char path[sizeof("/state/00000000-0000-0000-0000-000000000000/state")];
 +    char uuid[37];
 +
 +    uuid_unparse(xl_uuid.uuid, uuid);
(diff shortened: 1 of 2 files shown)


eric-ch:
Building [958](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/958)

jean-edouard:
Very good catch! There's a very good chance that this also caused the xenstored crashes!
@crogers1, please review, will merge soon after.

crogers1:
+1.  Miscounted while declaring that path size, thanks for the catch.

jean-edouard:
@crogers1 just to make sure, you also +1 the state logic fix, correct?

crogers1:
Yea syntactically it's better, and it better handles the case where the "reboot" entry in the xenstore has garbage in it.  

Functionally though, the toolstack never writes anything except "1" into that node.  Clearing that flag does an xs_rm (it doesn't write "0").  libxl_read_reboot() was mostly meant to see if the key existed/not existed.  The edge case that's being fixed here, where the reboot node exists but has garbage in it, causing the state to never set "shutdowning", would only occur if the user willingly wrote junk into that node.

I'm still +1 for the change, I'm just documenting the behavior of the surrounding code.

jean-edouard:
Cool thanks, merging soon.

PR #691, opened 2017-06-20
STABLE 7 : [OXT-1154] Fix buffer overflows in libxl patches

dozylynx:
OpenXT's libxl-domain-state.patch and libxl-fix-reboot.patch both
contained incorrectly sized arrays on the stack.

The arrays are used for manipulating xenstore strings and were overrun
by string operations. Testing with Xen 4.8 showed junk state in xenstore
and broken guest VM reboot function, revealing these preexisting
defects.

This commit also corrects edge case control flow in libxl-fix-reboot logic.

### DIFF
--- a/recipes-extended/xen/files/libxl-domain-state.patch
+++ b/recipes-extended/xen/files/libxl-domain-state.patch
@@ -120,7 +120,7 @@ Index: xen-4.6.4/tools/libxl/libxl_utils.c
  
 +int libxl_update_state_direct(libxl_ctx *ctx, libxl_uuid xl_uuid, const char * state)
 +{
-+    char path[48];
++    char path[sizeof("/state/00000000-0000-0000-0000-000000000000/state")];
 +    char uuid[37];
 +
 +    uuid_unparse(xl_uuid.uuid, uuid);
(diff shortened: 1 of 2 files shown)


recipes-extended/xen/files/libxl-domain-state.patch
```diff
@@ -120,7 +120,7 @@ Index: xen-4.6.4/tools/libxl/libxl_utils.c
  
 +int libxl_update_state_direct(libxl_ctx *ctx, libxl_uuid xl_uuid, const char * state)
 +{
-+    char path[48];
++    char path[sizeof("/state/00000000-0000-0000-0000-000000000000/state")];
```
dpsmith: Nothing wrong with this as I would expect that the compiler optimizes out the sizeof. So really just a preference/suggestion, but I tend to lean towards giving example in comment with a defined value that provides a sensible name below the comment. I would also do this for uuid array, but I would bet the uuid implementation already provides a defined value.

jean-edouard: This PR was already tested and merged in master, so I suggest that any cosmetic change to it should be done as a new PR. Will merge soon if that's ok.

dpsmith: Yep, that is why I signed off on it already.

PR #692, opened 2017-06-21
STABLE-7: OXT-1152: xsa: 216,217,218,219,220,221,222,224

eric-ch:
XSAs were released 2017-06-20, see https://xenbits.xen.org/xsa.

XSA-223 & XSA-225 are ARM specific, and were not ported in OpenXT
patch-queue for that reason.


jean-edouard:
Merging soon.

PR #693, opened 2017-06-21
STABLE-6: OXT-1152: xsa: 216,217,218,219,222,224

eric-ch:
XSA-220 and XSA-221 do not affect Xen 4.3.


eric-ch:
Building: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/970

jean-edouard:
Tried the build, got broken guest display and these errors on serial:
```
(XEN) mm.c:630:d6 Could not get page ref for pfn ffffffffffffffff
(XEN) mm.c:630:d6 Could not get page ref for pfn ffffffffffffffff
(XEN) memory.c:350:d0 Could not get page for gpfn f0000
```

eric-ch:
That would be XSA-222, the backport in this PR is probably incorrect as the p2m foreign mapping received some changes since xen 4.3.

PR #694, opened 2017-06-22
[tpm2.0] Fix arg parsing when passing in hash for

crogers1:
forward sealing.

  OXT-1155

### DIFF
--- a/recipes-openxt/tss/tpm2.0-tools/tpm2-fix-forward-seal.patch
+++ b/recipes-openxt/tss/tpm2.0-tools/tpm2-fix-forward-seal.patch
@@ -1,0 +1,146 @@
+Index: tpm2.0-tools/lib/pcr.c
+===================================================================
+--- tpm2.0-tools.orig/lib/pcr.c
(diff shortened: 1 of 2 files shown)


crogers1:
@jean-edouard Hold off on this for a sec, I missed an edge case.

crogers1:
3 issues solved here:

1) Fixed strtok logic so that it advances past the pcr id in the -r param, to read the forward hash.
2) In forward hash case, forgot to set pcrValues[i].t.size, causing the resulting pcrDigest to be calculated improperly.  This can be seen when forward sealing to all 0's, and having unseal fail even when the pcr is all 0's.
3) Removed the ability to pass a forward hash to tpm2_unsealdata, as that's not something we want the platform to be able to perform.

eric-ch:
building: [976](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/976)

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1000
Merging soon.

PR #699, opened 2017-06-26
OXT-1107: lvm2: install the expected initscripts.

eric-ch:
Redefinition of do_install() will not prevent do_install_append() from
the upstream recipe to run and install the version of the initscripts
that requires a /etc/init.d/functions which is not deployed in OpenXT.
Overwrite the files in a do_install_append(), since this is a .bbappend,
its do_install_append() should be after the one in the main recipe.


eric-ch:
I will re-open with other commits related to lvm2 mishandling.

PR #702, opened 2017-06-27
OXT-1164: xen-libxl: Install xl.conf.

eric-ch:
The dummy configuration file is not installed.
Amend that to silence:
Failed to read config file: /etc/xen/xl.conf: No such file or directory

PR #707, opened 2017-06-28
STABLE-7: OXT-1164: xen-libxl: Install xl.conf.

eric-ch:
The dummy configuration file is not installed.
Amend that to silence:
Failed to read config file: /etc/xen/xl.conf: No such file or directory

Origin: https://github.com/OpenXT/xenclient-oe/pull/702


jean-edouard:
Merging soon

PR #698, opened 2017-06-23
STABLE-7: [tpm2.0-tools] Convert from ascii hash to binary

crogers1:
OXT-1150

Signed-off-by: Chris Rogers <rogersc@ainfosec.com>

PR #695, opened 2017-06-22
STABLE-7: [tpm2.0] Fix arg parsing when passing in hash for

crogers1:
forward sealing.

  OXT-1155

Signed-off-by: Chris Rogers <rogersc@ainfosec.com>

PR #684, opened 2017-06-15
TPM2 forward seal support

dpsmith:
This updates the measured-launch forward seal to work with TPM2.

OXT-1004

Signed-off by: dpsmith@apertussolutions.com

PR #709, opened 2017-06-29
STABLE-7: TPM2 forward seal support

dpsmith:
This updates the measured-launch forward seal to work with TPM2.

OXT-1004

Signed-off by: dpsmith@apertussolutions.com


dpsmith:
This is dependent on PRs #695, #698, and installer.git #46.

jean-edouard:
Title fixed, custom build started.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/977

Fails just like the master PRs on my TPM 2.0 machine.

eric-ch:
Fails on my TPM2.0 machine as well (Optiplex 7040).

dpsmith:
Since we now know this is due to hardware deficiency, would it be possible to merge this and open a Jira issue about the failing devices. I will be working to have a solution as quickly as possible.

jean-edouard:
I would really like to see it working on any machine first... I'm not against merging something that doesn't work everywhere, but seeing it work on at least one machine would be really nice.

dpsmith:
If you have a NUC5i5MYHE, you can see it work there. Simply making it work for you machine would break it for every machine I have. Making a proper solution that detects and seals based on which way the calculation is done is a bit more involved and means likely holding this PR and RC1 up until next likely Friday at the earliest.

jean-edouard:
Isn't that NUC known for doing weird stuff with the TPM? Shouldn't we support everything but that one?
Maybe that discussion should be taken to the mailing list or something, but merging PRs without testing them has created huge problems in the past, and I want to make sure that doesn't happen again.
For that reason I am personally against merging this right now, but if people disagree I'll merge anyway.
Right now I'm trying to find an HP machine to test on.

richpersaud:
If testing an HP machine, ensure it is the same Intel generation as the Dell machine.

jean-edouard:
Build 977, that contains this, #695, #698, and https://github.com/OpenXT/installer/pull/46 reboots in a loop after install on my 2 TPM2.0 laptops (HP EliteBook 850 G3 and Dell Latitude 7480)...
They both don't have a serial port.
Same thing happens with equivalent master builds.
I also think I bricked my Dell Precision Tower 3420, which was the 3rd and I think last TPM2.0 machine I have access to, which is the only that actually booted after install.

jean-edouard:
On an interesting note, the reboot-in-a-loop also happens with unmeasured installs.

richpersaud:
Is Build 977 associated with more than one JIRA ticket number?  It will be easier to diagnose regressions if each build is associated with a single JIRA ticket.

dpsmith:
Just rebased from stable-7 & cherry-picked the latest from #684.

eric-ch:
[Build #1002](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1002)

Depends:
- https://github.com/OpenXT/xenclient-oe/pull/709
- https://github.com/OpenXT/xenclient-oe/pull/695
- https://github.com/OpenXT/xenclient-oe/pull/698/
- https://github.com/OpenXT/installer/pull/46

jean-edouard:
Merging tonight.

PR #712, opened 2017-07-06
[libxl] Ignore dead domains in uuid_to_domid

crogers1:
When a domain crashes and for some reason its resources can't
  be released, xl lists it as (null). This is similar to xenvm
  and the deadbeef uuids. Since they still have a domid, xenmgr
  gets hung up on them, preventing that guest from booting again
  until a host reset.

  This patch will ignore the dead domain when xenmgr asks for a
  domid, providing better fault tolerance and allow the platform
  to continue to be usable.

  OXT-1176


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/988

jean-edouard:
Merging soon.

PR #714, opened 2017-07-18
OXT-1182: tboot: Write-back the e820 after reserving evtlog.

eric-ch:
TBoot handles a copy of the E820 until post_launch() completes and
passes control to the next module (launch_kernel()). export_evtlog() has
to be called before replace_e820_map(), which writes the final e820 in
the MBI structures.

### DIFF
--- a/recipes-openxt/tboot/tboot-1.9.5/tboot-export-of-tpm-event-log.patch
+++ b/recipes-openxt/tboot/tboot-1.9.5/tboot-export-of-tpm-event-log.patch
@@ -14,8 +14,6 @@ Signed-off-by: Daniel P. Smith <dpsmith@apertussolutions.com>
  tboot/txt/txt.c         | 128 ++++++++++++++++++++++++++++++++++++++++++++++++
  4 files changed, 138 insertions(+)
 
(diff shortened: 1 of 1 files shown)


jean-edouard:
LGTM

eric-ch:
[Build 1006](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1006)

crogers1:
+1 looks good, export_evtlog() manipulates the e820 map with the extra reservation so yea, the final e820 write should happen after.

dpsmith:
LGTM

jean-edouard:
Merging soon, should PR stable-7 cherry-pick.

jean-edouard:
Actually, since this PR depends on another one being merged, it can't be PR-ed against stable-7 right now.
Also I think it's preferable to keep the existing stable-7 PRs as-is, so that they matche the ones in master, to avoid confusion.
So I suggest that once tboot/ML stuff work fine in master, we merge the open stable-7 PRs, then ask @eric-ch to cherry-pick this one into a new stable-7 PR.
It's not ideal but I don't think there's a better solution that keeps things clean.
Merging now.

PR #715, opened 2017-07-19
STABLE-7: Fix link state on emulated NICs

jean-edouard:
OXT-886


jean-edouard:
Note: I think this PR broke guest netboot in master, since it's propagating invalid NIC states from network-daemon.

PR #716, opened 2017-07-20
STABLE-7: OXT-1182: tboot: Write-back the e820 after reserving evtlog.

eric-ch:
TBoot handles a copy of the E820 until post_launch() completes and
passes control to the next module (launch_kernel()). export_evtlog() has
to be called before replace_e820_map(), which writes the final e820 in
the MBI structures.

(cherry picked from commit 50ddf29bb17ae2191216c7bd56eb7ec25796f4b2)


jean-edouard:
Merging very soon.

PR #730, opened 2017-08-03
[STABLE-7] OXT-1130: Password Rework

jandryuk:
This is the Stable-7 version.

These changes support OXT-1130's installer password querying re-work. Some changes add flexibility for validating the recovery password. Other changes support the TPM1 owner password checking.

This should go in with https://github.com/OpenXT/installer/pull/48


jean-edouard:
Note: this likely breaks non-OTA upgrades from 6.0.1.

dpsmith:
Does it actually break the upgrade or is there just a concern?

jean-edouard:
If you have a measured 6.0 install and try to upgrade to 7 using the installer, not OTA, I think it asks all the passwords, when it shouldn't. Also we don't know the tpm owner password, which is not customizable on 6.
However, few people seem to care about non-OTA upgrades, so this might be an acceptable regression.

jean-edouard:
@jandryuk, did you test non-OTA upgrades?
@dpsmith, do we want support for non-OTA upgrades in stable-7?
Magali, is there a ticket describing how this breaks 6 -> 7 non-OTA upgrades?

MagOXT:
I am working on it.

jandryuk:
For an upgrade, we only prompt for the recovery password.  I previously tested the upgrade path of this code with a "Refresh".   I just re-tested with a stable-6 to stable-7 (non-OTA) upgrade, and it only prompts for the recovery password.

Aside from taking ownership, the TPM password is needed for writing the tboot policy into NVRAM.  If it's already written, then we don't have to do anything.  Otherwise, we'd need the password to write to NVRAM.

For stable-6, no policy is installed in NVRAM, and tboot just uses the default one.  Stable-7 writes a policy in NVRAM, but this cannot be done on upgrade without the TPM password.  However an upgrade still boots into stable-7 after upgrade without the tboot policy.  But this is just waiting to fail a forward seal in the future if the embedded tboot policy changes with a tboot update.

We can use the recovery key to get a stable-6 owner key our of the RSA-encrypted file.  When use that to write to the NVRAM.

A note about reading the NVRAM:
For TPM1, if we are re-installing/upgrading, and we know the TPM NVRAM matches, we are good.  Reading with lcp_readpol doesn't seem to need the password (probably from how permissions are set?) so that is doable.

Stable-6 only supported TPM1, but for TPM2, we must supply the password to read the NVRAM, so that is not readily doable.  TPM2 also needs to ensure the correct set of handles are present.

jean-edouard:
Thanks for the details @jandryuk. Building now to test.

jandryuk:
To be clear, grabbing the owner password from RSA decryption needs to be
implemented for a non-OTA upgrade.

For a stable-6 to stable-7 OTA, I don't have a solution to writing the tpm
nvram. Maybe we'll have to patch the tboot default policy?

If an OTA is ever going to need with the TPM, we'll need to get the TPM
password there somehow.

jean-edouard:
This PR actually appears to play fine with offline upgrade, merging soon.

jean-edouard:
@dpsmith I just noticed this PR doesn't have your approval.

dpsmith:
I was waiting to see if upgrade was actually affected. Will sign off now.

PR #737, opened 2017-08-07
OXT-1172: ucode: Upgrade intel micro-code to 20170707.

eric-ch:


recipes-core/microcode/intel-microcode_20170707.bb
```diff
@@ -45,7 +48,7 @@ do_install() {
 	install -d ${D}${base_libdir}/firmware/intel-ucode/
 	install ${WORKDIR}/microcode_${PV}.bin ${D}${base_libdir}/firmware/intel-ucode/
 	cd ${D}${base_libdir}/firmware/intel-ucode/
-	ln -sf microcode_${PV}.bin microcode_intel.bin
+	ln -sf microcode_${PV}.bin microcode.bin
```
jean-edouard: Won't that change break everything that uses `microcode_intel.bin`? (grub, installer, build scripts...)

eric-ch: No, I believe that was a spurious replacement in the original PR that ends up changing the upstream recipe, but this is not necessary.
The renaming required for OpenXT is and should be done in the ```.bbappend``` file. Since this is installed in ```${base_libdir}/firmware/intel-ucode```, there is no need to qualify it as ```_intel```.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1059

jean-edouard:
LGTM, merging soon.

PR #738, opened 2017-08-07
STABLE-7: OXT-1172: ucode: Upgrade intel micro-code to 20170707.

eric-ch:


jean-edouard:
Merging soon.

PR #742, opened 2017-08-09
STABLE-6: OXT-1172: ucode: Upgrade intel micro-code to 20170707.

eric-ch:


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1076

jean-edouard:
Merging soon.

jandryuk:
intel-microcode was removed from packagegroup-xenclient-dom0 in 65f69179e022b4c9c6cf8f5ddf5b62af71210f8c OXT-706 so this doesn't actually change anything?

eric-ch:
@jandryuk Yes! The microcode was removed as part of a work-around ([OXT-796](https://openxt.atlassian.net/browse/OXT-706)) and is no longer installed. Nice catch! This was cherry-pick'ed from a stable-7/master development for sure.

 I can't remember if 6.0 was still loading the microcode using the Linux patch series or Xen's method through the boot command-line. Maybe this can be backported to Xen 4.3, but in any case this should be re-introduced in 6.1 with the Xen 4.6 upgrade.

PR #743, opened 2017-08-09
STABLE-7: OXT-1189: fix libxl memory corruption in disk crypto dir patch

dozylynx:
Fixes toolstack memory corruption when parsing VM configuration data
and during disk attach.

* Don't omit the string termination character in length calculations.

* Don't ignore the string handling functions already provided.

* Make the existing code that performs a deep struct copy also
  duplicate the new string that was added to the struct.

* Adhere to the formatting conventions of the code being patched.

* Add notes relevant for upstreaming to the patch.


jean-edouard:
Merging soon.

PR #745, opened 2017-08-14
STABLE-7: OXT-1130: Reuse owned TPM2

jandryuk:
Stable-7 version.

This PR provides TPM2 support for re-using an owned TPM2. When a TPM2 is already owned, we take care to verify the Owner, Endorsement, and Lockout passwords. Later, if the needed key object handles or NVRAM indices do not match expectations, new ones are defined using the TPM password.

OXT-1130


jean-edouard:
Merging soon.

PR #759, opened 2017-08-25
STABLE-7: libxl-iso-hotswap: "physical-device" is hex, not dec

jean-edouard:
OXT-1138


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1092

PR #761, opened 2017-08-25
Add patch documentation to evtchn-do-not-set-pending-if-s3.patch

dozylynx:
No code change, documentation in comments only.


jean-edouard:
Thanks, merging soon.

PR #762, opened 2017-08-29
base-passwd: use /bin/false instead of /bin/sh for system users

jean-edouard:
OXT-836


eric-ch:
Build [#1109](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1109)

eric-ch:
That might not always work as intended (packages are installed on target, build timings, etc). There is a [useradd.bbclass](http://www.yoctoproject.org/docs/2.1/ref-manual/ref-manual.html#ref-classes-useradd), and configuring the shell for users is possible through [USERADD_PARAM](http://www.yoctoproject.org/docs/2.1/ref-manual/ref-manual.html#var-USERADD_PARAM) variable.

jean-edouard:
Right, packages that use USERADD_PARAM are responsible for setting the right shell.

eric-ch:
I misunderstood the goal of this. This is only changing the content of `passwd.master` which is fetched by `base-passwd` recipe and already contains the usual user list.

LGTM, will merge soon.

PR #763, opened 2017-08-31
xen: remove patch-patching in libxl-xenmgr-support and libxl-domain-state

dozylynx:
The most recent version of libxl-xenmgr-support.patch included hunks that
revised code introduced in libxl-domain-state.patch. This commit corrects that
by updating libxl-domain-state.patch with the intended final code and removing
the amend hunks from libxl-xenmgr-support.patch.

No code change to the end result of applying the patch queue,
just tidier patches.


eric-ch:
LGTM, building [#1107](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1107)

PR #767, opened 2017-09-13
OXT-1216: xen: backport XSA-226 fix from upstream.

eric-ch:
https://xenbits.xen.org/xsa/advisory-226.html requires a fix that has
been backported in the stable branches upstream[1].

[1] https://xenbits.xen.org/gitweb/?p=xen.git;a=commit;h=9bf14bbf990843bfec16a5d69d36cf46c7593d88

PR #769, opened 2017-09-14
OXT-1216: xen: XSA-226 v7

eric-ch:
Update the XSA patch to include the latest fix.
https://xenbits.xen.org/gitweb/?p=xen.git;a=commit;h=b4660b4d4a35edac715c003c84326de2b0fa4f47

Alternative to https://github.com/OpenXT/xenclient-oe/pull/767. Uses the v7 of the XSA-226 advisory instead of back-porting another patch.


eric-ch:
This PR is no longer required, the patch will eventually get removed, either by upgrading to Xen 4.6.7 or using upstream stable-4.x recipes as proposed on the 09/14/2017 community call.

PR #771, opened 2017-09-20
OXT-1212: Fix TPM2 active status check

jandryuk:
Some BIOSes allow the TPM2 to be "on" but disabled.  In that case we need
to verify actual usability of the TPM2 to proceed with installation.  By returning inactive, the installer will allow the user to downgrade to a non-measured launch installation.

This needs testing on a machine that allows for an "on-but-disabled" TPM2.

OXT-1212


eric-ch:
Building [1150](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1150)

MagOXT:
+1
OK on TPM1.2 and TPM2.0

PR #772, opened 2017-09-20
[libxl] Fix flr for pci passthru devices

crogers1:
  guest boots.

  on guest shutdown.

  OXT-1217

### DIFF
--- a/recipes-extended/xen/files/libxl-fix-flr.patch
+++ b/recipes-extended/xen/files/libxl-fix-flr.patch
@@ -1,0 +1,194 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Fix device reset for pci devices in libxl
+
+################################################################################
+LONG DESCRIPTION:
+################################################################################
+This patch fixes the logic and the control flow path for pci device reset in
+libxl.  Additionally, it addresses an issue with Xenbus state synchronization
+during domain teardown.  More detailed information about the patch follows:
+
+The changes to Xenbus state are to prevent hitting a timeout when domains are 
+shutdown or destroyed. Xl is waiting for "XenbusStateConnected" (state 4), while 
+the state machine has already moved the device to Closing (state 5) and Closed 
+(state 6) by the time we reach this point tearing down the domain.
+
+The reset logic introduced in this patch is delayed until the domain has fully 
+been destroyed. The reason for this has to do with the 
+thorough-reset-interface-to-pciback-s-sysfs.patch in the linux patchqueue. 
+__pcistub_raw_device_reset tries multiple approaches for resetting the device, 
+flr, slot-level, and bus-level. If we encounter a device that, for example, 
+doesn't support flr but does support a slot-level reset, we want to make sure 
+all the functions on that device (GPUS are a good example, they often have 2 
+functions, video and HDMI audio, and are the only device in the slot) are 
+released from the domain before attempting the reset or it will fail. This 
+approach seems to be taken in an attempt to support a wide variety of PCI 
+devices.
+
+################################################################################
+CHANGELOG
+################################################################################
+Authors:
+Chris Rogers <rogersc@ainfosec.com>
+
+################################################################################
+REMOVAL
+################################################################################
+N/A
+
+################################################################################
+UPSTREAM PLAN
+################################################################################
+Depends on the feasibility of upstreaming thorough-reset-interface-to-pciback...
+
+################################################################################
+INTERNAL DEPENDENCIES
+################################################################################
+
+################################################################################
+PATCHES
+################################################################################
+Index: xen-4.9.0/tools/libxl/libxl_internal.h
+===================================================================
+--- xen-4.9.0.orig/tools/libxl/libxl_internal.h
++++ xen-4.9.0/tools/libxl/libxl_internal.h
+@@ -1412,12 +1412,19 @@ _hidden int libxl__pci_topology_init(lib
+ 
+ /* from libxl_pci */
+ 
++typedef struct libxl_pci_dev_wrap {
++    libxl_device_pci *pcidevs;
++    int num_devs;
++} libxl_pci_dev_wrap;
++
+ _hidden int libxl__device_pci_add(libxl__gc *gc, uint32_t domid, libxl_device_pci *pcidev, int starting);
+ _hidden int libxl__create_pci_backend(libxl__gc *gc, uint32_t domid,
+                                       libxl_device_pci *pcidev, int num);
+-_hidden int libxl__device_pci_destroy_all(libxl__gc *gc, uint32_t domid);
++_hidden int libxl__device_pci_destroy_all(libxl__gc *gc, uint32_t domid, libxl_pci_dev_wrap **pciw);
+ _hidden bool libxl__is_igd_vga_passthru(libxl__gc *gc,
+                                         const libxl_domain_config *d_config);
++_hidden int libxl__device_pci_reset(libxl__gc *gc, unsigned int domain, 
++                                    unsigned int bus, unsigned int dev, unsigned int func);
+ 
+ /* from libxl_dtdev */
+ 
+@@ -3609,6 +3616,7 @@ struct libxl__destroy_domid_state {
+     /* private to implementation */
+     libxl__devices_remove_state drs;
+     libxl__ev_child destroyer;
++    libxl_pci_dev_wrap *pciw;
+     bool soft_reset;
+ };
+ 
+Index: xen-4.9.0/tools/libxl/libxl_pci.c
+===================================================================
+--- xen-4.9.0.orig/tools/libxl/libxl_pci.c
++++ xen-4.9.0/tools/libxl/libxl_pci.c
+@@ -212,7 +212,7 @@ static int libxl__device_pci_remove_xens
+         return ERROR_FAIL;
+ 
+     if (domtype == LIBXL_DOMAIN_TYPE_PV) {
+-        if (libxl__wait_for_backend(gc, be_path, GCSPRINTF("%d", XenbusStateConnected)) < 0) {
++        if (libxl__wait_for_backend(gc, be_path, GCSPRINTF("%d", XenbusStateClosed)) < 0) {
+             LOGD(DEBUG, domid, "pci backend at %s is not ready", be_path);
+             return ERROR_FAIL;
+         }
+@@ -235,13 +235,12 @@ static int libxl__device_pci_remove_xens
+ retry_transaction:
+     t = xs_transaction_start(ctx->xsh);
+     xs_write(ctx->xsh, t, GCSPRINTF("%s/state-%d", be_path, i), GCSPRINTF("%d", XenbusStateClosing), 1);
+-    xs_write(ctx->xsh, t, GCSPRINTF("%s/state", be_path), GCSPRINTF("%d", XenbusStateReconfiguring), 1);
+     if (!xs_transaction_end(ctx->xsh, t, 0))
+         if (errno == EAGAIN)
+             goto retry_transaction;
+ 
+     if (domtype == LIBXL_DOMAIN_TYPE_PV) {
+-        if (libxl__wait_for_backend(gc, be_path, GCSPRINTF("%d", XenbusStateConnected)) < 0) {
++        if (libxl__wait_for_backend(gc, be_path, GCSPRINTF("%d", XenbusStateClosed)) < 0) {
+             LOGD(DEBUG, domid, "pci backend at %s is not ready", be_path);
+             return ERROR_FAIL;
+         }
+@@ -1126,13 +1125,13 @@ out:
+     return rc;
+ }
+ 
+-static int libxl__device_pci_reset(libxl__gc *gc, unsigned int domain, unsigned int bus,
++int libxl__device_pci_reset(libxl__gc *gc, unsigned int domain, unsigned int bus,
+                                    unsigned int dev, unsigned int func)
+ {
+     char *reset;
+     int fd, rc;
+ 
+-    reset = GCSPRINTF("%s/do_flr", SYSFS_PCIBACK_DRIVER);
++    reset = GCSPRINTF("%s/reset_device", SYSFS_PCIBACK_DRIVER);
+     fd = open(reset, O_WRONLY);
+     if (fd >= 0) {
+         char *buf = GCSPRINTF(PCI_BDF, domain, bus, dev, func);
+@@ -1484,10 +1483,6 @@ skip1:
+         fclose(f);
+     }
+ out:
+-    /* don't do multiple resets while some functions are still passed through */
+-    if ( (pcidev->vdevfn & 0x7) == 0 ) {
+-        libxl__device_pci_reset(gc, pcidev->domain, pcidev->bus, pcidev->dev, pcidev->func);
+-    }
+ 
+     if (!isstubdom) {
+         rc = xc_deassign_device(ctx->xch, domid, pcidev_encode_bdf(pcidev));
+@@ -1636,7 +1631,7 @@ out:
+     return pcidevs;
+ }
+ 
+-int libxl__device_pci_destroy_all(libxl__gc *gc, uint32_t domid)
++int libxl__device_pci_destroy_all(libxl__gc *gc, uint32_t domid, libxl_pci_dev_wrap **pciw)
+ {
+     libxl_ctx *ctx = libxl__gc_owner(gc);
+     libxl_device_pci *pcidevs;
+@@ -1654,8 +1649,9 @@ int libxl__device_pci_destroy_all(libxl_
+         if (libxl__device_pci_remove_common(gc, domid, pcidevs + i, 1) < 0)
+             rc = ERROR_FAIL;
+     }
+-
+-    free(pcidevs);
++    *pciw = malloc(sizeof(libxl_pci_dev_wrap));
++    (*pciw)->pcidevs = pcidevs;
++    (*pciw)->num_devs = num;
+     return rc;
+ }
+ 
+Index: xen-4.9.0/tools/libxl/libxl_domain.c
+===================================================================
+--- xen-4.9.0.orig/tools/libxl/libxl_domain.c
++++ xen-4.9.0/tools/libxl/libxl_domain.c
+@@ -999,6 +999,15 @@ static void domain_destroy_callback(libx
+         dds->rc = rc;
+     }
+ 
++    if(dis->pciw) {
++        for (int i = 0; i < dis->pciw->num_devs; i++) {
++            libxl_device_pci *pcidev = &dis->pciw->pcidevs[i];
++            libxl__device_pci_reset(gc, pcidev->domain, pcidev->bus, pcidev->dev, pcidev->func);
++        } 
++        free(dis->pciw->pcidevs);
++        free(dis->pciw);
++    }
++
+     dds->domain_finished = 1;
+     destroy_finish_check(egc, dds);
+ }
+@@ -1065,8 +1074,11 @@ void libxl__destroy_domid(libxl__egc *eg
+         goto out;
+     }
+ 
+-    if (libxl__device_pci_destroy_all(gc, domid) < 0)
++    rc = libxl__device_pci_destroy_all(gc, domid, &(dis->pciw));
++    if (rc < 0) {
+         LOGD(ERROR, domid, "Pci shutdown failed");
++    }
++
+     rc = xc_domain_pause(ctx->xch, domid);
+     if (rc < 0) {
+         LOGEVD(ERROR, rc, domid, "xc_domain_pause failed");
--- a/recipes-extended/xen/xen-common.inc
+++ b/recipes-extended/xen/xen-common.inc
@@ -97,6 +97,7 @@ SRC_URI_append = " \
     file://libxl-disable-json-updates.patch \
     file://libxl-allow-non-qdisk-cdrom.patch \
     file://libxl-domain-build-fix-array-size-bug.patch \
+    file://libxl-fix-flr.patch \
     file://xsa226-4.9/0001-gnttab-dont-use-possibly-unbounded-tail-calls.patch \
     file://xsa226-4.9/0002-gnttab-fix-transitive-grant-handling.patch \
     file://xsa226-4.9/0003-gnttab-fix-don-t-use-possibly-unbounded-tail-calls.patch \


eric-ch:
Building [1151](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1151)

crogers1:
This one is fine.  I had to reopen the stable-7 one because the patch was cherry-picked from this commit, which obviously won't work because master is on 4.9 and stable-7 is on 4.6

dozylynx:
@crogers1 : thanks for working on this. The patch looks like it should do the trick.

To the code in the PR here: with an eye to getting these patches suitable for upstream acceptance at some point, is it the case that the need to reset the PCI bus actually _causes_ the need to change the state machine within libxl, or is that a separate behaviour that is being fixed? Are there PCI devices that do not need the bus reset FLR method, but do need it to be invoked later, as per the modified state machine?

If it's not the bus reset that causes the need to change the state machine, then these are separate fixes and these should be two separate patches in the patch queue:
1) one that just changes the sysfs node used from `do_flr` to `reset_device`
2) a second that changes the state machine

Upstreaming 1 has a challenge in that it uses a sysfs node not present in mainline Linux, where as 2 does not.

I like @dpsmith 's suggestion to use local variables in the review.

The patch is missing the standard documentation patch header that all the others in the patch queue have -- please could could that be added with the necessary text.

crogers1:
@dozylynx take a look a my comment on https://github.com/OpenXT/xenclient-oe/pull/774.  As for what's not covered there...

1) I'm not sure who is moving the state machine past XenbusStateConnected underneath libxl, but it's likely been in OpenXT since the stable-5 era (or before).  I'm not sure xenvm had this same synch point in domain teardown.  The fix in this patch addresses the timeout we end up hitting because the Xenbus state is not what libxl expects.  While annoying to wait for the timeout, the device should be reset.  I encountered the bug while implementing the flr fix, so they went into the same patch, if that's wrong I guess we'll have to break them out.

2) The ```__pcistub_raw_device_reset``` function in _thorough-reset-interface-to-pciback-s-sysfs.patch_ seems to be an attempt to handle a wide range of PCI devices by incrementally trying more forceful reset methods, rather than quitting after just an FLR.  I have absolutely no idea how challenging it would be to upstream this to Linux.

3) I can add the patch documentation and use the local vars.

dpsmith:
I think you need to reinspect your commit. You changed the return type on ``libxl__device_pci_destroy_all`` but do not update the invocation for ``libxl__device_pci_destroy_all`` to handle it and then free the devices like you did in the stable-7 resubmitted PR.

crogers1:
@dpsmith thanks for the catch, I must've lost that hunk while I was working with the patch.  Give me a few to update it and make sure this and stable-7 match up in terms of content.

crogers1:
Updated with patch header and style recommendations

eric-ch:
Building [1154](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1154)

jean-edouard:
@dozylynx, please review

dozylynx:
> @dozylynx, please review

ack, will do

dozylynx:
Chris, firstly: thanks for being so responsive on this work.

I'm reviewing this PR with an eye to this going upstream ASAP.

## Positive news

With OpenXT engagement in Xen community discussion last week, there is new effort under way by Oracle to provide device reset support, which will hopefully provide a path to retiring our current kernel patch:

https://lists.xen.org/archives/html/xen-devel/2017-09/msg02968.html

## Review of this PR

1. Minor cosmetic code placement and formatting
2. Function attribute in declaration
3. Concerns

## Minor cosmetic code placement and formatting

In libxl_internal.h:

* Recommend moving the struct definition and function declarations into the section lower down, marked `/* from libxl_pci */`

* Put bus, dev, func arguments all on the same line as they belong together
* and align the whitespace with the open bracket on the line above, ie.:

```
_hidden int libxl__device_pci_reset(libxl__gc *gc, unsigned int domain,
                                    unsigned int bus, unsigned int dev, unsigned int func);
```

Super trivial: there's also a trailing whitespace in the first `libxl_domain.c` hunk...

## attribute in function declaration

* In libxl_internal.h: Should `libxl__device_pci_reset` be declared as: `_hidden` ?

All the other `libxl__device_pci_*` function declarations have that attribute so it looks appropriate.


## Concerns

### Error handling in `libxl_pci.c: libxl__device_pci_destroy_all`

The error code return value is being _read and then dropped_.

Even if the current sources of errors in that code are unlikely, I don't think we should ignore errors and then issue reset to the device anyway.


### Ownership of the device until reset issued

The device reset is being issued after the device has been unassigned with the ownership nodes removed from xenstore. Is there any obstacle to another VM claiming the device before reset is issued?

Does this introduce a race condition where the device is then assigned to a different VM and then reset the device after that VM has started?

ie. this interleaved sequence:

- VM 1: `libxl__device_pci_destroy_all`
- VM 2: _start a new VM with that PCI device assigned to it_
- VM 1: `domain_destroy_callback` => reset device while in use by VM2

If so, that doesn't looks like an acceptable behaviour to introduce upstream.


### Existing code that expects device reset needs to continue to work.

With moving the device reset code out of `do_pci_remove`, the functions `libxl_device_pci_destroy` and `libxl_device_pci_remove` will no longer cause a device reset to occur.

This is incorrect and will break "xl pci-detach".


### Changing the states that are waited for from XenbusStateConnected to XenbusStateClosed, and dropping the XenbusStateReconfiguring cycle.

There two important cases of the device being removed from the domain:
- domain shutdown
- device hot remove, via `xl pci-detach`

Possibly each may need separate handling.

That state machine has been there unchanged since 2009 though, so some caution here is appropriate: can we get a better understanding of what is causing the xenbus nodes to be in the Closing or Closed stated when that code is executed?

Will reset be issued to devices more than once?

## Concluding remarks

We need to understand whether these changes are OpenXT-specific and due to behaviour in our OpenXT-specific parts of the toolstack or not.

If not, this work should go upstream very soon: if it is needed for correct passthrough behaviour then it will definitely have value to other Xen projects and users.

It will benefit from further review within the Xen community.

crogers1:
@dozylynx Thanks for the review, I'll need some time to address your concerns.  I'll try to answer as best I can during the community call this morning.

crogers1:
## Cosmetic Stuff
Moved struct and func declaration into ```/* from libxl_pci */``` section and aligned whitespace/arg groupings.  Added ```_hidden``` modifier.  According to libxl.h, all 'private' libxl functions (functions prefixed by libxl__) may only be called from other libxl functions and must be declared ```_hidden```, so this certainly applies to our device_reset func that I've changed from a static declaration.

## Error Code Handling
Fixed call to ```libxl__device_pci_destroy_all```.  It no longer drops the error code, and it returns the created pci_dev_wrap struct through a pointer.  Fixed up declaration to reflect new arg and return type.

## Breaking Existing Code
libxl actually performs a device reset on pci_add as well (so either at domain boot or when hotplugging).  So the next time it would be used, either hotplugged into a running domain, or assigned to another guest which is then started, it would be reset anyway.  The fact that a hot-detach doesn't reset the device is fine.

Perhaps this actually solves the race condition problem...

## Race Condition
With respect to the race condition, OpenXT should mitigate this, as a domain _A_ will still be in the xenmgr 'stopping' state and the toolstack won't let you start a vm _B_ with PCI device _x_ if vm _A_ has _x_ assigned and state != 'stopped'.  

Continuing my thoughts from the above section, Xl has multiple calls in ```libxl__device_pci_add``` that check if a device is already assigned, it first tries ```xc_test_assign_device``` to see if Xen is ok with it.  Then it checks xenstore to see if it's assigned to a domain.  If we don't reset_device on domain destruction and simply reset it on hotplug-add/domain creation, it should be deassigned from the previous domain, and there won't be a callback queued to fire ```device_reset``` while it's assigned to the new guest.

If this sounds like a sound approach, I can update the patch to only perform device reset on hotplug-add/domain creation.

## Xenbus
I will follow up this comment with whatever I can find about the xenbus states.  Still not optimistic about finding the cause in a reasonable amount of time.

jandryuk:
## Xenbus
This is may summary of a domain shutdown call chain:
libxl__domain_destroy
--libxl__destroy_domid
----libxl__device_pci_destroy_all
------libxl__device_pci_remove_common
--------do_pci_remove
----------libxl__device_pci_reset [0]
----------!stubdom : -> xc_deassign_device
----------stubdomid : -> libxl__device_pci_remove_common
----------libxl__device_pci_remove_xenstore
----xc_domain_pause
----libxl__devices_destroy
------libxl__initiate_device_remove [1] (Sets state to XenbusStateClosing)

domain_destroy_callback
--libxl__device_pci_reset [2]

@crogers1 's patch moves libxl_device_pci_reset later in the domain shutdown process from [0] to [2].  At location [1], libxl__devices_destroy iterates all entries in xenstore and libxl__initiate_device_remove sets the devices to XenbusStateClosing.  So when libxl__device_pci_reset is called later in the shutdown process, devices are in XenbusStateClosing or closed.

Full disclosure, I haven't traced through how or when domain_destroy_callback is actually invoked.

I haven't really looked, but as @dozylynx mentions, a hotplug removal will call libxl__device_pci_remove_common from a shorter call chain than the full domain destroy.  Tying PCI reset to domain shutdown means a PCI device will not be reset in that case.

crogers1:
@jandryuk Thanks for tracking down the state changes.

Once @dozylynx has a chance to look over my changes and comment, I still need to port the relevant fixes to my stable-7 PR.

crogers1:
If we don't move the actual libxl__device_pci_reset call from where it is currently, device_reset will fail if the kernel code determines it needs to perform a slot-level reset for that particular device.  Upstream xl code was designed to do function-level resets (FLR) only, but the kernel patch expands that capability from a simple FLR to:

 FLR --(if failed or can't then try)--> Slot-level reset ---(if failed or can't then try)----> bus-level reset.  

I explained the example case in one of my earlier comments.

eric-ch:
@crogers1 Right! I got off-track... sorry :).

I do not think ```xl pci-attach/pci-detach``` could be handled correctly with the FLR->SLR->SBR method, at least not without some logic in libxl to handle each PCI passed-through device capabilities and reset requirement, or deferring the reset logic to something else that has knowledge of the VM current configuration (xenmgr?).

Taking the hot-plug off the picture, devices are reset at domain creation and destruction. Since the FLR->SLR->SBR logic needs to be done when all the devices are de-assigned, wouldn't it be easier to move ```libxl__device_pci_destroy_all``` from ```libxl__destroy_domid``` to ```domain_destroy_callback``` directly, preserving the PV PCI pass-through logic (Xenbus states) and ```libxl__``` signatures from upstream?
The PCI devices will be de-assigned and the Xenstore nodes handled before being reset when the ```xl create <cfg>``` daemon terminates. There might be some double-reset though, that could be trouble, but it is already the case with this patch and it seems fine.

Although I am unable to find it, during OpenXT 6 testing some GPUs were reported to not support FLR (other devices likely share that trait). PCI pass-through will probably fail on these without this PR.

eric-ch:
@crogers1 Right! I got off-track... sorry :).

I do not think ```xl pci-attach/pci-detach``` could be handled correctly with the FLR->SLR->SBR method, at least not without some logic in libxl to handle each PCI passed-through device capabilities and reset requirement, or deferring the reset logic to something else that has knowledge of the VM current configuration (xenmgr?).

Although I am unable to find it, during OpenXT 6 testing some GPUs were reported to not support FLR (other devices likely share that trait). PCI pass-through will probably fail on these without this PR.

dozylynx:
Excellent to see this analysis and the patch progressing.
It's now clear that a patch such as this is going to be necessary:

* to reintroduce use of the multi-level device reset sysfs node in order to retain compatibility with devices supported in earlier releases.

* to change the point in libxl where the reset occurs, because the more forceful reset methods won't work correctly where it currently is.

----
I'm still unsure on the Xenbus state details so far though. It looks like there are only two places in libxl that write `XenbusStateClosing`:

* `libxl__initiate_device_generic_remove` : writes it to main device xs node
* `libxl__device_pci_remove_xenstore` : writes it to the subdevice xs node

and I don't see from the call flow how `libxl__initiate_device_generic_remove` gets called before `libxl__device_pci_remove_xenstore`, or that it writes to the same node that we're observing is Closed (`state` vs. `state-%d`).

Instead, is the frontend writing `XenbusStateClosing` in response to something it observes?

Notice: The patch moves the device reset from before the call to `xc_deassign_device` to after it. `xc_deassign_device` invokes a Xen domctl that pulls the device from the domain.
It reconfigures the IOMMU. It shouldn't affect the actual device reset logic, but it's a serious enough change that it makes practical testing important.

It would be useful to find whether the guest is notified that the assigned permissions have changed due to `xc_deassign_device`. If the frontend is made aware of this permission change, it would presumably trigger the device release logic in the frontend kernel, and conveniently explain why the XenbusState is
Closed, not Connected at that point. I haven't confirmed this theory though.

----
In Eric-ch's comment above, it looks like there are lines being printed by error conditions:

```
    rc = xc_physdev_unmap_pirq(ctx->xch, domid, irq);
    if (rc < 0) {
        LOGED(ERROR, domainid, "xc_physdev_unmap_pirq irq=%d", irq);
    }
    rc = xc_domain_irq_permission(ctx->xch, domid, irq, 0);
    if (rc < 0) {
        LOGED(ERROR, domainid, "xc_domain_irq_permission irq=%d", irq);
    }
```
(It would be nicer if those log lines indicated the error code.)

I think the irq permission has already been removed before that runs.
OpenXT has a relevant Xen patch which modifies that IRQ permission code:
* `stubdomain-msi-irq-access.patch`

That patch revokes stubdomain irq permissions when the guest permissions are dropped, so I think the errors in Eric's log are OpenXT-specific. They look harmless.

----
Questions still open:

* Is the XenBus state machine behaviour OpenXT-specific?
* Are these state machine changes required?
-> feeds into deciding whether the device hot-add/remove code paths have to be separated out from domain destroy.

* Is it acceptable to only reset a device on adding it to a domain?

-> I think it would be ok. I think it's preferable to issuing a reset to a device that Xen thinks is deassigned and that is absent from xenstore, to avoid any race.

* What is the correct way to reset devices for hot-add and hot-remove?

-> On this one, my instinct is to trust the kernel of the hardware domain to be the canonical place for querying what PCI devices are assigned to a particular domain and what their capabilities and reset requirements are, rather than a higher-level toolstack, since that would hopefully enable sharing the development and testing with other projects.

dpsmith:
In the case of a Linux guest, ```pcifront_backend_change```[[1]](http://elixir.free-electrons.com/linux/latest/source/drivers/pci/xen-pcifront.c#L1087) is fired when the backend changes and with the new state being ```XenbusStateClosing```, then ```pcifront_try_disconnet```[[2]]( http://elixir.free-electrons.com/linux/latest/source/drivers/pci/xen-pcifront.c#L925) is fired. I haven't confirmed this but I believe when entering ```pcifront_try_disconnet``` that ```prev_state``` is ```XenbusStateConnected``` which would result in the disconnecting the device and the xenbus state changing to ```XenbusStateClosed```.

jandryuk:
I think I was wrong:
```
libxl__device_pci_remove_common
  do_pci_remove
    libxl__device_pci_reset [0]
    !stubdom : -> xc_deassign_device
    stubdomid : -> libxl__device_pci_remove_common
    libxl__device_pci_remove_xenstore
```
libxl__device_pci_reset is the moved function, but it does not directly write to Xenstore.
libxl__device_pci_remove_xenstore has XenbusState waits.

So unless the PCI reset changes the xenstore from the kernel side, that code movement is not directly responsible.

The other thing to consider is HVM and stubdom.  In that case, 

> libxl__domain_destroy
```
  libxl__destroy_domid
----libxl__device_pci_destroy_all
------libxl__device_pci_remove_common
--------do_pci_remove
----------libxl__device_pci_reset [0]
----------!stubdom : -> xc_deassign_device
----------stubdomid : -> libxl__device_pci_remove_common
------------do_pci_remove
--------------libxl__device_pci_reset [0]
--------------!stubdom : -> <noop>
--------------libxl__device_pci_remove_xenstore
----------libxl__device_pci_remove_xenstore
----xc_domain_pause
----libxl__devices_destroy
------libxl__initiate_device_remove [1] (Sets state to XenbusStateClosing)```

**Comment by eric-ch** [comment]
> it looks like there are lines being printed by error conditions[...]

Yes, the ```syslog``` patch looses the ```errno``` value instead of printing it. It can be shoved back in the log. I can PR that, but I think this should be reworked entirely. It gives:
```
xl: libxl_pci.c:1492:do_pci_remove: (Invalid argument) Domain 4:xc_physdev_unmap_pirq irq=168
xl: libxl_pci.c:1496:do_pci_remove: (Operation not permitted) Domain 4:xc_domain_irq_permission irq=168
```
```Libxl``` provides a way to register a logging scheme, it should be used instead of replacing that logic with a call to ```syslog(3)```, but that is another discussion.

> Is it acceptable to only reset a device on adding it to a domain?

That seemed easy to try, so I removed the reset from the domain destruction path and tried with NDVM. Weird behaviors ensued after NDVM shutdown. Although the platform appeared fine, the switch to which it was connected stopped worked more often than not, with other devices on that same switch being cut-off. Probably the device going ballistic and storming packets clogging the switch...

> What is the correct way to reset devices for hot-add and hot-remove?

I think PCI pass-through should be limited to hierarchies depending on the requirement of what is to be passed-through. ```xl``` already handles passing "all functions" of a device ([1d05a5dafb xl: pci multi-function passthrough v2](https://xenbits.xen.org/gitweb/?p=xen.git;a=commit;h=1d05a5dafbb1fd1ca0872f61a374c0bc12f618b5)), which could be extended, but it does not seem to be trivial to check each capability ahead which throws a wrench in this. *thorough-reset-interface-to-pciback-s-sysfs.patch* already relies on the tool-stack to do the *right thing* and make sure, for example, when a slot-level reset is required, all functions will be assigned to pciback, yet that has to be in ```xenmgr```. If that logic is added to ```libxl```, the upper level toolstack (e.g, xenmgr) will have to abide to pass the required hierarchy, or ```libxl``` would return an error. That would leave the details of the reset to the kernel (although it has to expose it for user-land use).

Side-note: I was going through the *thorough-reset-interface-to-pciback-s-sysfs.patch*. It falls through [__pci_dev_reset](http://elixir.free-electrons.com/linux/v4.9.52/source/drivers/pci/pci.c#L3970), which seems to try every reset method (adds PM reset to the 3-reset-list) until one succeeds. Since this is eventually exposed, is it required to check support before hand instead of just going through this?

**Comment by richpersaud** [comment]
Qubes is using a combinaion of LibXL and Libvirt to reset PCI devices.  I've updated OXT-1217 with links to Qubes LibXL patches and the relevant PCI reset code in Libvirt.

**Comment by eric-ch** [comment]
Regarding the discussion on the 10/05 call:
```++        if (libxl__wait_for_backend(gc, be_path, GCSPRINTF("%d", XenbusStateClosed)) < 0) {```
Checks for (assuming the PV guest is dom1):
```/local/domain/0/backend/pci/1/0/state = "4"   (n0,r1)```
Which is obviously specific to the set of devices passed-through to that guest. My apologies for the (embarrassing) confusion.

**Comment by eric-ch** [comment]
After some tracing with and without this ```libxl``` series, I could not trigger a race on shutdown. The following logs are trimmed for readability.
### Without PCI reset deferred in ```domain_destroy_callback```:
```
Oct  5 22:30:05.002109 root: /state/00000000-0000-0000-0000-000000000002/state: running
Oct  5 22:30:05.002164 root: /local/domain/0/backend/pci/1/0/state: 4
Oct  5 22:30:12.699333 xec: invoked with: list-vms
Oct  5 22:30:12.736656 xec: invoked with: -o /vm/00000000_0000_0000_0000_000000000002 shutdown
Oct  5 22:30:12.756728 xenmgr: shutting down VM 00000000-0000-0000-0000-000000000002
Oct  5 22:30:12.758002 xenmgr: active vms, state = "running"
Oct  5 22:30:12.758126 xenmgr: active vms, state = "running"
Oct  5 22:30:12.773254 xl: shutdown -c 1
Oct  5 22:30:12.775449 xenmgr: vm state change 00000000-0000-0000-0000-000000000002: shutdowning
Oct  5 22:30:12.779227 xenmgr: active vms, state = "shutdowning"
Oct  5 22:30:12.779385 xenmgr: active vms, state = "shutdowning"
Oct  5 22:30:12.788564 root: /state/00000000-0000-0000-0000-000000000002/state: shutdowning
Oct  5 22:30:12.806259 xenconsoled: Network (1): shutdown: shutting down for system halt
[... NDVM kernel shutting down...]
Oct  5 22:30:25.175279 xenconsoled: Network (1): [  221.892606] reboot: System halted
Oct  5 22:30:25.178721 xenmgr: vm state change 00000000-0000-0000-0000-000000000002: shutdowning
Oct  5 22:30:25.181300 xenmgr: active vms, state = "shutdowning"
Oct  5 22:30:25.181443 xenmgr: active vms, state = "shutdowning"
Oct  5 22:30:25.186914 xl: libxl_pci.c:1152:libxl__device_pci_reset: Domain 0:Reset device 00:1f.6...
Oct  5 22:30:25.191548 root: /state/00000000-0000-0000-0000-000000000002/state: shutdowning
Oct  5 22:30:25.192521 root: /local/domain/0/backend/pci/1/0/state: 6
Oct  5 22:30:25.307100 xl: libxl_pci.c:218:libxl__device_pci_remove_xenstore: Domain 1:wait_for_backend(/local/domain/0/backend/pci/1/0, XenbusStateConnected)...
Oct  5 22:30:36.265896 xl: libxl_device.c:1384:libxl__wait_for_backend: Backend /local/domain/0/backend/pci/1/0 not ready
Oct  5 22:30:36.268912 xl: libxl_pci.c:1500:do_pci_remove: (Invalid argument) Domain 1:xc_physdev_unmap_pirq irq=168
Oct  5 22:30:36.268942 xl: libxl_pci.c:1504:do_pci_remove: (Operation not permitted) Domain 1:xc_domain_irq_permission irq=168
Oct  5 22:30:36.269220 xl: libxl_pci.c:1152:libxl__device_pci_reset: Domain 0:Reset device 01:00.0...
Oct  5 22:30:36.389567 xl: libxl_pci.c:218:libxl__device_pci_remove_xenstore: Domain 1:wait_for_backend(/local/domain/0/backend/pci/1/0, XenbusStateConnected)...
Oct  5 22:30:47.315907 xl: libxl_device.c:1384:libxl__wait_for_backend: Backend /local/domain/0/backend/pci/1/0 not ready
Oct  5 22:30:47.319541 xl: libxl_pci.c:1152:libxl__device_pci_reset: Domain 0:Reset device 02:00.0...
Oct  5 22:30:47.439197 xl: libxl_pci.c:218:libxl__device_pci_remove_xenstore: Domain 1:wait_for_backend(/local/domain/0/backend/pci/1/0, XenbusStateConnected)...
Oct  5 22:30:58.385891 xl: libxl_device.c:1384:libxl__wait_for_backend: Backend /local/domain/0/backend/pci/1/0 not ready
Oct  5 22:30:58.482026 root: /local/domain/0/backend/pci/1/0/state: no longer exists.
[... tapdisk stuff...]
[xl create dies... -> domain_destroy_callback]
Oct  5 22:30:58.751368 xenmgr: vm state change 00000000-0000-0000-0000-000000000002: shutdown
Oct  5 22:30:58.761818 root: /state/00000000-0000-0000-0000-000000000002/state: shutdown
[... network-daemon things...]
Oct  5 22:30:58.783625 xenmgr: vm 00000000-0000-0000-0000-000000000002 shutdown, reason: Halt
```
### Just for control, with build 1178 unmodified:
(no trace, one script watching ```/local/domain/0/backend/pci/1/0/state``` and ```/state/00000000-0000-0000-0000-000000000002/state``` and printing the state/lack of node).P
```
Oct  5 23:56:31.820981 root: /local/domain/0/backend/pci/1/0/state: 4
Oct  5 23:56:31.821002 root: /state/00000000-0000-0000-0000-000000000002/state: running
Oct  5 23:56:39.608122 xec: invoked with: list-vms
Oct  5 23:56:39.629480 xec: invoked with: -o /vm/00000000_0000_0000_0000_000000000002 shutdown
Oct  5 23:56:39.641429 xenmgr: shutting down VM 00000000-0000-0000-0000-000000000002
Oct  5 23:56:39.642576 xenmgr: active vms, state = "running"
Oct  5 23:56:39.642710 xenmgr: active vms, state = "running"
Oct  5 23:56:39.653386 xenmgr: vm state change 00000000-0000-0000-0000-000000000002: shutdowning
Oct  5 23:56:39.654632 xenmgr: active vms, state = "shutdowning"
Oct  5 23:56:39.654710 xenmgr: active vms, state = "shutdowning"
Oct  5 23:56:39.658641 xenconsoled: Network (1): shutdown: shutting down for system halt
Oct  5 23:56:39.659009 root: /state/00000000-0000-0000-0000-000000000002/state: shutdowning
Oct  5 23:56:39.659976 xenconsoled: Network (1): INIT: Switching to runlevel: 0
[... NDVM kernel shutting down...]
Oct  5 23:56:51.923351 xenconsoled: Network (1): [  217.171228] reboot: System halted
Oct  5 23:56:51.925101 xenmgr: vm state change 00000000-0000-0000-0000-000000000002: shutdowning
Oct  5 23:56:51.928484 xenmgr: active vms, state = "shutdowning"
Oct  5 23:56:51.928581 xenmgr: active vms, state = "shutdowning"
Oct  5 23:56:51.932781 root: /local/domain/0/backend/pci/1/0/state: 6
Oct  5 23:56:51.933024 root: /state/00000000-0000-0000-0000-000000000002/state: shutdowning
Oct  5 23:57:03.041800 xl: libxl_device.c:1384:libxl__wait_for_backend: Backend /local/domain/0/backend/pci/1/0 not ready
Oct  5 23:57:03.129625 root: /local/domain/0/backend/pci/1/0/state: no longer exists.
[... tapdisk stuff...]
Oct  5 23:57:03.387546 xenmgr: vm state change 00000000-0000-0000-0000-000000000002: shutdown
Oct  5 23:57:03.393103 root: /state/00000000-0000-0000-0000-000000000002/state: shutdown
[... network-daemon things...]
Oct  5 23:57:03.402975 xenmgr: vm 00000000-0000-0000-0000-000000000002 shutdown, reason: Halt
```
Note: This machine only has ```00:19.0 Ethernet controller: Intel Corporation Ethernet Connection I217-LM (rev 04)```, hence only one "backend not ready".

### With PCI reset deferred in ```domain_destroy_callback```:
```
Oct  5 22:48:44.876681 root: /state/00000000-0000-0000-0000-000000000002/state: running
Oct  5 22:48:44.876684 root: /local/domain/0/backend/pci/1/0/state: 4
Oct  5 22:48:57.049092 xec: invoked with: -o /vm/00000000_0000_0000_0000_000000000002 shutdown
Oct  5 22:48:57.068816 xenmgr: shutting down VM 00000000-0000-0000-0000-000000000002
Oct  5 22:48:57.070163 xenmgr: active vms, state = "running"
Oct  5 22:48:57.070273 xenmgr: active vms, state = "running"
Oct  5 22:48:57.084544 xl: shutdown -c 1
Oct  5 22:48:57.088201 xenmgr: vm state change 00000000-0000-0000-0000-000000000002: shutdowning
Oct  5 22:48:57.092258 xenmgr: active vms, state = "shutdowning"
Oct  5 22:48:57.092463 xenmgr: active vms, state = "shutdowning"
Oct  5 22:48:57.096040 xenconsoled: Network (1): shutdown: shutting down for system halt
[... NDVM kernel shutting down...]
Oct  5 22:49:09.478488 xenconsoled: Network (1): [  214.213781] reboot: System halted
Oct  5 22:49:09.482773 xenmgr: vm state change 00000000-0000-0000-0000-000000000002: shutdowning
Oct  5 22:49:09.485596 xenmgr: active vms, state = "shutdowning"
Oct  5 22:49:09.485815 xenmgr: active vms, state = "shutdowning"
Oct  5 22:49:09.491296 root: /local/domain/0/backend/pci/1/0/state: 6
Oct  5 22:49:09.492258 xl: libxl_pci.c:221:libxl__device_pci_remove_xenstore: Domain 1:wait_for_backend(/local/domain/0/backend/pci/1/0, XenbusStateClosed)...
Oct  5 22:49:09.492870 xl: libxl_pci.c:258:libxl__device_pci_remove_xenstore: Domain 1:wait_for_backend(/local/domain/0/backend/pci/1/0, XenbusStateClosed)...
Oct  5 22:49:09.501200 root: /state/00000000-0000-0000-0000-000000000002/state: shutdowning
Oct  5 22:49:09.503609 xl: libxl_pci.c:1500:do_pci_remove: (Invalid argument) Domain 1:xc_physdev_unmap_pirq irq=168
Oct  5 22:49:09.503688 xl: libxl_pci.c:1504:do_pci_remove: (Operation not permitted) Domain 1:xc_domain_irq_permission irq=168
Oct  5 22:49:09.505683 xl: libxl_pci.c:221:libxl__device_pci_remove_xenstore: Domain 1:wait_for_backend(/local/domain/0/backend/pci/1/0, XenbusStateClosed)...
Oct  5 22:49:09.506644 xl: libxl_pci.c:258:libxl__device_pci_remove_xenstore: Domain 1:wait_for_backend(/local/domain/0/backend/pci/1/0, XenbusStateClosed)...
Oct  5 22:49:09.514631 xl: libxl_pci.c:221:libxl__device_pci_remove_xenstore: Domain 1:wait_for_backend(/local/domain/0/backend/pci/1/0, XenbusStateClosed)...
Oct  5 22:49:09.515439 xl: libxl_pci.c:258:libxl__device_pci_remove_xenstore: Domain 1:wait_for_backend(/local/domain/0/backend/pci/1/0, XenbusStateClosed)...
Oct  5 22:49:09.535131 root: /local/domain/0/backend/pci/1/0/state: no longer exists.
[... tapdisk stuff...]
[xl create dies... -> domain_destroy_callback]
Oct  5 22:49:09.879744 xl: libxl_pci.c:1152:libxl__device_pci_reset: Domain 0:Reset device 00:1f.6...
Oct  5 22:49:09.994020 xl: libxl_pci.c:1152:libxl__device_pci_reset: Domain 0:Reset device 01:00.0...
Oct  5 22:49:10.116623 xl: libxl_pci.c:1152:libxl__device_pci_reset: Domain 0:Reset device 02:00.0...
Oct  5 22:49:10.237634 xenmgr: vm state change 00000000-0000-0000-0000-000000000002: shutdown
Oct  5 22:49:10.248617 root: /state/00000000-0000-0000-0000-000000000002/state: shutdown
[... network-daemon things...]
Oct  5 22:49:10.449769 xenmgr: vm state change 00000000-0000-0000-0000-000000000002: shutdown
Oct  5 22:49:10.454017 xenmgr: vm 00000000-0000-0000-0000-000000000002 shutdown, reason: Halt
```

I did not reach the race under these conditions, but I am having trouble understanding the Haskell logic that monitors the VM state ```xenmgr/Vm/React.hs:vmEventProcessor```.

```notifyVmStateUpdate``` is reading the ```/state/<uuid>/state``` Xenstore node, and this is written in ```xl/xl_vmcontrol.c:destroy_domain``` (in the case of a shutdown). So that will happen before the ```xl create``` daemon runs ```domain_destroy_callback```. That has potential to race depending on how ```xenmgr``` reacts on Xenstore node ```/state/<uuid>/state``` and ```xl create``` daemon terminating.

crogers1:
It's actually pretty simple.  xenmgr puts a xenstore-watch on the /state/<uuid>/state node when vm is created (like from Install VM).  ```notifyVmStateUpdate``` is fired when that xenstore node is written to by libxl.  That sends a dbus call to all listeners (such as the UI, other xenmgr threads) that a vm state change occurred.

I used ```update_state_direct``` (which takes a uuid, instead of a domid) in domain_destroy because the domid was no longer valid by the time I reached that point in execution.  Even so, I don't think the lack of a domid is a strong enough assertion to guarantee that the ```domain_destroy_cb``` is called before xenmgr makes its state transition.  Maybe we should move ```update_state_direct``` to the last callback on the domain cleanup path.

eric-ch:
> xl/xl_vmcontrol.c:destroy_domain (in the case of a shutdown).

```xl/xl_vmcontrol.c:destroy_domain``` (in the case of a **destroy**), but that is required. In case I confused anyone, for shutdown, or reboot, the state change is done in ```xl/xl_vmcontrol.c:create_domain``` (see the ```switch case```) which is daemonized.

The ```destroy``` case needs to set the node since the daemon will not handle it.
No race there as ```xenmgr``` is synchronized on the state node (without using ```xl``` directly that is) and no need to move ```update_state_direct``` in the ```domain_destroy_cb``` since the node is set after everything is done.

Issuing a ```destroy``` relatively soon after ```reboot``` to NDVM will have the ```destroy ``` fail and NDVM reboot. ```xl``` had a segfault once in 15 tests (with delays between ```reboot``` and ```destroy``` between 0 and 3 seconds, so that should be checked again.

Since I did not try that on a build without this series, that might happen on master as well.

eric-ch:
> ```xl``` had a segfault once in 15 tests

Cannot reproduce. This might have been some of the instrumentation that I put in.

eric-ch:
Building [1208](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1208)

eric-ch:
Following 10/19/2017 Triage & community call: Will merge soon.

PR #773, opened 2017-09-20
STABLE-7: [libxl] Fix flr for pci passthru device

crogers1:
  guest boots.

  on guest shutdown.

  OXT-1217


eric-ch:
Building [1152](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1152)

PR #774, opened 2017-09-20
STABLE-7: [libxl] Fix flr for pci passthru devices

crogers1:
  guest boots.

  on guest shutdown.

  OXT-1217

### DIFF
--- a/recipes-extended/xen/files/libxl-fix-flr.patch
+++ b/recipes-extended/xen/files/libxl-fix-flr.patch
@@ -1,0 +1,207 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Fix device reset for pci devices in libxl
+
+################################################################################
+LONG DESCRIPTION:
+################################################################################
+This patch fixes the logic and the control flow path for pci device reset in
+libxl.  Additionally, it addresses an issue with Xenbus state synchronization
+during domain teardown.  More detailed information about the patch follows:
+
+The changes to Xenbus state are to prevent hitting a timeout when domains are
+shutdown or destroyed. Xl is waiting for "XenbusStateConnected" (state 4), while
+the state machine has already moved the device to Closing (state 5) and Closed
+(state 6) by the time we reach this point tearing down the domain.
+
+The reset logic introduced in this patch is delayed until the domain has fully
+been destroyed. The reason for this has to do with the
+thorough-reset-interface-to-pciback-s-sysfs.patch in the linux patchqueue.
+__pcistub_raw_device_reset tries multiple approaches for resetting the device,
+flr, slot-level, and bus-level. If we encounter a device that, for example,
+doesn't support flr but does support a slot-level reset, we want to make sure
+all the functions on that device (GPUS are a good example, they often have 2
+functions, video and HDMI audio, and are the only device in the slot) are
+released from the domain before attempting the reset or it will fail. This
+approach seems to be taken in an attempt to support a wide variety of PCI
+devices.
+
+################################################################################
+CHANGELOG
+################################################################################
+Authors:
+Chris Rogers <rogersc@ainfosec.com>
+
+################################################################################
+REMOVAL
+################################################################################
+N/A
+
+################################################################################
+UPSTREAM PLAN
+################################################################################
+Depends on the feasibility of upstreaming thorough-reset-interface-to-pciback...
+
+################################################################################
+INTERNAL DEPENDENCIES
+################################################################################
+
+################################################################################
+PATCHES
+################################################################################
+Index: xen-4.6.6/tools/libxl/libxl.c
+===================================================================
+--- xen-4.6.6.orig/tools/libxl/libxl.c
++++ xen-4.6.6/tools/libxl/libxl.c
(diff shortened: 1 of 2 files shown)


eric-ch:
Building [1153](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1153)

recipes-extended/xen/files/libxl-fix-flr.patch
```diff
@@ -0,0 +1,154 @@
...
+--- xen-4.6.6.orig/tools/libxl/libxl.c
++++ xen-4.6.6/tools/libxl/libxl.c
+@@ -1604,7 +1604,13 @@ static void domain_destroy_callback(libx
+         LOG(ERROR, "unable to destroy guest with domid %u", dis->domid);
+         dds->rc = rc;
+     }
+-
++    if(dis->pciw) {
++        for (int i = 0; i < dis->pciw->num_devs; i++) {
++            libxl__device_pci_reset(gc, dis->pciw->pcidevs[i].domain, dis->pciw->pcidevs[i].bus, dis->pciw->pcidevs[i].dev, dis->pciw->pcidevs[i].func);
```
dpsmith: Just a stylistic comment, but you could have done a local reference variable here(see below) and it would have reduced keystrokes and line length
```c
        for (int i = 0; i < dis->pciw->num_devs; i++) {
             libxl_device_pci *pcidev = &dis->pciw->pcidevs[i];
￼             libxl__device_pci_reset(gc, pcidev->domain, pcidev->bus, pcidev->dev, pcidev->.func);
        }
```

recipes-extended/xen/files/libxl-fix-flr.patch
```diff
@@ -0,0 +1,154 @@
...
+     destroy_finish_check(egc, dds);
+ }
+@@ -1672,8 +1678,8 @@ void libxl__destroy_domid(libxl__egc *eg
+         goto out;
+     }
+ 
+-    if (libxl__device_pci_destroy_all(gc, domid) < 0)
+-        LIBXL__LOG(ctx, LIBXL__LOG_ERROR, "pci shutdown failed for domid %d", domid);
++    libxl_pci_dev_wrap *pciw = libxl__device_pci_destroy_all(gc, domid);
++    dis->pciw = pciw;
```
dpsmith: Another stylistic comment. Why use the extra variable to temporarily store the return value and not just store the return value directly into struct instance?

dpsmith:
Outside of the style comments, looks good. I will wait for @eric-ch to report on the build and checkout and then will sign-off.

recipes-extended/xen/files/libxl-fix-flr.patch
crogers1: Yea, I was aiming for fewer total lines, but if the shorter line length is desired I can edit the patch

recipes-extended/xen/files/libxl-fix-flr.patch
crogers1: Can't remember specifically, might have planned on using _pciw_ later on in the function, never did, and then never performed the direct assignment.

dpsmith:
If @eric-ch and @jean-edouard okay, you could make the changes for the merge and not do a rebuild/retest since the are not logic changes.

jandryuk:
Changing "do_flr" -> "reset_device" makes sense to actually perform the reset.

Why is the reset delayed until after the the domain is destroyed?  It's not clear if this delayed reset is needed for the AMD GPUs or just actually performing the reset.

Are the XenBus changes just to support the movement of reset until after the domain is destroyed?

The changelog should be improved to clarify these points.

crogers1:
Xenbus state changes are to prevent hitting a timeout when domains are shutdown/destroyed.  Xl is waiting for "XenbusStateConnected" (state 4) and from my testing, the state machine has already moved the device to Closing (state 5) or Closed (state 6) by the time we reach this point tearing down the domain.

As for the delayed reset, if you look in thorough-reset-interface-to-pciback-s-sysfs.patch in the linux patchqueue, ```__pcistub_raw_device_reset``` tries multiple approaches for resetting the device, flr, slot-level, and bus-level.  If we encounter a device that, for example, doesn't support flr but _does_ support a slot-level reset, we want to make sure all the functions on that device (GPUS are a good example, they often have 2 functions, video and HDMI audio, and are the only device in the slot) are released from the domain before attempting the reset or it will fail.

If @eric-ch or someone with more xen/pci passthru experience sees that I've missed something in the above explanation or have the wrong assumption, please correct me.  Otherwise I believe that covers it.

jandryuk:
@crogers1, thanks for the info.  I think the patch would benefit from this extended description of why the delayed reset is needed.

crogers1:
Updated with patch header and style recommendations.

eric-ch:
Building [1155](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1155)

crogers1:
Fixed malformed patch error, patched and built ok on my machine.  Should be ready for oxt builder now.

crogers1:
Updated with syntax/logic fixes from review on the master PR

eric-ch:
Building [#1193](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1193)

PR #776, opened 2017-09-22
seabios: Change URL to avoid HTTP 308.

eric-ch:
Wget will fail to fetch from the http URL after receiving a 308. Use the
https source instead.

This seems to fail when building a new tree without cache.


dpsmith:
Is there any reason we couldn't just move up to the seabios recipe in meta-virtualization? I did a quick scan of our patch set and they all seem very unintrusive.

eric-ch:
@dpsmith meta-virtualization recipe targets ```SeaBIOS 1.9```, apart from that none (actually @cjp256 upstreamed therecipe to meta-virt from OpenXT initially).

jean-edouard:
LGTM, merging soon.

PR #781, opened 2017-10-12
OXT-1227: xen: XSA 237-244.

eric-ch:
https://xenbits.xen.org/xsa/
Released 2017-10-12 12:00.

XSA-245 is ARM specific, so left out of the OpenXT layer for now.


jean-edouard:
Tested on stable-7, LGTM, merging soon.

PR #785, opened 2017-10-18
refpolicy: rpc-proxy: allow to signal itself

jean-edouard:
OXT-1228


eric-ch:
Will merge soon.

PR #789, opened 2017-10-20
ustr: use github instead of unreliable and.org

jean-edouard:


jean-edouard:
Note: not sure if the patches are getting applied, they don't seem to but it also doesn't seem to be a problem...

jean-edouard:
Note2: we should consider upgrading the SELinux tools to version 2.7, which doesn't seem to require ustr anymore.

eric-ch:
building [#1220](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1220)

jean-edouard:
Failed to build, will have a look tomorrow.

tklengyel:
I think this would have been a good thing, build was broken on my machine because `and.org` was down. I see applying an old patch failed on the latest master branch, but maybe that patch is just no longer needed?

PR #792, opened 2017-10-27
python-ipy: fix URL to use https and make wget happy

jean-edouard:


eric-ch:
Avoid wget bailing out after
```
403 SSL is required
```
LGTM, will merge soon.

PR #796, opened 2017-11-03
xen: Xen debug builds.

eric-ch:
Xen with OpenXT patch-queue should pass typesafe.h verification used when building with debug (CONFIG_DEBUG=y -> #define NDEBUG).


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1239

jean-edouard:
LGTM, merging soon.

PR #798, opened 2017-11-09
Install ca-certificates in dom0 and the installer

jean-edouard:
OXT-1248

### DIFF
--- a/recipes-core/images/xenclient-dom0-image.bb
+++ b/recipes-core/images/xenclient-dom0-image.bb
@@ -9,8 +9,7 @@ IMAGE_FSTYPES = "xc.ext3.gz"
 
 # No thanks, we provide our own xorg.conf with the hacked Intel driver
 # And we don't need Avahi
-# Nor a collection of questionable CA certificates
-BAD_RECOMMENDATIONS += "xserver-xorg avahi-daemon avahi-autoipd ca-certificates"
+BAD_RECOMMENDATIONS += "xserver-xorg avahi-daemon avahi-autoipd"
 # The above seems to be broken and we *really* don't want avahi!
 # Also remove the unwanted version of xenstored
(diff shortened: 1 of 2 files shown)


recipes-core/images/xenclient-dom0-image.bb
```diff
@@ -9,8 +9,7 @@ IMAGE_FSTYPES = "xc.ext3.gz"
 
 # No thanks, we provide our own xorg.conf with the hacked Intel driver
 # And we don't need Avahi
-# Nor a collection of questionable CA certificates
-BAD_RECOMMENDATIONS += "xserver-xorg avahi-daemon avahi-autoipd ca-certificates"
+BAD_RECOMMENDATIONS += "xserver-xorg avahi-daemon avahi-autoipd"
```
dpsmith: Instead of an 'always install' or 'always not install' configuration, why not just make it configurable? The suggestion below would make it so that if a builder wants the standard web ca certs included, then they would follow [OE convention](http://www.yoctoproject.org/docs/2.3/dev-manual/dev-manual.html#usingpoky-extend-customimage-imagefeatures) by adding "web-certificates" to EXTRA_IMAGE_FEATURES in their local.conf. This also would make it the default not to include the certificates, thus making it an explicit action to source them.

```sh
BAD_RECOMMENDATIONS += "xserver-xorg avahi-daemon avahi-autoipd ${@base_contains('IMAGE_FEATURES', 'web-certificates', '', 'ca-certificates', d)}"
```

eric-ch: @dpsmith +1 to making it configurable.

jean-edouard: Done. See here: https://github.com/OpenXT/xenclient-oe/pull/804
and there: https://github.com/OpenXT/openxt/pull/276

jean-edouard:
Replaced, closing.

PR #807, opened 2017-11-29
OXT-1253: ocaml: Use the refactored OpenXT layer.

eric-ch:
No functional change.
Use the bbclass ocaml & findlib and associated variables.
Take the opportunity to remove dead-code.

### DIFF
--- a/recipes-openxt/xenclient/uid_git.bb
+++ b/recipes-openxt/xenclient/uid_git.bb
@@ -1,12 +1,7 @@
-inherit findlib
 DESCRIPTION = "UID - User Interface Daemon"
 LICENSE = "GPLv2"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/GPL-2.0;md5=801f80980d171dd6425610833a22dbe6"
-DEPENDS = "ocaml-cross ocaml-dbus xen-ocaml-libs xenclient-toolstack"
-
-# Ocaml stuff is built with the native compiler with "-m32".
-
-PV = "0+git${SRCPV}"
+DEPENDS = "ocaml-dbus xen-ocaml-libs xenclient-toolstack"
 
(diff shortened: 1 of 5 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1259
LGTM, will merge with the other refactorisation PRs.

recipes-openxt/xenclient/uid_git.bb
```diff
@@ -18,7 +13,7 @@ SRC_URI += "file://uid_dbus.conf \
 
 S = "${WORKDIR}/git"
 
-inherit xenclient update-rc.d
```
jean-edouard: uid doesn't seem to "inhibit rm-work" anymore, probably because it doesn't depend on xenclient.
It doesn't matter for the builder, but it probably does for development builds...
There might be other recipes that have this problem, please have a look.

eric-ch: I removed instances where recipes inherit `xenclient.bbclass` because it appeared that [`rm_work`](http://www.yoctoproject.org/docs/current/ref-manual/ref-manual.html#ref-classes-rm-work) is not done by default. It would need to inherit `rm_work.bbclass` then use exclusion through `RM_WORK_EXCLUDE` or disable `do_rm_work` like `xenclient.bbclass` does.

Does the builder have `INHERIT += "rm_work"` in the `local.conf` file?

jean-edouard: I checked build that don't include this PR, and the uid work directory also gets removed there, so my comment is irrelevant.
Will merge once https://github.com/OpenXT/meta-openxt-haskell-platform/pull/8 is settled.

jean-edouard:
Merging soon.

PR #809, opened 2017-12-05
Only spawn a getty on tty1 when booting in console mode

jean-edouard:

### DIFF
--- a/conf/machine/xenclient-dom0.conf
+++ b/conf/machine/xenclient-dom0.conf
@@ -18,3 +18,4 @@ MACHINE_FEATURES = "kernel26 screen keyboard ethernet pci usbhost acpi ext2 x86"
 
 SERIAL_CONSOLE = "115200 hvc0"
 
+USE_VT = "0"
--- a/recipes-openxt/xenclient/surfman/surfman.initscript
+++ b/recipes-openxt/xenclient/surfman/surfman.initscript
@@ -27,7 +27,11 @@ SURFMAN_OPTS=""
 
 case "$1" in
   start)
-        grep -q no-graphics /proc/cmdline && SURFMAN_OPTS="${SURFMAN_OPTS} -c"
+        grep -q no-graphics /proc/cmdline && {
+            SURFMAN_OPTS="${SURFMAN_OPTS} -c"
+            # We only want a getty on tty1 when in no-graphics mode (console mode)
+            while true; do /sbin/getty 38400 tty1; sleep 1; done &
+        }
         grep -q safe-graphic /proc/cmdline && SURFMAN_OPTS="${SURFMAN_OPTS} -s"
 
         # Allow 55MB for core dumps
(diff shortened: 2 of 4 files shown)


dpsmith:
I understand the reasoning but would have to say that this approach seems more like a work around than a solution. In this case you are working around the sysv init instead of leveraging it. My suggestion would be to properly use the sysv init runlevels and make so surfman only runs under runlevel 5 and that the getty on tty1 only runs under runlevel 3. This would be beneficial for my team as we leverage a keybinding to kill graphics when we want/need to drop back down to a tty. This would allow us to implement a more proper solution where the key binding would trigger a runlevel change when we need to drop to a tty. In the end, this is just a suggestion and I leave it to you and the community to consider.

richpersaud:
It would be helpful to have a JIRA ticket or RFC to document the requirement driving this change, so we can better evaluate possible implementations.

jean-edouard:
@dpsmith yes, the approach you're describing is the right way of doing it, and I tried it first but ran into issues. We would need all recipes to produce rc5 initscripts, which is not the case by default and would require a lot of bbappend work. Then once the work is done, we would need to make sure nothing ever breaks that mirroring, otherwise console boot and graphics boot would differ, which would make console boot mostly useless...
I sort of put this PR out there to see if anybody could come up with a better solution, so thanks for chiming in :) I'd be curious to see if you have more insight on how this could be done nicely in OE.
Please keep in mind that console boot is an unsupported debug mode, so we're allowed to get a bit dirty!

Either way, I think the fact that we can potentially get a shell for a few seconds while OpenXT is booting is a security concern... A human can't do much in 1 or 2 seconds, but an software defined OTG USB keyboard could surely do some damage!

dpsmith:
First, blocking tty access in non-debug is what I figured why you where doing the PR and definitely support the need for this!

Did a quick double check to make sure, the only daemons with an rc 5 and no rc 3 scripts are xenmgr and updatemgr, everything else is started in both rc 3 and 5. To fix xenmgr and updatemgr would just be to have the recipes correctly setup their sysv entries, i.e. inherit update-rc.d and provide the correct [INITSCRIPT_PARAMS](https://www.yoctoproject.org/docs/1.8/ref-manual/ref-manual.html#var-INITSCRIPT_PARAMS) variable. Doing the same for the surfman init script such that  it only runs in rc 5 and stops in 1-4. This should be how all the upstream recipes are setting up their initscripts.

eric-ch:
I would rather have the solution using the runlevels. @jandryuk seems to have one https://github.com/jandryuk/xenclient-oe/tree/runlevels (discussed in https://github.com/OpenXT/xenclient-oe/pull/846).

Since the problem needs to be addressed, I am in favor of leaving this open until the runlevel approach is PR'ed. If the runlevel solution is not PR'ed, this can be merged to work-around the issue.

Tracked under [OXT-1300](https://openxt.atlassian.net/browse/OXT-1300).

jandryuk:
I note some issues with my approach in the https://github.com/OpenXT/xenclient-oe/pull/846 discussion.  I can put up a PR with the existing code for discussion and feedback.  I don't have immediate availability for fixing the runlevel transition limitations.

eric-ch:
https://github.com/OpenXT/xenclient-oe/pull/858 got merged. This PR can now be closed.

PR #808, opened 2017-12-04
STABLE-7: OXT-1255: XSA 246 & 247

jandryuk:
This adds the patches for XSA 246 and 247.  While in there, I re-ordered the patch queue to keep all the XSAs at the end.  I made a separate commit for the refresh of xsa246-4.7.patch to clearly distinguish adding the upstream patch and making a modification to apply cleanly.

OXT-1255

PR #813, opened 2017-12-11
Revert "STABLE-7: OXT-1255: XSA 246 & 247"

jean-edouard:
Reverts OpenXT/xenclient-oe#808


jean-edouard:
These XSAs are about a feature that's not enabled in OpenXT (Populate on Demand).
Since we're very late in the release process of OpenXT 7.0.0, a lot of testing has already been done without these patches, so I suggest we remove them, at least for now.

dpsmith:
Another option to consider is that since testing thus far is against the RC2 tag, testing can finish against that tag the revision hash can be used for the 7.0.0 tag. I suggest this since there are four more XSA's coming tomorrow which as soon as 7.0.0 is released, theses two will need to be pulled right back in along with the four new ones for a 7.0.1 release. Thoughts?

MagOXT:
+1 for regression testing on RC2

jean-edouard:
@dpsmith that would work only as long as no critical issue is found in RC2... It also breaks my release script but I can fix that.
Your call, please approve/disapprove and someone will merge/close this PR.
Thanks

PR #814, opened 2017-12-13
OXT-1253: toolstack: Use the ocamlincdir from the bbclass.

eric-ch:
This should have been part of the original commit as it was defined in
the local.conf which got move to the bbclass.

Also use the correct variable to find xen headers.

### DIFF
--- a/recipes-openxt/xenclient/xenclient-toolstack_git.bb
+++ b/recipes-openxt/xenclient/xenclient-toolstack_git.bb
@@ -60,9 +60,9 @@ PACKAGE_ARCH = "${MACHINE_ARCH}"
 S = "${WORKDIR}/git"
 
 # TODO: ocamlc can figure it out in the build-system.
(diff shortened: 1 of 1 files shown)


eric-ch:
See https://github.com/OpenXT/openxt/pull/279

recipes-openxt/xenclient/xenclient-toolstack_git.bb
```diff
@@ -60,9 +60,9 @@ PACKAGE_ARCH = "${MACHINE_ARCH}"
 S = "${WORKDIR}/git"
 
 # TODO: ocamlc can figure it out in the build-system.
-CFLAGS_append = " -I${OCAML_HEADERS}"
+CFLAGS_append = " -I${ocamlincdir}"
```
tklengyel: I'm getting a build error on a fresh setup with this definition as is here. Switching it back to `OCAML_HEADERS` seems to have fixed it.

PR #815, opened 2017-12-14
Add blktap3 support

mahantesh-ais:
OXT-1221


mahantesh-ais:
This PR should be merged along with:
https://github.com/OpenXT/installer/pull/62
https://github.com/OpenXT/manager/pull/107
https://github.com/OpenXT/toolstack/pull/33
https://github.com/OpenXT/xsm-policy/pull/17
https://github.com/OpenXT/blktap3/pull/1

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,703 @@
...
+-                                                disk->format);
+-                    if (!dev) {
+-                        LOGD(ERROR, domid, "Failed to get blktap devpath for %p",
+-                             disk->pdev_path);
+-                        rc = ERROR_FAIL;
+-                        goto out;
+-                    }
+-                }
+-                flexarray_append(back, "tapdisk-params");
++		rc = 0;
```
dpsmith: this is pedantic

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,703 @@
...
++		dev = libxl__blktap_devpath(gc, disk->pdev_path, disk->format);
```
dpsmith: please don't reformat code, especially when your change breaks the 80 column rule that xen code conforms to

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,703 @@
...
++		if (!dev) {
++		     LOG(ERROR, "failed to get blktap devpath for %s: %s\n",
```
dpsmith: As I understand it, by switching from LOGD to LOG we have lost the domain information for which the disk add failed. Is there a reason why you chose to drop that info?

recipes-extended/xen/files/libxl-crypto-key-dir.patch
```diff
@@ -175,3 +113,66 @@ Index: xen-4.9.0/tools/xl/xl_parse.c
...
+Index: xen-4.9.1/tools/libxl/libxl_disk.c
+===================================================================
+--- xen-4.9.1.orig/tools/libxl/libxl_disk.c
++++ xen-4.9.1/tools/libxl/libxl_disk.c
+@@ -322,7 +322,7 @@ static void device_disk_add(libxl__egc *
+ 
+             case LIBXL_DISK_BACKEND_TAP:
+ 		rc = 0;
+-		dev = libxl__blktap_devpath(gc, disk->pdev_path, disk->format);
++		dev = libxl__blktap_devpath(gc, disk->pdev_path, disk->format, disk->crypto_key_dir);
```
dpsmith: 80 col please

recipes-extended/xen/files/libxl-blktap3-support.patch
mahantesh-ais: I'm not sure what's pedantic here, could you please be more specific ?

recipes-extended/xen/files/libxl-blktap3-support.patch
mahantesh-ais: Ack, I will fix it.

recipes-extended/xen/files/libxl-crypto-key-dir.patch
mahantesh-ais: Ack.

recipes-extended/xen/files/libxl-blktap3-support.patch
dpsmith: While the rc variable has been reused up to this point, every time it has been assigned a value, it is checked for a non-zero value and if true results in a jump to the out label. In your modifications you do not ever use rc (and thus not reliant on it actually being zero) except to assign it an error value, and the original author(s) already clear it for you at the end before going into the out labeled section. So IMHO setting rc to zero, when it is already expected to be zero, you not even dependent on it being zero, and your only operation to set the value to a non-zero value is a bit pedantic. For me, I am reviewing your patch in context of the function being modified and suddenly I see rc being zeroed then I have to stop and ask why did you do this, is there something here to be concerned with, if so, then why not fix up the rest of the function to provide the same assurances?

mahantesh-ais: rc being set to zero was not necessary, agreed. I will remove it but I didn't intentionally leave it there. There were issues interfacing with blktap3 because of which I had introduced it. After fix, I missed to remove it in my self review. Anyways, will fix it now.

mahantesh-ais:
Fixed the issues.

recipes-openxt/xenclient/xenclient-installer_git.bb
```diff
@@ -5,7 +5,7 @@ LIC_FILES_CHKSUM = "file://COPYING;md5=4641e94ec96f98fabc56ff9cc48be14b"
 PV = "0+git${SRCPV}"
 
 SRCREV = "${AUTOREV}"
-SRC_URI = "git://${OPENXT_GIT_MIRROR}/installer.git;protocol=${OPENXT_GIT_PROTOCOL};branch=${OPENXT_BRANCH} \
+SRC_URI = "git://github.com/mahantesh-ais/installer.git;protocol=${OPENXT_GIT_PROTOCOL};branch=pr \
```
cjp256: You'll need to drop this... for future reference, you can override this in your local.conf for dev/testing purposes.

mahantesh-ais: Ack, will fix it.

mahantesh-ais:
Fixed.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1280

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,703 @@
...
++    const char *type = NULL;
++    char *params, *devname = NULL;
++    struct tap_list tap;
++    int err = 0;
++    int flags = 0;
++
++    type = libxl__device_disk_string_of_format(format);
++    if (!(err = blktap_find(type, disk, &tap))) {
++        LOG(DEBUG, "found tapdisk\n");
++	devname = libxl__sprintf(gc, "/dev/xen/blktap-2/tapdev%d", tap.minor);
```
crogers1: Are we keeping this devpath as blktap-2 for a reason?

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,703 @@
...
++        if (devname)
++            return devname;
++		/* return 0; revert-back-changes*/
```
crogers1: Any idea what the /* revert-back-changes */ comments mean?  It's not immediately clear what they're for, so anyone reading through this patchset might be confused.

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,703 @@
...
++
++int libxl__device_destroy_tapdisk(libxl__gc *gc, const char *be_path)
++{
++    char *disk;
++    int err;
++    struct tap_list tap;
++
++	LOG(DEBUG, "destroying tapdisk %s\n", be_path);
++
++    disk = strchr(be_path, ':');
```
crogers1: What does the backend path look like now for blktap3?  I don't remember there being a colon in blktap2 paths, so if you could just post an example oh what that looks like, that would help me out a lot.  Thanks!

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,703 @@
...
++    if (!disk) {
++        LOG(ERROR, "Unable to parse params %s", be_path);
++        return ERROR_INVAL;
++    }
++
++    *disk++ = '\0';
```
crogers1: This seems really hackish to me.  I would rather see ```strncpy``` from 0 to the element pointed to by ```strchr``` into a new local variable that is the exact size we want (plus room for the '\0').  

Perhaps more importantly, 
```
disk = strchr(be_path, ';'); 
``` 
returns a pointer into ```be_path```, which you're then modifying with
```
*disk++ = '\0';
```
be_path is declared as const but it's just the pointer to first element of the char array that's const.  You're offset into the char* so const won't apply, and thus you're modifying the original be_path.  Doesn't look like that's supposed to be the intended behaviour of the function so in my mind, this absolutely needs to be fixed.

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,703 @@
...
++ * Local variables:
++ * mode: C
++ * c-basic-offset: 4
++ * indent-tabs-mode: nil
++ * End:
++ */
+Index: xen-4.9.1/tools/libxl/list.h
+===================================================================
+--- /dev/null
++++ xen-4.9.1/tools/libxl/list.h
```
crogers1: Libxl already has a list implementation, _libxl_list.h, it's autogenerated at compile time.  I would imagine that should be sufficient for whatever blktap3 needs it for.  No need to grab this from NetBSD.

Take a look at LIBXL_LIST_HEAD, LIBXL_LIST_HEAD_INITIALIZER, etc.

recipes-extended/xen/files/xl-shutdown-wait-for-domain-death.patch
```diff
@@ -0,0 +1,44 @@
...
+################################################################################
+Index: xen-4.9.0/tools/xl/xl_vmcontrol.c
+===================================================================
+--- xen-4.9.0.orig/tools/xl/xl_vmcontrol.c
++++ xen-4.9.0/tools/xl/xl_vmcontrol.c
+@@ -232,12 +232,6 @@ static void wait_for_domain_deaths(libxl
+             libxl_evdisable_domain_death(ctx, deathws[event->for_user]);
+             count++;
+             break;
+-        case LIBXL_EVENT_TYPE_DOMAIN_SHUTDOWN:
```
crogers1: DOMAIN_DEATH and SHUTDOWN_DOMAIN are two separate events, and they're both relevant.   I'm inclined not to remove this code block because libxl is supposed to handle the SHUTDOWN_DOMAIN event according to the reason code Xen sent back to it.  Completely ignoring the event seems like a bad choice to me.

What is the impact of leaving this code block in?

recipes-kernel/linux/4.9/patches/blktap2.patch
```diff
@@ -810,28 +810,13 @@ Index: linux-4.9.40/drivers/block/blktap/device.c
 +		goto fail;
 +	}
 +
-+	switch (req_op(rq)) {
-+		case REQ_OP_DISCARD:
```
crogers1: Are REQ_OP_DISCARD and REQ_OP_FLUSH no longer implemented?

rneilturner:
Is there a reason why we are keeping blktap2 around? For example, we are still carrying forward the Xen patches for blktap2. If we are going to make the move to blktap3 I would rather see the technical debt of blktap2 removed. In addition, is the blktap2 kernel patch servicing both? If we decide to keep blktap2 in our builds then please add a comment to the patch on forward supporting blktap3. Maybe consider renaming the patch to make it clearer.

recipes-extended/xen/files/libxl-blktap3-support.patch
mahantesh-ais: Just that the blktap3 repository from xenserver (https://github.com/xapi-project/blktap) has not changed it, new tapdev is written to /dev/xen/blktap-2/ devpath. Since the base for my changes is xenserver repository I didn't change it. Let me know if this is okay or needs to be changed.

recipes-extended/xen/files/libxl-blktap3-support.patch
mahantesh-ais: The libxl patch from xenserver (https://github.com/qiaohaiming/blktap3/blob/master/libxl.diff) used to return 0 instead of devname. This line is left over from base patch, anyways I will remove the line to avoid confusion, thanks.

recipes-extended/xen/files/libxl-blktap3-support.patch
mahantesh-ais: Actually this is a bit confusing, the arguments passed to the lilbxl__device_destroy_tapdisk function is tapdisk-params and not backend_path, if we see the base libxl patch https://github.com/qiaohaiming/blktap3/blob/master/libxl.diff , the second parameter in the function is named as be_path. It's confusing why it was named that way. I will fix this and change the name of the parameter to something relevant. Since, tapdisk-params look like "aio:/storage/isos/xc-tools.iso" the colon makes sense. As far as the backend path is concerned it looks similar to blktap2 ("/local/domain/0/backend/vbd3/8/5632")

recipes-extended/xen/files/libxl-blktap3-support.patch
mahantesh-ais: Yes, this needs to be fixed, I will do so by copying the element into a new local variable. As earlier this change was retained from the base libxl patch (https://github.com/qiaohaiming/blktap3/blob/master/libxl.diff)

recipes-extended/xen/files/xl-shutdown-wait-for-domain-death.patch
mahantesh-ais: Actually this code path is triggered only when wait option (-w) is passed with 'xl shutdown', in which case it should wait for the domain to go down (DOMAIN_DEATH).  As DOMAIN_DEATH always comes after SHUTDOWN_DOMAIN, the wait function returns  without actually waiting for the domain to go down completely. Yes, as you said DOMAIN_DEATH and SHUTDOWN_DOMAIN are two separate events but the libxl code which is supposed to handle these events is unchanged, please see tools/xl/xl_vmcontrol.c:(line 1094 to line 1169). The code here is just to make the wait return only when the domain shutdown is complete with its death.

recipes-kernel/linux/4.9/patches/blktap2.patch
mahantesh-ais: Yes, this was creating problems for blktap3. Need to be checked if these functionalities are required.

mahantesh-ais:
The blktap kernel patch is still used by blktap3, but the xen patches
are not needed anymore. As insisted, a commit will be added to remove them.

Note: Since I am currently on vacation, Jed offered to take over some of
the blktap3 work, so you may see commits from him.

recipes-extended/xen/files/libxl-blktap3-support.patch
mahantesh-ais: This was added as part of the base libxl patch, check here: https://github.com/qiaohaiming/blktap3/blob/master/libxl.diff, the list API's are conforming to the list.h added as part of this patch.  As suggested, the libxl list implementation might be sufficient but this will require changing the API's and list structure used for blktap3 support.

jean-edouard: Yeah, the lists are shared with the tap-ctl lib from blktap3.git, so they need to use the same list.h...
I suggest keeping stuff as is, but renaming list.h to tap-ctl-list.h to make things more obvious.
@crogers1 how does that sound to you?

mahantesh-ais: Fixed, no list file created anymore.

recipes-kernel/linux/4.9/patches/blktap2.patch
eric-ch: The fact you were seeing problems without removing this piece is troubling.
This seems wrong as `blktap_ring_submit_request()` is not changed and is therefor inconsistent with this.

Could you describe the problems you were having?

recipes-extended/xen/blktap3.bb
```diff
@@ -0,0 +1,31 @@
...
+
+INITSCRIPT_NAME = "tapback-daemon"
+INITSCRIPT_PARAMS = "defaults 61"
+
+do_configure_prepend() {
+	touch ${S}/EXTRAVERSION
+}
+
+do_install_append() {
+	rm -rf ${D}/usr/lib/systemd
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will fix it.

recipes-extended/xen/blktap3.bb
```diff
@@ -0,0 +1,31 @@
...
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM="file://LICENSE;md5=a9e8593dad23434929bc6218a17b5634"
+DEPENDS = "xen" 
+
+PV = "0+git${SRCPV}"
+
+SRCREV = "${AUTOREV}"
+SRC_URI = "git://${OPENXT_GIT_MIRROR}/blktap3.git;protocol=${OPENXT_GIT_PROTOCOL};branch=${OPENXT_BRANCH} \
+	file://tapback.initscript \
+"
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will fix it.

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,353 @@
...
++		LOG(DEBUG,"\nBLKTAP3_DEBUG: dev path = %s \n", dev);
++		if (!disk->script && disk->backend_domid == LIBXL_TOOLSTACK_DOMID) {
++		    int major, minor;
++		    if (!libxl__device_physdisk_major_minor(dev, &major, &minor)) {
++			LOG(DEBUG, "\nBLKTAP3_DEBUG: major:minor = %x:%x\n",major,minor);
++			flexarray_append_pair(back, "physical-device",
++				GCSPRINTF("%x:%x", major, minor));
++		    }
++		}
++		flexarray_append(back, "tapdisk-params");
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will fix it.

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,353 @@
...
+@@ -547,6 +548,12 @@ static int libxl__device_disk_from_xenst
+     }
+     libxl_string_to_backend(ctx, tmp, &(disk->backend));
+ 
++    tmp = libxl__xs_read(gc, XBT_NULL,
++                         GCSPRINTF("%s/tapdisk-params", libxl_path));
++    if (tmp && strcmp(tmp, "")) {
++    	disk->backend = LIBXL_DISK_BACKEND_TAP;
++    }
++
```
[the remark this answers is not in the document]
mahantesh-ais: This is actually a quirk required while dealing with iso hotswap where the device type needs to be "phy" and backend needs to be "LIBXL_DISK_BACKEND_TAP". Amending the `libxl_string_to_backend()` breaks the initial disk setup during domain creation. Yes, I will add a comment for this.

recipes-extended/xen/files/libxl-crypto-key-dir.patch
```diff
@@ -175,3 +113,58 @@ Index: xen-4.9.0/tools/xl/xl_parse.c
...
+     char *params, *devname = NULL;
+@@ -83,7 +84,12 @@ char *libxl__blktap_devpath(libxl__gc *g
+ 
+ 	/* TODO Should we worry about return codes other than ENOENT? */
+ 
+-    params = libxl__sprintf(gc, "%s:%s", type, disk);
++    if (!keydir || !strncmp(keydir, "", 1))
++	setenv("TAPDISK3_CRYPTO_KEYDIR", "/config/platform-crypto-keys", 1);
++    else
++	setenv("TAPDISK3_CRYPTO_KEYDIR", keydir, 1);
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will fix it.

mahantesh-ais:
Fixed.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1334

recipes-extended/xen/files/libxl-blktap3-support.patch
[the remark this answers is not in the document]
eric-ch: Ack, thanks.

recipes-extended/xen/blktap3/tapback.initscript
```diff
@@ -0,0 +1,59 @@
...
+
+case "$1" in
+    start)
+        start-stop-daemon -S -x "$PROG"
+	echo "OK"
+        ;;
+    stop)
+	echo -n "Stopping tapback: "
+        start-stop-daemon -K -x "$PROG"
+	echo "OK"
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will fix it. Thanks.

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,358 @@
...
++    tap_list_for_each_entry_safe(entry, next_t, &list) {
++
++        if (type && (!entry->type || strcmp(entry->type, type)))
++            continue;
++
++        if (path && (!entry->path || strcmp(entry->path, path)))
++            continue;
++
++        *tap = *entry;
++	tap->type = tap->path = NULL;
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will fix it. Thanks.

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,358 @@
...
++    struct tap_list tap;
++    int err = 0;
++    int flags = 0;
++
++    type = libxl__device_disk_string_of_format(format);
++
++    err = blktap_find(type, disk, &tap);
++    if (!err) {
++        LOG(DEBUG, "found tapdisk\n");
++	devname = libxl__sprintf(gc, "/dev/xen/blktap-2/tapdev%d", tap.minor);
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will fix it. Thanks.

recipes-extended/xen/files/libxl-blktap3-support.patch
```diff
@@ -0,0 +1,358 @@
...
++++ xen-4.9.1/tools/libxl/libxl_device.c
+@@ -297,11 +297,6 @@ static int disk_try_backend(disk_try_bac
+ 
+         /* Note: keep going if the disk is a cdrom. tapdisk can handle them as "raw" */
+ 
+-        if (!libxl__blktap_enabled(a->gc)) {
+-            LOG(DEBUG, "Disk vdev=%s, backend tap unsuitable because blktap "
+-                       "not available", a->disk->vdev);
+-            return 0;
+-        }
```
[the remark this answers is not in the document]
mahantesh-ais: This check is not required now as `libxl__blktap_enabled` always returns '1', because in blktap3 blkfront directly contacts tapdisk (unlike blktap2 where the path is blkfront --> blkback -->blktap2-kernel-module --> tapdisk). In blktap2 this check made sense as blktap2-kernel-module was involved in the path. However, this check is required by tapdisk/tap-ctl processes as blktap2-kernel-module is still used here and `tap_ctl_check_blktap` is being done in this place as required.

P.S: This is in accordance with the base patch for blktap3, given here : https://github.com/qiaohaiming/blktap3/blob/master/libxl.diff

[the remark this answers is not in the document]
mahantesh-ais: Although the above lines are redundant, I will revert back the changes for the sake of blktap2 retro-compatibility support.
The implementation searching for "blktap/control" instead of "blktap-control" is already fixed, I think you might have looked at older code.
Thanks.

mahantesh-ais:
Fixed.
Built here : http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1352

rneilturner:
+1 I have been testing these changes with no issues.

eric-ch:
Will merge on Monday unless someone has more concerns.
This should let some time for people to catch up.

rneilturner:
I'm seeing some problems when rebooting PV guests. I get an error in the tapback frontend failing to connect to the shared ring. @mahantesh-ais are you seeing any issues with rebooting PV guests such as the UIVM with blktap3?

eric-ch:
@rneilturner I confirm, same behavior here, which I overlooked earlier. It seems limited to PV guests on build 1352 as well.

From the logs:
```
tapback[1884]: tapback.c:447 slave tapback daemon started, only serving domain 3
tapdisk[1877]: VBD 2 got disk info: sectors=679936 sector size=512, info=0
tapdisk: tap-err:tapdisk_server_init: Failed to initialize low memory handler: No such file or directory
[...]
tapback[1884]: frontend.c:216 51713 front-end supports persistent grants but we don't
tapback[1884]: couldn't connect to /var/run/blktap-control/ctl1877: No such file or directory
tapback[1884]: frontend.c:235 51713 tapdisk[1877] failed to connect to the shared ring: No such file or directory
tapback[1884]: tapback.c:144 failed to process XenStore watch on /local/domain/3/device/vbd/51713/state: No such file or directory
tapback[1884]: frontend.c:216 51714 front-end supports persistent grants but we don't
tapdisk[1901]: connecting VBD 3 domid=3, devid=51714, pool (null), evt 10, poll duration 0, poll idle threshold 0
[...]
xenconsoled: uivm (3): [    1.093952] blkfront: xvda2: barrier: enabled; persistent grants: disabled; indirect descriptors: disabled;
tapback[1884]: frontend.c:216 51715 front-end supports persistent grants but we don't
tapdisk[1917]: connecting VBD 4 domid=3, devid=51715, pool (null), evt 11, poll duration 0, poll idle threshold 0
tapdisk[1917]: ring 0x96948b0 connected
xenconsoled: uivm (3): [    1.095795] blkfront: xvda3: barrier: enabled; persistent grants: disabled; indirect descriptors: disabled;
xenconsoled: uivm (3): [    1.120046] clocksource: tsc: mask: 0xffffffffffffffff max_cycles: 0x2e03d169996, max_idle_ns: 440795324381 ns
xenconsoled: uivm (3): [    6.150150] xenbus_probe_frontend: Waiting for devices to initialise: 25s...20s...15s...10s...5s...0s...
xenconsoled: uivm (3): [   36.180294] 235s...230s...225s...220s...[   55.740264] random: crng init done
xenconsoled: uivm (3): [   56.200064] 215s...210s...205s...200s...195s...190s...185s...180s...175s...170s...165s...160s...155s...150s...145s...140s...135s...130s...125s...120s...115s...110s...105s...100s...95s...90s...85s...80s...75s...70s...65s...60s...55s...50s...45s...40s...35s...30s...25s...20s...15s...10s...5s...0s...


xenconsoled: uivm (3): [  271.140539] xenbus_probe_frontend: Device with no driver: device/vfb/0
xenconsoled: uivm (3): [  271.140546] xenbus_probe_frontend: Device with no driver: device/v4v/0
xenconsoled: uivm (3): [  271.140708] VFS: Cannot open root device "xvda1" or unknown-block(0,0): error -6
xenconsoled: uivm (3): [  271.140720] Please append a correct "root=" boot option; here are the available partitions:
xenconsoled: uivm (3): [  271.140732] 0100            4096 ram0 [  271.140737]  (driver?)
xenconsoled: uivm (3): [  271.140741] 0101            4096 ram1 [  271.140745]  (driver?)
xenconsoled: uivm (3): [  271.140749] 0102            4096 ram2 [  271.140753]  (driver?)
xenconsoled: uivm (3): [  271.140757] 0103            4096 ram3 [  271.140760]  (driver?)
xenconsoled: uivm (3): [  271.140764] 0104            4096 ram4 [  271.140768]  (driver?)
xenconsoled: uivm (3): [  271.140772] 0105            4096 ram5 [  271.140776]  (driver?)
xenconsoled: uivm (3): [  271.140780] 0106            4096 ram6 [  271.140784]  (driver?)
xenconsoled: uivm (3): [  271.140788] 0107            4096 ram7 [  271.140792]  (driver?)
xenconsoled: uivm (3): [  271.140796] 0108            4096 ram8 [  271.140815]  (driver?)
xenconsoled: uivm (3): [  271.140819] 0109            4096 ram9 [  271.140837]  (driver?)
xenconsoled: uivm (3): [  271.140841] 010a            4096 ram10 [  271.140845]  (driver?)
xenconsoled: uivm (3): [  271.140853] 010b            4096 ram11 [  271.140856]  (driver?)
xenconsoled: uivm (3): [  271.140860] 010c            4096 ram12 [  271.140864]  (driver?)
xenconsoled: uivm (3): [  271.140868] 010d            4096 ram13 [  271.140872]  (driver?)
xenconsoled: uivm (3): [  271.140887] 010e            4096 ram14 [  271.140891]  (driver?)
xenconsoled: uivm (3): [  271.140894] 010f            4096 ram15 [  271.140913]  (driver?)
xenconsoled: uivm (3): [  271.140917] ca02          262144 xvda2 [  271.140921]  driver: vbd
xenconsoled: uivm (3): [  271.140926] ca03           10240 xvda3 [  271.140930]  driver: vbd
xenconsoled: uivm (3): [  271.140948] Kernel panic - not syncing: VFS: Unable to mount root fs on unknown-block(0,0)
```

mahantesh-ais:
Reboot of PV domains works perfectly on my test machine 'xec-vm -n uivm reboot' OR 'xec-vm -n Network reboot'.

rneilturner:
The error @eric-ch is seeing is the same one I have also encountered with PV reboots.

mahantesh-ais:
Will try this on different machines and get back. Thanks.

rneilturner:
The machine I'm running with is a Dell 7040 with 32GB of RAM.

mahantesh-ais:
I'm able to reproduce this bug on Dell 7040 but on Dell E6540 uivm reboot works fine. However, this issue is not introduced in this PR(blktap3) as the problem exists in master branch (was able to reproduce the bug using latest master build 1353 on Dell 7040). I created a ticket 'OXT-1270' to track this issue and suggest fixing this in a separate PR. Thanks.

jean-edouard:
+1, the bug also repro-es on my Dell E7480 using a regular master build

eric-ch:
@mahantesh-ais +1 for fixing OXT-1270 in a separate PR. Will merge tomorrow unless someone speaks out.

richpersaud:
We discussed the ability to build either blktap2 or blktap3, until we have confidence in switching.

After this PR is merged, will blktap2 or blktap3 be the default?  Is there documentation on changing between the two?

eric-ch:
@richpersaud Although I cannot find a report of the discussion, I recall someone said it could be a possibility. This PR makes blktap3 the default though.

Supporting both blktap3 and blktap2 is not trivial and doubles the testing effort, which is already a (... the) bottleneck for any decision. If someone lacks confidence in switching to blktap3, please speak out. This PR has been out there far longer than most, more invasive, changes and went through quite a long review process, during which people tested it revealing prior unknown bugs with blktap2 and its use in OpenXT.

IMHO: -1 for supporting both blktap2 (which is phased out of upstream anyway) and blktap3.

dpsmith:
@eric-ch It was I that raised the issue on the January call and we had a short offline discussion, the email chain started on 18 January. You raised only one concern that, that as I highlighted with a small sample, it would not be difficult to make it a build time opt-in configuration (i.e. adding blktap2 to DISTRO_FEATURES in local.conf). If set, then the build to use blktap2 otherwise, by default, blktap3 would be built and used.

As I stressed on the call when making the suggestion, there is no expectation that blktap2 configurations would be tested by QA. Right now we are hard wired for blktap2 and this change set is going to persist that hard wiring for blktap3. A little forethought here can save pain down the road.

eric-ch:
@dpsmith Ack, so there are interests to add a build-time switch between blktap2 and blktap3. With your following argument, lets take QA out of the picture and assume this would leave blktap2 as an experimental fall-back, which seems fair. I take it you -1 this PR going in as it is.

I still say making a DISTRO_FEATURES for blktap3 is not that straight-forward (aka "not trivial"). It would likely require splitting recipes to avoid hacking around patch-queues being partially applied and moving changes to repository as optional patches in the layer, etc, to make it a case of building a set of recipes instead of another. Which in turns makes the layer more complicated and harder to maintain.

With blktap2 not to be supported, I do not see the point.

richpersaud:
- blktap2 is tested and proven on OpenXT
- blktap3 is not yet equivalently tested on OpenXT
- we don't know who, if anyone, is using the github-hosted blktap3
- previous versions of XenServer may be using blktap3 with patches not on github (see XS source CD)
- we don't yet have robust automated QA coverage for OpenXT storage

If we were to apply standard software development practices, where "new" technology is considered experimental until proven otherwise, the "experimental" code base here is blktap3 and not blktap2.  Under normal circumstances, blktap2 would be the default and blktap3 would be the optional storage layer.  In the interest of increasing manual test coverage of blktap3, I would support blktap3 being the default storage layer.  If an OpenXT user wants a tested storage layer, the only option today is blktap2.

When we have automated storage tests for OpenXT, with equivalent results on blktap2 and blktap3, across a range of storage devices, we can deprecate blktap2.

recipes-extended/xen/files/libxl-iso-hotswap.patch
```diff
@@ -236,28 +230,28 @@ Index: xen-4.9.0/tools/libxl/libxl_internal.h
  _hidden char *libxl__device_libxl_path(libxl__gc *gc, libxl__device *device);
  _hidden int libxl__parse_backend_path(libxl__gc *gc, const char *path,
                                        libxl__device *dev);
-@@ -1765,6 +1766,9 @@ _hidden char *libxl__blktap_devpath(libx
-                                     libxl_disk_format format,
-                                     char *keydir);
+@@ -1764,6 +1765,9 @@ _hidden char *libxl__blktap_devpath(libx
+                                     const char *disk,
+                                     libxl_disk_format format);
```
dpsmith: Why did you re-order the parameters and rename one? seems like unnecessary churn to me.

dpsmith:
@eric-ch I just stepped through the patches (obvious from the comment I just made). I did not see any non-blktap related patches being altered as the result of the removal of blktap2 patches, alteration of existing blktap2 related patching to accommodate blktap3, and the introduction of blktap3. To me that means that you should be able to select between the older version (blktap2) of the patches and the new version (blktap3) without causing patch queue failures in the non-blktap related patches. Yes, it means that for what I believe was three or four patches, there would be two versions, one for each of the blktap versions, but the only ones being maintained would be the blktap3 related ones. Then its just a series of lines like the following in the SRC_URI definition.
`
${@base_contains('DISTRO_FEATURES', 'blktap2', 'blktap2.patch', 'blktap3.patch',d)}
`

As for nack-ing the PR, unlike January when I had some spare cycles to spare to help, I no longer have any spare cycles to help. Therefore I do not feel it would be right for me to hold up the PR from going in, but am just making my concern known and mention that a new PR might come that re-introduces blktap2 as optional feature as I have highlighted.

eric-ch:
@richpersaud As I tried to say, since bug were uncovered in the current blktap2 implementation while reviewing blktap3. It appears we lack a good point of reference. Also this is master, downstream testing may uncover problems in stable-* branches, in which case reverting would still be possible (admittedly less convenient than a switch).

@dpsmith (`base_contains` is becoming "deprecated", but yes that is how I would have done it too ;)...) plus duplication of configuration and scripts files (like `vhdutils`) in different recipes. Again I am not saying this is impossible. A later PR that re-introduce a blktap2 `DISTRO_FEATURES` I would have no problem with either. Even if, assuming your reviewed comment gets addressed too, you are +/-0 vote on this, as I presume you do not +1, your concerns are shared by @richpersaud .

So this is not getting merged in these conditions.

richpersaud:
When a fundamental component of the system is changed (like the toolstack or storage), some issues detected post-merge may be caused by that change, or may have been present all along but never tested.  We saw that last year with the toolstack. If testers can have access to builds where the only difference is blktap2 vs. blktap3, it will be easier to isolate the root cause.

We need a list of test cases, manual or otherwise, including hardware/storage specs, which have been evaluated for blktap3.  This can be put on the wiki and linked from the JIRA ticket.  Over time, this list can be extended with automated and additional manual tests.  But for now, we can document what has already been tested.

eric-ch:
> We need a list of test cases, manual or otherwise, including hardware/storage specs, which have been evaluated for blktap3

@richpersaud I cannot recall any previous contribution that went through that sort of requirements. Changing the entire submission process for this PR seems arbitrary. Still the requirement for a build-time switch between blktap2 and blktap3 is well noted. Unless having a later PR introducing it is acceptable, we still have not reached a consensus on that topic.

richpersaud:
Storage is a special case.  If there's any doubt, ask someone who has worked in a storage company.

The goal here is to get blktap3 integrated into OpenXT with at least equivalent functionality and stability relative to blktap2, ideally better  I've listed some benefits of maintaining access to the stable and proven blktap2, which would actually *reduce* pressure on blktap3 and make it easier to compare behaviors.  Whether pre or post merge, blktap3 will need to be tested.  

If a list of tests is started now, it can be improved later.  If we want to merge high-risk storage code without a list of tests, and without a fallback option to blktap2, we can do that, but (a) the tests still have to be done at some point, (b) if any storage issues are found, there's no fallback and other work could be blocked.

I will leave it to others to evaluate the effort of enabling a build-time switch vs. committing code with/without documented tests.  The less data on code quality, the more benefits in having a fallback comparison.  The more data (either from tests or whitebox knowledge) on code quality, the less risk incurred in committing code without a fallback.

recipes-extended/xen/files/libxl-iso-hotswap.patch
mahantesh-ais: I didn't, I'm not sure why git is showing it this way. May be because of the review fixes and patch updates. You can check all the patches here `https://github.com/mahantesh-ais/xenclient-oe/tree/prv4/`. The order and naming has always been 
`(libxl__gc *gc,`
 `const char *disk,`
 `libxl_disk_format format,`
 `char *keydir);`
and its the same now.

eric-ch:
The last commit (and relevant ones: https://github.com/OpenXT/installer/pull/62 and https://github.com/OpenXT/manager/pull/107) introduce the demanded fallback switch. Due to differences between blktap2 & blktap3 handling by the toolstack, it requires to take changes made in https://github.com/OpenXT/toolstack/pull/33 and apply them only when using blktap3 in the layer as a patch.

This makes the list of dependencies:
- https://github.com/OpenXT/installer/pull/62
- https://github.com/OpenXT/manager/pull/107
- https://github.com/OpenXT/xsm-policy/pull/17
- https://github.com/OpenXT/blktap3/pull/1

Building:
- Default configuration: [$1383](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1383/steps/Build/logs/stdio)
- Blktap2 fallback: [$1382](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1382/steps/Build/logs/stdio)

With these latest changes, this PR add _blktap3_ as the default storage back-end.
To use _blktap2_, add `blktap2`to `DISTRO_FEATURES` (either in your _Bitbake_ `local.conf`, or in a specific distribution configuration file, like [openxt-main.conf](https://github.com/OpenXT/xenclient-oe/blob/master/conf/distro/openxt-main.conf))

jean-edouard:
LGTM, merging around end of day.

PR #817, opened 2018-01-02
OXT-1265: rpcgen: Aggregate common configuration in a bbclass

eric-ch:
The IDL used in a variety of OpenXT's project require some configuration at build time which is currently done in the top-level local.conf and further in bitbake recipes includes, or even duplicated in relevant recipes.

This is confusing and causes problems with recent versions of OE as it also relies on a common shared sysroot between recipes (which is no longer the case in Pyro, each recipe has its own sysroot, built with its dependencies).

Since the extent of the configurations required is different depending on the language used, I propose a set of bbclasses:
- `xc-rpcgen-c-1.0.bbclass`
- `xc-rpcgen-ocaml-1.0.bbclass`
- `xc-rpcgen-haskell-1.0.bbclass`

Each will include generic configuration for each language (e.g. `xc-rpcgen-c.inc`), that in its turn will include the main `xc-rpcgen-common.inc` configuration with the definition commonly shared by all languages and all versions: paths to IDL files definitions, templates, and build `DEPENDS`.


eric-ch:
Depends on:
- https://github.com/OpenXT/openxt/pull/281
- https://github.com/OpenXT/xctools/pull/42
- https://github.com/OpenXT/vusb-daemon/pull/16
- https://github.com/OpenXT/libxcdbus/pull/1
- https://github.com/OpenXT/input/pull/26

recipes-openxt/libxcdbus/libxcdbus_git.bb
```diff
@@ -1,16 +1,16 @@
...
 SRCREV = "${AUTOREV}"
 SRC_URI = "git://${OPENXT_GIT_MIRROR}/libxcdbus.git;protocol=${OPENXT_GIT_PROTOCOL};branch=${OPENXT_BRANCH}"
 
-EXTRA_OECONF += "--with-idldir=${STAGING_IDLDIR}"
-
 S = "${WORKDIR}/git"
 
-inherit autotools-brokensep pkgconfig lib_package xenclient
-
+inherit autotools-brokensep pkgconfig lib_package xc-rpcgen-c-1.0
```
jean-edouard: Why remove `inherit xenclient` here but not in libicbinn?

eric-ch: I tend to remove it as I go and apparently missed it for libicbinn recipe. I will amend that.
The only thing provided by `xenclient.bbclass` is to cripple `do_rm_work()`, which should be done differently, and is not done by default anyway.

eric-ch: Actually I seem to have left it in general... So removing it should probably be another PR all together.

jean-edouard:
Merging soon

PR #818, opened 2018-01-15
OXT-1265: Refactor rpcgen classes to standard

eric-ch:
Introduce `xc-rpcgen.conf` to handle project not using `bbclass` but installing IDL files. Also rename variables to comply with OE usual naming conventions. Lower-case `[a-z]+dir` variables are commonly used for install paths, whereas upper-case `STAGING_.+` point to path of files installed in the sysroot.

Also use the `xc-rpcgen-ocaml-1.0.bbclass` for db-tools as it was not part of the first PR.


dpsmith:
I would disagree with the refactoring you have done here. The original work IMHO was closer to following OE convention. The STAGING paths are good but no need to create a new peer level config file for them. My suggestion is to move them under the distro configuration (because this is a distro configuration), and you can keep as a standalone .conf file but have it included by openxt-main.conf, not by the individual recipes. I know I missed commenting on the PR that introduced it, but with the .inc files in the classes directory, not sure why you have them broken up into a common and specific and naming them .inc. I see no reason why they couldn't just be combined into a single xc-rpcgen.bbclass file. If you really wanted to get fancy, you could even add a do_rpcgen task after do_config that would iterate through a variable exported by the recipe that contains the name of files that recipe needs to have generated. A second recipe variable would declare the language that the bindings should be generated in, which could be multiple languages.

eric-ch:
@dpsmith thanks for the review. I do not mind re-working the previous PR through this one, on the contrary. Regarding your points:
- Moving the xc-rpcgen.conf to distro configuration: Agreed, for some reason I did not think of it that way and tried to do a [bitbake.conf](https://github.com/openembedded/openembedded-core/blob/master/meta/conf/bitbake.conf)
- The `.inc` files in `classes`: I tried to stay consistent and avoid redundancy. I realize there are no `.inc` files in `classes` directories so that was weird for sure. The C version of `rpcgen` that requires `EXTRA_CONF`, which should not be inherited with Haskell or OCaml use of `rpcgen`. Also `rpcgen` defines a version number, that should be taken into account, yet a lot of things will be common to different versions, hence include files. Anyway, with the paths in distro configuration, the classes are mostly useless and over-complicated, they might as well be removed.
- Add a `do_rpcgen`: Each language uses the xc-rpcgen binary in a different way, one of which does in the project's build system. Doing it this way, I would advise removing the xc-rpcgen from the C projects build-system, or add it to the ocaml/haskell ones for consistency.

Will update the PR soon.

eric-ch:
> it is equally fair to wait to see how a new IDL generation would work before adjusting recipes.

Although I agree on reducing the recipe churn, which is quite significant in this case, I would prefer taking that last advice, at least for now :).

eric-ch:
Building [#1293](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1293)

jean-edouard:
LGTM, merging soon.

PR #821, opened 2018-01-19
[linux] Adjust Hypercall id for txt evtlog

crogers1:
Xen-4.9 added a new HYPERCALL_dm_op that uses the current id
  for the txt evtlog hypercall.  This commit adjusts the patch
  to account for that new hypercall.


crogers1:
@eric-ch @jean-edouard, I'll be on vacation next week so I won't be around to comment on this PR, but @rneilturner can answer questions for me.

eric-ch:
Building [#1296](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1296)

LGTM.
That will blow up otherwise since the hypercall is `42` in the current Xen 4.9 patch-queue. Good catch.

eric-ch:
Will merge soon.

PR #822, opened 2018-02-19
linux-input: upgrade to stop using broken sourceforge repo

jean-edouard:


eric-ch:
Building [#1347](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1347)

PR #826, opened 2018-03-05
Amend mirror.openxt.org layout changes and add new mirror.

eric-ch:
~~- python-daemon still uses the deprecated http:// interfaces, switch to https as other recipes did.~~
- mirror.openxt.org changed its layout sometime recently. So attend for it. Since ainfosec.com also provide a mirror, add it for redundancy.

### DIFF
--- a/recipes-openxt/acms/acms.bb
+++ b/recipes-openxt/acms/acms.bb
@@ -43,18 +43,21 @@ LIC_FILES_CHKSUM = "file://GM45_GS45_PM45-SINIT_51/license.txt;md5=60d123634e0b9
                     file://7th_gen_i5_i7-SINIT_74/license.txt;md5=4897766beede80773ddd162b2724a8bb \
 "
 
-SRC_URI = "${OPENXT_MIRROR}/GM45_GS45_PM45-SINIT_51.zip;name=gm45 \
-           ${OPENXT_MIRROR}/Q45_Q43-SINIT_51.zip;name=q45 \
-           ${OPENXT_MIRROR}/Q35-SINIT_51.zip;name=q35 \
-           ${OPENXT_MIRROR}/i5_i7_DUAL-SINIT_51.zip;name=i5 \
-           ${OPENXT_MIRROR}/i7_QUAD-SINIT_51.zip;name=i7 \
-           ${OPENXT_MIRROR}/Xeon-5600-3500-SINIT-v1.1.zip;name=xeon_5600 \
-           ${OPENXT_MIRROR}/Xeon-E7-8800-4800-2800-SINIT-v1.1.zip;name=xeon_e7 \
-           ${OPENXT_MIRROR}/3rd-gen-i5-i7-sinit-67.zip;name=ivb_snb \
-           ${OPENXT_MIRROR}/4th-gen-i5-i7-sinit-75.zip;name=hsw \
-           ${OPENXT_MIRROR}/5th_gen_i5_i7-SINIT_79.zip;name=bdw \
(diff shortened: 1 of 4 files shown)


jean-edouard:
+1 for python-daemon, -1 for the mirror changes.
Since mirror.openxt.org is unreliable we should switch to another one in:
https://github.com/OpenXT/openxt/blob/master/example-config#L37
I can PR that.

dpsmith:
I'm sorry, but what is your criteria for declaring mirror.openxt.org as "unreliable"?

jean-edouard:
The file layout randomly changes there, without any notice, that I know of at least.

dpsmith:
How many times has that happen over the last four years?

jean-edouard:
No idea

eric-ch:
Taking python-daemon in a different PR.

eric-ch:
@dpsmith I found two documented occurrences[[1](https://groups.google.com/d/msg/openxt/5Wuz99AHss0/HatjPDvxAgAJ)][[2](https://groups.google.com/forum/#!msg/openxt/NMoLqEEl0mc/wLtU47cgBgAJ)], this one being the third.

Adding a mirror (regardless of the priority) would at least offer some redundancy... And there are sourceforge fetcher failures to keep things into perspective :).

@jean-edouard 's alternative: https://github.com/OpenXT/openxt/pull/284.

recipes-openxt/acms/acms.bb
```diff
@@ -43,18 +43,21 @@ LIC_FILES_CHKSUM = "file://GM45_GS45_PM45-SINIT_51/license.txt;md5=60d123634e0b9
...
-           ${OPENXT_MIRROR}/Xeon-E7-8800-4800-2800-SINIT-v1.1.zip;name=xeon_e7 \
-           ${OPENXT_MIRROR}/3rd-gen-i5-i7-sinit-67.zip;name=ivb_snb \
-           ${OPENXT_MIRROR}/4th-gen-i5-i7-sinit-75.zip;name=hsw \
-           ${OPENXT_MIRROR}/5th_gen_i5_i7-SINIT_79.zip;name=bdw \
-           ${OPENXT_MIRROR}/6th_gen_i5_i7-SINIT_71.zip;name=skl \
-           ${OPENXT_MIRROR}/7th_gen_i5_i7-SINIT_74.zip;name=kbl \
+MIRRORS_append += " \
+    http://.*/.*    https://openxt.ainfosec.com/mirror/ \
+"
+SRC_URI = "${OPENXT_MIRROR}/openxt/GM45_GS45_PM45-SINIT_51.zip;name=gm45 \
```
jandryuk: Should "/openxt" just be added to the OPENXT_MIRROR definition?  I don't know how MIRRORS_append works, but https://openxt.ainfosec.com/mirror/openxt/GM45_GS45_PM45-SINIT_51.zip does not exist - it does without the extra openxt directory.

eric-ch: @jandryuk [`MIRRORS`](https://www.yoctoproject.org/docs/2.4.2/ref-manual/ref-manual.html#var-PREMIRRORS) will replace matching `http://.*/.*` (so expanding `OPENXT_MIRROR` -> `http://mirror.openxt.org/openxt`) with `https://openxt.ainfosec.com/mirror` (at least that is the behavior I have seen so far) and fetch, for example, https://openxt.ainfosec.com/mirror/GM45_GS45_PM45-SINIT_51.zip. We could add `/openxt` to `OPENXT_MIRROR` though. I believe the confusion came because some acms where initially (presumed?) to live in http://mirror.openxt.org/acms.

PR #828, opened 2018-03-07
OXT-1280: OpenXT UEFI support

tklengyel:
In this PR OpenXT is extended to allow booting under UEFI while maintaining current security properties.

## Changes to Measured Launch
Booting OpenXT under UEFI introduces significant changes to the Measured Launch system of OpenXT by switching to the use of SRTM PCRs. Performing DRTM with TXT when booting under UEFI is not supported but can be implemented at a later date. OpenXT under UEFI relies on the shim EFI application to perform necessary measurements of critical boot components of OpenXT. OpenXT's static PCR list is extended to include PCR4, PCR5, PCR7 and PCR8. Specific to OpenXT's context, PCR4 holds the measurements of the shim, Xen and the dom0 kernel while PCR8 holds the measurements of openxt.cfg, the dom0 initrd image and the XSM policy. Any change in these components, selecting a boot entry other then the one used when Sealing took place and/or booting any application before the shim will result in the Measured Launch system tripping.

## Changes to Xen
When booting OpenXT under UEFI no boot manager (ie. GRUB) is being utilized. To provide the end-user with a boot menu, the UEFI firmware's EFI_LOAD_OPTION specification is being used. Xen is extended to properly parse EFI_LOAD_OPTIONs, such that the appropriate boot configuration is selected from the openxt.cfg config file. Note that selecting a boot entry other then the default after the system has been sealed will trip Measured Launch, thus its use should be limited to debugging. It would however be possible to install a config file with boot options containing tweaks for different hardware platforms and then adjusting which boot entry is considered to be the default before sealing happens.

Xen's EFI loader is extended to hardcode the config file name "openxt.cfg" for convenience.

Xen's EFI loader interaction with the shim is changed to perform measurements of all critical boot components. As this PR doesn't yet add support for SecureBoot, the dom0 kernel is switched to just being measured. Once SecureBoot support is added, the dom0 kernel can be both measured and verified using the shim Verify ABI.

Xen's EFI loader is changed such that if the shim protocol is not loaded (in case xen.efi is launched directly) an error message is printed and booting is halted. This is for convenience only, launching xen.efi directly would also trip the Measured Launch system.

## Changes to the dom0 kernel
Several EFI related options are enabled necessary for dom0 to be able to use the EFI framebuffer interface.

## Changes to forward sealing
Forward sealing allows OTA upgrade to seal against expected PCR values after an upgrade is applied. The forward sealing scripts are updated to properly distinguish whether SRTM or DRTM PCRs need to be re-calculated. Calculation for the SRTM PCRs (PCR4 and PCR8) is implemented, the other SRTM PCRs should not change during an upgrade.

## Changes to the installer
The OpenXT installer kernel is changed to be 64-bit, which is necessary for the kernel to interact with the 64-bit UEFI firmware when installing boot entries. The Linux kernel prohibits interaction between a 32-bit kernel and a 64-bit firmware. For this a new MACHINE_TYPE is introduced and the installer kernel is compiled with all kernel-modules built-in.

The ISO image that is being created is also updated to allow booting the same ISO on both legacy and UEFI systems. The ISO creation mechanism is modeled after the Debian installer, where the ISO is created with a FAT image being placed in the ISO9660 partition that acts as the UEFI ESP. This FAT image only contains a GRUB2 bootloader compiled for UEFI with an early-config embedded to locate the rest of the installer files on the ISO9660 partition. Using syslinux under UEFI is not possible.

Note that while the installer boots with Xen loaded, the 32-bit Xen toolstack can't communicate with the 64-bit interface. Xen is kept in-place simply to aid the future upgrade to 64-bit.

## Changes to the partition table

UEFI requires the hard-drive to be partioned using GPT. Upgrading from MBR partitioning to GPT is not supported. A fresh install is necessary and no data is preserved when this change is performed. The partitioning layout with GPT includes the EFI System Partition (ESP), a 100MB FAT32 formatted partition, and the BIOS Boot Partition (BBP), a 1MB unformatted partition that will be used by GRUB when OpenXT is installed in legacy-mode on a GPT partitioned disk.

## EFI Removable Path

OpenXT installs the shim to the EFI Removable Path in order to ensure maximum compatibility with firmwares where adding boot options using efibootmgr is not possible.

## Known issues

Upgrade form MBR to GPT is not and will not be supported. 

To upgrade from a previous installation the recommended path is to perform upgrade in two steps, first with an upgrade image containing only commit 7f3c1df, followed by a second upgrade image which contains the rest of the commits in this PR. This is necessary as the set of PCRs used for unsealing is extended in this PR and thus the forward sealing scripts need to get updated first.

## Related PRs

https://github.com/OpenXT/openxt/pull/285
https://github.com/OpenXT/installer/pull/63


richpersaud:
Please include JIRA ticket number on each PR.

tklengyel:
@richpersaud I don't see any open JIRA tickets related to UEFI and I don't have access to JIRA to open new ones.

richpersaud:
Tamas, the usual process for requesting a JIRA account is to send an email to the mailing list.  Last night, after you posted this comment, I sent a JIRA invitation to your gmail account.  If you have other questions about JIRA, please post to the mailing list.

tklengyel:
@richpersaud Thanks, I received the invite and I've opened the ticket for UEFI.

recipes-bsp/grub/grub-efi-cross_2.02.bb
```diff
@@ -0,0 +1,103 @@
...
+BUILD_LDFLAGS = ""
+
+# ldm.c:114:7: error: trampoline generated for nested function 'hook' [-Werror=trampolines]
+# and many other places in the grub code when compiled with some native gcc compilers (specifically, gentoo)
+CFLAGS_append_class-native = " -Wno-error=trampolines"
+
+EXTRA_OECONF = "--with-platform=efi \
+		--target=x86_64 \
+                --disable-grub-mkfont \
+                --program-prefix="" \
```
jandryuk: Mix of spaces and tabs.

recipes-core/base-files/base-files_3.0.14.bb
```diff
@@ -147,6 +148,15 @@ do_install_append_xenclient-dom0 () {
         install -m 0644 ${WORKDIR}/fstab.xenclient-dom0 ${D}${sysconfdir}/fstab
 }
 
+do_install_append_openxt-installer () {
+	install -m 0755 -d ${D}/config
+	install -m 0755 -d ${D}/var/cores
+	install -m 0755 -d ${D}/var/log
+	install -m 0755 -d ${D}/storage
+        install -m 0755 -d ${D}/var/lib/dbus
+        install -m 0644 ${WORKDIR}/fstab.openxt-installer ${D}${sysconfdir}/fstab
```
jandryuk: Mix of spaces and Tabs.

recipes-core/images/xenclient-installer-image.bb
```diff
@@ -38,7 +39,6 @@ IMAGE_INSTALL = "\
     packagegroup-base \
     packagegroup-xenclient-common \
     packagegroup-xenclient-installer \
-    kernel-module-e1000e \
```
jandryuk: Is this now built in?

recipes-core/initscripts/initscripts-1.0/openxt-installer/urandom
```diff
@@ -0,0 +1,61 @@
+#! /bin/sh
```
jandryuk: Saving the random seed doesn't make sense for an ephemeral installer image, so maybe remove this init script?

recipes-core/packagegroups/packagegroup-xenclient-installer.bb
```diff
@@ -22,6 +23,7 @@ RDEPENDS_${PN} = " \
     xen-libxlutil \
     xen-xenstat \
     xen-xl \
+    xen-xenstored-ocaml \
```
jandryuk: Why xenstored now needed?

recipes-extended/xen/files/efi-require-shim.patch
```diff
@@ -0,0 +1,30 @@
...
+ 
++        if ( !shim_lock )
++        {
++            PrintStr(L"No shim found, booting xen.efi directly with OpenXT is not supported.");
++            efi_bs->Stall(10000000);
++            blexit(L"No shim found");
++        }
++
+         if ( shim_lock &&
+             (status = shim_lock->Measure(cfg.ptr, cfg.size, 4)) != EFI_SUCCESS )
```
jandryuk: Would it make sense remove all the following "shim_lock &&" with the exit above?  Can this be squashed into the shim-support-for-shim-lock-measure.patch?

recipes-openxt/xenclient/xenclient-root-ro/init.root-ro
```diff
@@ -314,13 +314,6 @@ unlock_config()
     case "$sig" in
     LUKS)
         if [ -e /boot/system/tpm/enabled ]; then
-            [ "$measured" != 0 ] && {
```
jandryuk: Maybe explicitly state in the commit messages that this is okay?  A DRTM boot doesn't need the TXT check since the unseal will fail?

recipes-openxt/xenclient/xenclient-root-ro/init.root-ro
```diff
@@ -314,13 +314,6 @@ unlock_config()
...
-                maybe_break "txt-fail"
-                txt_failure
```
jandryuk: txt_failure is now unused.  But maybe it would better to move the txt_failure error message after the unseal fails?  Display it conditionally depending on whether or not this is an EFI boot.

conf/machine/openxt-installer.conf
```diff
@@ -0,0 +1,21 @@
...
+
+require xenclient-common.conf
+
+# set empty provider for initscripts as they are different for installer and dom0
+# which share the same machine and require different initscritpts
+# therefore for every machine using dom0 arch initscripts have to
+# be installed separately
+VIRTUAL-RUNTIME_initscripts = ""
+
+PREFERRED_PROVIDER_libgl1 = "nvidia-lib"
```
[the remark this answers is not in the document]
tklengyel: Probably not, we pretty much just copy-pasted the existing dom0 conf.

recipes-support/pesign/files/0001-Disable-warning.patch
```diff
@@ -0,0 +1,25 @@
+From b27e3c385e66594687f58a6f7eedff4374f5929c Mon Sep 17 00:00:00 2001
+From: Tamas K Lengyel <tamas@tklengyel.com>
+Date: Fri, 2 Mar 2018 12:59:39 -0700
+Subject: [PATCH] Disable #warning
```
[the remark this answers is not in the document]
tklengyel: Correct, only because of -Werror

recipes-core/images/xenclient-installer-image.bb
tklengyel: The kernel used for the installer has all kernel modules built in. OTOH if this is an out-of-tree kernel module that is not part of the standard linux distribution, then it is not built because OE can't handle the 32/64-bit discrepancy between the build environment and the 64-bit kernel.

recipes-core/initscripts/initscripts-1.0/openxt-installer/urandom
tklengyel: SGTM

recipes-core/packagegroups/packagegroup-xenclient-installer.bb
tklengyel: Strictly speaking it is not needed but the build was complaining about missing dependencies otherwise. I suspect this was needed beforehand but just happened to work because the installer shared the same image with dom0.

recipes-extended/xen/files/efi-require-shim.patch
tklengyel: It is kept in this form because the patch that errors out when the shim protocol is not found is not something that can be ever upstreamed into Xen, whereas calling Measure could be once my PR is merged into the shim, in which case they will need to be gated on shim_lock being defined.

recipes-openxt/xenclient/xenclient-root-ro/init.root-ro
tklengyel: Correct, I'll update the commit msg.

recipes-openxt/xenclient/xenclient-root-ro/init.root-ro
tklengyel: Not possible unfortunately right now because the 32-bit kernel lacks the capacity to check if it's booted under UEFI. Once dom0 is upgraded to 64-bit that distinction can be made.

recipes-core/images/xenclient-installer-image.bb
jandryuk: Ah, okay.  You did INSANE_SKIP_kernel-module-txt = "arch" elsewhere, and I recognized it for the mixed 32/64 compilation.

recipes-extended/xen/files/efi-require-shim.patch
jandryuk: Okay.

recipes-openxt/xenclient/xenclient-root-ro/init.root-ro
jandryuk: Ah, okay.

recipes-core/images/xenclient-installer-image.bb
tklengyel: Correct, I believe we tried that for the kernel modules too and it would still fail so that's why ended up compiling everything into the kernel.

jandryuk: @tklengyel , I just pushed a branch here: https://github.com/jandryuk/xenclient-oe/commits/stable-6-64bit with changes for building 64bit compat kernels and modules.
'ERROR_QA_remove += "arch" ' lets bitbake build and package 64bit modules from the kernel recipe.  You could do the same if you want to use modules with the installer kernel.

tklengyel: @jandryuk thanks! At the moment I don't believe there is a problem with the selection of kernel modules for the installer, so unless someone can point to a hardware that needs a special out-of-tree kernel module for the installer to work, I think it's safe to keep things as is.

jean-edouard:
Merging soon.

PR #830, opened 2018-03-14
OXT-880: OE upgrade to pyro

eric-ch:
Upgrade xenclient-oe layer to use Bitbake 1.34 and Openembedded 2.3 (Pyro).

# Points of interest:
- `python3` is required on the host;
- `base_contains` is deprecated;
- Images no longer use `xc.` prefix and instead leverage `CONVERSION_CMD` (see classes/openxt_image_types.bbclass);
- layer recipes/append try to use upstream provided ones as much as possible;
- `ipxe`, `vgabios`, `seabios` now append meta-virtualization provided recipes;
- `udev` is now provided by `eudev`;
- unused and deprecated bbclass in the layers are gone;
- read-only images now use `read-only-image` in `IMAGE_FEATURES` to leverage upstream support;
- Use `populate-volatiles.sh` to handle volatile file-systems where possible (almost everywhere) and comply with `read-only-image`;
- `debug-tweaks` in `IMAGE_FEATURES` will now start a tty on `hvc0` for service-vms;
- `debug-tweaks` in `IMAGE_FEATURES` will now make sysvinit and compliant initscripts verbose;
- `debug-tweaks` in `IMAGE_FEATURES` will now configure SELinux to `permissive`;
- read-only images use `mount-early.sh` to mount required volatile file-systems early enough, this should fix previous problems seen when switching to between read-write and read-only (and silence a few warnings), this is configured through `/etc/fstab.early`;
- `/run` size is extended to 5M.
- `openssh` will use volatile keys when deployed in a read-only image, same for the v4v daemon (except dom0 for `sshd` that will use `/config` as usual);
- `sshd-v4v` is now the `sshd` over `v4v` and `sshd` is the regular TCP/IP ssh daemon;
- To comply with usptream, `/config/etc/ssh/enabled` now is controlled through `/config/etc/ssh/sshd_not_to_be_run`, `sshd-v4v` will use the similar `/config/etc/ssh/v4v_not_to_be_run` instead of former `/config/etc/ssh/disable-v4v`;
- Additional steps to recipes no longer change the `CWD` to `${B}`, to restore the behavior,  `do_step[dirs] = "${B}"`;
- Each recipe now has a separate sysroot-native and sysroot, this may expose missing DEPENDS and cross-sysroot contamination in downstream projects.
- `openxt-installer` `MACHINE` configuration is now `core2-64`.

# Work-around:
- GHC 6 does not guaranty the stack to be 16 bytes aligned, so images have to be compiled with `-mstackalign`;
- eudev does not play nice with LVM yet, some warnings can be seen during installation, lvcreate will not Zero or wipe signatures until figured out;
- Some SELinux avcs are left;
- The font of the debug terminal is now misc-fixed isntead of... DejaVu Mono?
- NetworkManager & NetworkManager Applet are version 0.9.2 still
- Surf 0.7 and Webkit 1.8.3 still
- `libvhd` has textrel that should be investigated

# Dependencies:
- https://github.com/OpenXT/openxt/pull/287
- https://github.com/OpenXT/meta-openxt-haskell-platform/pull/9
- https://github.com/OpenXT/icbinn/pull/2
- https://github.com/OpenXT/installer/pull/65
- https://github.com/OpenXT/manager/pull/110
- https://github.com/OpenXT/blktap3/pull/5
- https://github.com/OpenXT/v4v/pull/31


eric-ch:
Add https://github.com/OpenXT/v4v/pull/31 to handle V4V compilation with GCC 6.4.

eric-ch:
Build [$1423](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1423)

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1426
LGTM, merging soon along with dependencies.

PR #832, opened 2018-03-22
OXT-1282: blktap3: add icbinn support

mahantesh-ais:
OXT-1282

### DIFF
--- a/recipes-extended/xen/blktap3.bb
+++ b/recipes-extended/xen/blktap3.bb
@@ -1,13 +1,14 @@
 DESCRIPTION = "blktap3"
 LICENSE = "BSD-3-Clause"
 LIC_FILES_CHKSUM="file://LICENSE;md5=a9e8593dad23434929bc6218a17b5634"
-DEPENDS = "xen" 
+DEPENDS = "xen libicbinn"
 
 PV = "0+git${SRCPV}"
 
 SRCREV = "${AUTOREV}"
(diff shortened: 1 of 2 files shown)


mahantesh-ais:
To be merged along with:
https://github.com/OpenXT/blktap3/pull/4

rneilturner:
This PR may cause an issue with the recently merge UEFI. libicbinn-resolved depends on libicbinn which depends on libv4v. Bringing in libv4v will cause issues with the 64-bit installer image due to the inclusion of v4v-module.

eric-ch:
@rneilturner Yes it will likely require some `INSANE_SKIP = "arch"` statement in v4v-module since the `MACHINE` is `core2-32` and the kernel `x86_64`. This should build in the `core2-64` with Pyro.

A preferable solution imho, would be to split the recipe, have a regular `blktap3.bb` and `blktap3-icbinn.bb` for images where icbinn support is desired. That would keep dependencies in check since libicbinn brings a few (`xc-rpcgen` for one).

mahantesh-ais:
icbinn support is now added as a patch so that it can separated during compile time if needed. As said by @eric-ch this PR now depends on Pyro upgrade. Thanks.

jean-edouard:
LGTM, started a build here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1445
Please PR against stable-8 as well.

eric-ch:
@mahantesh-ais Please PR against stable-8 as well.

eric-ch:
I do not have a way to test `icbinn`.
The patch port LGTM and I cannot justify why `libicbinn-resolved` is used, LGTM too.

PR #834, opened 2018-03-23
remove useless cdrom VBD nodes from xenstore

mahantesh-ais:
blkfront doesn't support CDROMs in HVMs. However, XL creates guest CDROM VBD
nodes for the virtual CD, when it only makes sense for their stubdoms. XL also
creates stubdom and guest nodes for atapi-pt drives, which doesn't make sense.
To summarize, the creation of guest virtual cd nodes and all atapi-pt nodes
should be removed from XL.

OXT-1259

### DIFF
--- a/recipes-extended/xen/files/libxl-iso-hotswap.patch
+++ b/recipes-extended/xen/files/libxl-iso-hotswap.patch
@@ -10,9 +10,8 @@ LONG DESCRIPTION:
 ################################################################################
 CHANGELOG
 ################################################################################
-Author: Chris Rogers <rogersc@ainfosec.com>
-Date:   Tue Apr 18 14:43:21 2017 -0400
-
+Authors: Chris Rogers <rogersc@ainfosec.com>
+         Mahantesh Salimath <salimathm@ainfosec.com>
 ################################################################################
 REMOVAL
 ################################################################################
@@ -125,9 +124,9 @@ Index: xen-4.9.1/tools/libxl/libxl_disk.c
 +    disk_empty.vdev = libxl__strdup(NOGC, olddisk->vdev);
 +    disk_empty.pdev_path = libxl__strdup(NOGC, "");
 +    disk_empty.is_cdrom = 1;
-+    libxl__device_disk_setdefault(gc, &disk_empty, domid);
++    libxl__device_disk_setdefault(gc, &disk_empty, stubdomid);
 +
-+    lock = libxl__lock_domain_userdata(gc, domid);
++    lock = libxl__lock_domain_userdata(gc, stubdomid);
 +    if (!lock) {
 +        rc = ERROR_LOCK_FAIL;
 +        goto out;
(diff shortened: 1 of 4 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1430

jean-edouard:
LGTM, works as expected with virtual CD drives, even fixes OXT-1249 as a side effect! :)
I will grab a laptop with a CDROM drive and make sure atapi-pt vbd nodes are also gone.

recipes-extended/xen/files/libxl-iso-hotswap.patch
```diff
@@ -125,9 +124,9 @@ Index: xen-4.9.1/tools/libxl/libxl_disk.c
 +    disk_empty.vdev = libxl__strdup(NOGC, olddisk->vdev);
 +    disk_empty.pdev_path = libxl__strdup(NOGC, "");
 +    disk_empty.is_cdrom = 1;
-+    libxl__device_disk_setdefault(gc, &disk_empty, domid);
++    libxl__device_disk_setdefault(gc, &disk_empty, stubdomid);
```
jandryuk: Are these stubdomid changes correct?  Is this to force the CDROM VBDs to the stubdom?  The entire commit messages is replicated in the libxl-remove-useless-nodes.patch description, so it's not immediately clear why these are needed.

Will this break the non-stubdom case?

recipes-extended/xen/files/libxl-remove-useless-nodes.patch
```diff
@@ -0,0 +1,86 @@
...
++        }
++        d_config->num_disks = j;
++        d_config->disks = libxl__realloc(NOGC,
++                                         d_config->disks,
++                                         sizeof(*(d_config->disks)) * j);
++    }
++
+     libxl__multidev_begin(ao, &dcs->multidev);
+     dcs->multidev.callback = domcreate_launch_dm;
+     libxl__add_disks(egc, ao, domid, d_config, &dcs->multidev);
```
jandryuk: Would it make sense to just skip creating these entries down in libxl_disk.c:device_disk_add() instead of filtering the disk config entries?  Or is it harder to do at that level?

jean-edouard:
atapi-pt nodes are gone, will merge after review comments are addressed.

jean-edouard:
@jandryuk do you have more questions? Do you approve this PR? Thanks

jandryuk:
I think the chances are okay. I'd improve the commit message to explicitly stated what @mahantesh-ais said about the libxl-iso-hotswap.patch changes. Maybe it should be a separate change?

Libxl-remove-useless-nodes.patch would be more accurately named libxl-dont-create-atapi-cdrom-vdb-xs-nodes.patch since that is the the actual functionality.

How does that sound?

jean-edouard:
More comments and a better filename, sounds good to me! Thanks

mahantesh-ais:
Fixed comments and renamed the patch to something more appropriate. Thanks

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1436
LGTM, merging Monday.

PR #839, opened 2018-03-27
service-vms: Remove getty on consoles.

eric-ch:
There is no need for it to start getty on tty[0-9] in service-vm.


jandryuk:
If you set USE_VT="0" for the machine conf, you can omit the getty lines from inittab in the first place.  They are added by the sysvinit-inittab recipe.

eric-ch:
@jandryuk Indeed, will re-open with the fix. Thanks!

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1430

PR #844, opened 2018-03-29
tpm-functions: use dmesg instead of not-always-there sys node to detect TPM 2.0

jean-edouard:

### DIFF
--- a/recipes-openxt/xenclient/xenclient-tpm-scripts/tpm-functions
+++ b/recipes-openxt/xenclient/xenclient-tpm-scripts/tpm-functions
@@ -30,7 +30,15 @@ clean_old_tpm_files () {
 }
 
 is_tpm_2_0 () {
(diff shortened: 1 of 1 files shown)


crogers1:
Jed we're going to be submitting a PR soon with an alternative approach that still queries the sysfs device, but handles the general case better, I recommend taking a look at that.  

I don't think this PR is a bad way of checking for TPM version, but the dmesg buffer could empty.  Asking the sysfs device, in my mind, will be more reliable.

jean-edouard:
Thanks for the info @crogers1, I guess we should hold off the merge then.
For the record, this PR is built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1436

crogers1:
@jean-edouard no problem.  The engineer working this is out today, I can make sure he pushes it first thing Monday

eric-ch:
@crogers1 Any news?

crogers1:
I believe Rich will be posting a PR soon.

eric-ch:
https://github.com/OpenXT/xenclient-oe/pull/879 got merged. This PR can now be closed.

PR #846, opened 2018-03-30
STABLE-8: Only spawn a getty on tty1 when booting in console mode

jean-edouard:
(cherry picked from commit 1466aa070086577183dc8560fd79be73f0a0a47c)

### DIFF
--- a/conf/machine/xenclient-dom0.conf
+++ b/conf/machine/xenclient-dom0.conf
@@ -18,3 +18,4 @@ MACHINE_FEATURES = "kernel26 screen keyboard ethernet pci usbhost acpi ext2 x86"
 
 SERIAL_CONSOLE = "115200 hvc0"
 
+USE_VT = "0"
--- a/recipes-openxt/xenclient/surfman/surfman.initscript
+++ b/recipes-openxt/xenclient/surfman/surfman.initscript
@@ -27,7 +27,11 @@ SURFMAN_OPTS=""
 
 case "$1" in
   start)
-        grep -q no-graphics /proc/cmdline && SURFMAN_OPTS="${SURFMAN_OPTS} -c"
+        grep -q no-graphics /proc/cmdline && {
+            SURFMAN_OPTS="${SURFMAN_OPTS} -c"
+            # We only want a getty on tty1 when in no-graphics mode (console mode)
+            while true; do /sbin/getty 38400 tty1; sleep 1; done &
+        }
         grep -q safe-graphic /proc/cmdline && SURFMAN_OPTS="${SURFMAN_OPTS} -s"
 
         # Allow 55MB for core dumps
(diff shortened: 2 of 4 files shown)


jandryuk:
So, I'd rather do this via changing the runlevels.  I'm prep-ing a patch to try out that approach.

eric-ch:
@jandryuk I was curious about it too, if there is anything in that old branch that might help: https://github.com/eric-ch/xenclient-oe/tree/debug-runlevel3 (there was not much iirc).

jandryuk:
@eric-ch by the time I saw this, I'd basically made equivalent changes.  The are some differences for the pyro uprev, though.  Here's a branch: https://github.com/jandryuk/xenclient-oe/tree/runlevels , but there are some quirks.  I added commit "surfman: Restart input server when stopping" , so "input-daemon: Only run at runlevel 5" can be dropped - they are two different approaches described below.

Booting to runlevel 3 works for console access with a getty on tty1.
Runlevel 5 boots into surfman without the tty1 getty.

The tricky part is input_server.  We can either run input_server at all runlevels or only runlevel 5 with surfman.  If input_server is disabled at runlevel 3, then booting to console has log spam about from xenmgr about the missing com.cixtrix.xenclient.input DBus endpoint.  If input_server is enabled, booting into each runlevel is fine.  The problem is switching from runlevel 5 to runlevel 3 ```telinit 3```.  That kills surfman and drops to a getty on tty1.  However, if input_server keeps running, then there is no keyboard input.  input_server must not release the input/keyboard when surfman exits, so it don't end up at dom0.  Again, we can stop input_server at runlevel 3, but that produces the log spam.

The right solution would be to somehow signal input server to release the input devices when surfman exits.  The hack would be to restart input_server after exiting surfman - commit "surfman: Restart input server when stopping".  That would put input_server back to its "ungrabbed" state so dom0 could still access the keyboard.

Another issue is transitioning runlevel 3 -> 5.  surfman start and displays the Loading... splash screen.  But surfman does know about UIVM, so it never switches back to displaying the UI.  Looks like xenvm calls set_pv_display to register the VM, but that doesn't happen on the restart.  SELinux prevents the following, but some dbus-send commands as sysadm made surfman switch back to UIVM.  dbus-send in an initscript may be fine...

```
dbus-send --print-reply --system --dest=com.citrix.xenclient.surfman / com.citrix.xenclient.surfman.set_pv_display int32:2 string:""
dbus-send --print-reply --system --dest=com.citrix.xenclient.surfman / com.citrix.xenclient.surfman.set_visible int32:2 int32:5 boolean:false
dbus-send --print-reply --system --dest=com.citrix.xenclient.input / com.citrix.xenclient.input.attach_vkbd int32:2
```

In summary, straight booting works.  Switching runlevels is a little weird.

jandryuk:
Oh, my branch is against master.

jandryuk:
Also, "Looks like xenvm calls set_pv_display to register the VM" - the dbus calls are in xenmgr after the xl conversion.

eric-ch:
[OXT-1300](https://openxt.atlassian.net/browse/OXT-1300).

eric-ch:
https://github.com/OpenXT/xenclient-oe/pull/858 got merged. This PR can now be closed.

PR #853, opened 2018-04-04
Change the tpm2.0 identification method

rneilturner:
Linux Kernel 4.11 changed the location yet again on where to find the TPM description. Make the function for finding the TPM version more robust for any future changes.

### DIFF
--- a/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
+++ b/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
@@ -28,7 +28,14 @@ DEFINIT=/sbin/init
 FIRSTBOOT_FLAG=/boot/system/firstboot
 
 is_tpm_2_0 () {
-    [ -e /sys/class/tpm/tpm0/device/description ] && cat /sys/class/tpm/tpm0/device/description | grep "2.0" &>/dev/null
+   files=$(find /sys/ | grep "description$")
+   for file in $files; do
+      tpm=$(cat ${file})
+      if [ "${tpm}" == "TPM 2.0 Device" ]; then
+         return 0
+      fi
+   done
+   return 1
 }
 
 #listpcrs sample output:
--- a/recipes-openxt/xenclient/xenclient-tpm-scripts/tpm-functions
+++ b/recipes-openxt/xenclient/xenclient-tpm-scripts/tpm-functions
@@ -30,7 +30,14 @@ clean_old_tpm_files () {
 }
 
 is_tpm_2_0 () {
-   [ -e /sys/class/tpm/tpm0/device/description ] && cat /sys/class/tpm/tpm0/device/description | grep "2.0" &>/dev/null
+   files=$(find /sys/ | grep "description$")
+   for file in $files; do
+      tpm=$(cat ${file})
+      if [ "${tpm}" == "TPM 2.0 Device" ]; then
+         return 0
+      fi
+   done
+   return 1
 }
 
 pcr_bank_exists () {


recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
```diff
@@ -28,7 +28,14 @@ DEFINIT=/sbin/init
 FIRSTBOOT_FLAG=/boot/system/firstboot
 
 is_tpm_2_0 () {
-    [ -e /sys/class/tpm/tpm0/device/description ] && cat /sys/class/tpm/tpm0/device/description | grep "2.0" &>/dev/null
+   files=$(find /sys/ | grep "description$")
```
jean-edouard: `find` can look for filenames using -name, no need for a pipe and a grep

recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
```diff
@@ -28,7 +28,14 @@ DEFINIT=/sbin/init
...
+   for file in $files; do
+      tpm=$(cat ${file})
+      if [ "${tpm}" == "TPM 2.0 Device" ]; then
```
jean-edouard: The proper syntax is a single = sign

recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
dpsmith: Jason and I were discussing this, instead of search for the generic name "description", which results in multiple hits(and time to do the I/O and comparison), a variation to Jed's approach to grepping dmesg for the ACPI discovery message. Unlike TPM1.2, it appears all TPM 2.0 devices use the ACPI id of MSFT0101 [[1]](https://elixir.bootlin.com/linux/v4.14.4/source/drivers/char/tpm/tpm_tis.c#L99). Therefore you can drop the switch to `find` and can instead check for the existence of `/sys/bus/acpi/devices/MSFT0101:00` (only confirms the existence of the first tpm 2 device, should a system ever exist that had more than one). For redundancy, you can compare the contents of `/sys/bus/acpi/devices/MSFT0101:00/description`, which is where `tpm0/device/description` looks to be getting the value it provides. The one down side to checking ACPI is that it tells you that there is a TPM 2.0 device in your system, it does not confirm that the tpm driver loaded and the device node is tpm0.

crogers1: @dpsmith 
I couldn't find MSFT0101 in any of the TCG specs, it seems to be a Linux thing, which suggests to me that it could be changed in a future release.  When asked about it, I recommended against hardcoding the MSFT path for the above reason.   You've raised the issue of "what happens if there are multiple TPM devices?", which is another valid reason.  Third, handling this lookup more generically increases the fault tolerance of this code overall, hopefully requiring less maintenance.

Finally, I think the time to read and compare the output of multiple 'description' files is trivial compared to the whole workstation install and sealing/unsealing process, which takes a couple of minutes.

@jean-edouard 
```find /sys -type f -name description``` works for me.

File: recipes-openxt/initrdscripts/initramfs-xenclient/init.sh (hunk 2 above)
Comment (crogers1) [reply recipes-openxt/initrdscripts/initramfs-xenclient/init.sh]: ```==``` is a bash alias for ```=```, and it's behaviour only changes between double-brackets.

File: recipes-openxt/initrdscripts/initramfs-xenclient/init.sh (hunk 1 above)
Comment (dpsmith) [reply recipes-openxt/initrdscripts/initramfs-xenclient/init.sh]: It would not be in the TCG specs and it is not a Linux thing. Its an ACPI thing that requires registration, in this case MSFT is one of Microsoft's ACPI registrations[[1]](http://www.uefi.org/acpi_id_list). As you can see from this Dell help[[2]](http://www.dell.com/support/article/aw/en/awbsdt1/sln302979/unknown-device-with-windows-7-on-newer-systems-with-tpm-20?lang=en), it is even how MS identifies TPM2 devices on Windows[[3]](https://support.microsoft.com/en-us/help/2920188/update-to-add-support-for-tpm-2-0-in-windows-7-and-windows-server-2008)

Comment (jandryuk) [reply recipes-openxt/initrdscripts/initramfs-xenclient/init.sh]: If you look at https://elixir.bootlin.com/linux/v4.14.4/source/drivers/char/tpm/tpm_tis.c#L99 you'll see that Linux keys off MSFT0101 to decide if it's a TPM2.

But backing up, what are you keying off of?  An ACPI sysfs entry?  If you are going to use find, I think it should at be focused on /sys/device/LNXSYSTEM* or at least /sys/device/.  Why search through totally unrelated parts of /sys?  I agree with writing flexible code, but it should be balanced with avoiding unnecessary work and false positives.

$ find /sys/ -name description
find: ‘/sys/kernel/debug’: Permission denied
/sys/devices/platform/i8042/serio1/description
/sys/devices/platform/i8042/serio0/description
/sys/devices/LNXSYSTM:00/LNXSYBUS:00/PNP0A08:00/device:08/PNP0C31:00/description
find: ‘/sys/fs/pstore’: Permission denied
/sys/bus/serio/drivers/serio_raw/description
/sys/bus/serio/drivers/atkbd/description
/sys/bus/serio/drivers/psmouse/description

File: recipes-openxt/initrdscripts/initramfs-xenclient/init.sh (hunk 2 above)
Comment (jean-edouard) [reply recipes-openxt/initrdscripts/initramfs-xenclient/init.sh]: Actually `[` is a binary, not a built-in, and `man [` doesn't mention `==`, most `[` implementations just happen to alias `==` to `=` because it's a common mistake.

File: recipes-openxt/initrdscripts/initramfs-xenclient/init.sh (hunk 1 above)
Comment (jandryuk) [reply recipes-openxt/initrdscripts/initramfs-xenclient/init.sh]: Interesting.  So for right now, the lack of the the caps file indicates TPM2.  But that could change in the future, in which case, the TCG version should be checked.  So something like this untested:
```
is_tpm2() {
    [ ! -e /sys/class/tpm/tpm0/device/caps ] && return 1
    [ $( awk '/TCG version:/ { print $3 }' ) = "2.0" ] && return 1
    return 0
}
```

eric-ch: > lack of the the caps file indicates TPM2

@jandryuk Yes, or lack of TPM altogether I guess, which is not checked in the current implementation either it seems.

eric-ch:
https://github.com/OpenXT/xenclient-oe/pull/879 got merged. This PR can now be closed.

PR #854, opened 2018-04-10
OXT-1297: linux-4.14: objtool for out-of-tree modules.

eric-ch:
CONFIG_STACK_VALIDATION=y will require objtool available for out-of-tree
modules to generate object files.

Taken from commit:
https://patchwork.openembedded.org/patch/148047/

### DIFF
--- a/recipes-kernel/linux/4.14/linux-openxt_4.14.22.bb
+++ b/recipes-kernel/linux/4.14/linux-openxt_4.14.22.bb
@@ -5,8 +5,6 @@ require recipes-kernel/linux/linux.inc
 require recipes-kernel/linux/linux-openxt.inc
 
 PV_MAJOR = "${@"${PV}".split('.', 3)[0]}"
-INSANE_SKIP_kernel-module-txt = "arch"
-INSANE_SKIP_kernel-vmlinux = "arch"
 
 FILESEXTRAPATHS_prepend := "${THISDIR}/patches:${THISDIR}/defconfigs:"
 SRC_URI += "${KERNELORG_MIRROR}/linux/kernel/v${PV_MAJOR}.x/linux-${PV}.tar.xz;name=kernel \
--- a/recipes-kernel/linux/linux-openxt-copy-objtool.inc
+++ b/recipes-kernel/linux/linux-openxt-copy-objtool.inc
@@ -1,0 +1,11 @@
+# Backport from upstream commit to attend 4.14 and later kernels and
+# out-of-tree modules:
+# https://patchwork.openembedded.org/patch/148047/
+do_shared_workdir_append () {
+    # With CONFIG_UNWINDER_ORC (the default in 4.14), objtool is required for
+    # out-of-tree modules to be able to generate object files.
+    if [ -x tools/objtool/objtool ]; then
+        mkdir -p ${kerneldir}/tools/objtool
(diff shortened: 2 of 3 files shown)


recipes-kernel/linux/4.14/linux-openxt_4.14.22.bb
```diff
@@ -3,6 +3,7 @@ DESCRIPTION = "Linux kernel for OpenXT service VMs."
 # Use the one from meta-openembedded/meta-oe
 require recipes-kernel/linux/linux.inc
 require recipes-kernel/linux/linux-openxt.inc
+require recipes-kernel/linux/linux-openxt-4.14.inc
```
jandryuk: Do you want to just put add linux-openxt.inc?  It'll be required for all version going forward, and the code is conditioned on the existence of the objtool in the first place.

Also, objtool depends on the elfutils to build objtool.
"Cannot use CONFIG_STACK_VALIDATION, please install libelf-dev, libelf-devel or elfutils-libelf-devel"
DEPENDS += "elfutils-native" didn't work, and I ended up installing libelf-dev on the build host.

eric-ch: > Do you want to just put add linux-openxt.inc

As I took the patch from patchwork directly, I expect this will make its way upstream sooner rather than later. Splitting it in `linux-openxt-4.14.inc` was to make the revert easier.

`libelf-dev` is already in the [dependency requirement list](https://github.com/OpenXT/openxt/blob/master/build-scripts/oe/setup.sh#L35) for the environment to build OpenXT. I believe it was added for that reason when the changes for UEFI support introduced the 64bit kernel. I do not remember why the native recipe does not cut it either.

jandryuk: I was surprised pyro didn't have the patch already, but pyro doesn't support Linux 4.14.

I think I searched for the wrong thing, elfutils, in the OpenXT PKGS var, which naturally failed. Good to see it already listed.

eric-ch: > I was surprised pyro didn't have the patch already

Me too! Even `sumo` does not have it yet :).

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1452

jean-edouard:
@jandryuk good to go?

jandryuk:
The new file isn't version-specific, so linux-openxt-copy-objtool.inc would be a more descriptive name for the functionality. It's otherwise good, so I'm not going to block if @eric-ch wants to leave it as is.

eric-ch:
@jandryuk `linux-openxt-copy-objtool.inc` that makes more sense, I misunderstood the initial comment.

jandryuk:
I originally said put it in linux-openxt.inc, but I saw your point about keeping it separate, too. The name change only came up now when I looked again.

eric-ch:
Name change, and include in `linux-openxt.inc`.
Also remove the `INSANE_SKIP` since the installer is built for a 64bit MACHINE.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1455

jean-edouard:
Merging soon.

PR #858, opened 2018-04-12
OXT-1300: Use runlevels to control console login

jandryuk:
We only want to run a getty on tty1 in console mode, and suppress the getty for graphical boots.  Use init runlevels to differentiate between text console (runlevel 3) and graphical (5) boots.

This is an alternate approach to https://github.com/OpenXT/xenclient-oe/pull/809
Discussion also continued here - https://github.com/OpenXT/xenclient-oe/pull/846 - and has been updated and repeated below.

Straight booting works fine, but transitioning between runlevels (e.g. ```telinit 3```) is broken.

One issue is input_server. We can either run input_server at all runlevels (default) or only runlevel 5 (https://github.com/jandryuk/xenclient-oe/commit/e38148221cfd61136e12a18302652697a2dc11d0  not included in PR) with surfman. If input_server is disabled at runlevel 3, then booting to console has log spam from xenmgr about the missing com.cixtrix.xenclient.input DBus endpoint. If input_server is enabled, booting into each runlevel is fine. The problem is switching from runlevel 5 to runlevel 3 ```telinit 3```. That kills surfman and drops to a getty on tty1. However, input_server keeps running, so there is no keyboard input. input_server must not release the input/keyboard when surfman exits, so it don't end up at dom0. Again, we can stop input_server at runlevel 3, but that produces the log spam.

The right solution would be to somehow signal input server to release the input devices when surfman exits. The hack would be to restart input_server after exiting surfman - commit "surfman: Restart input server when stopping" https://github.com/jandryuk/xenclient-oe/commit/1a3a9a7ae2d4d76bd98495ef334820c8a895c90c. That puts input_server back to its "ungrabbed" state so dom0 could still access the keyboard.

Another issue is transitioning runlevel 3 -> 5. surfman start and displays the Loading... splash screen. But surfman does not know about UIVM, so it never switches back to displaying the UI. Normally, xenmgr calls set_pv_display to register the VM, but that doesn't happen on surfman restart. SELinux prevents the following, but manually running dbus-send commands as sysadm makes surfman switch back to UIVM.  dbus-send in an initscript would be another hack, but a preliminary attempt at such a hack did not work.
```
dbus-send --print-reply --system --dest=com.citrix.xenclient.surfman / com.citrix.xenclient.surfman.set_pv_display int32:2 string:""
dbus-send --print-reply --system --dest=com.citrix.xenclient.surfman / com.citrix.xenclient.surfman.set_visible int32:2 int32:5 boolean:false
dbus-send --print-reply --system --dest=com.citrix.xenclient.input / com.citrix.xenclient.input.attach_vkbd int32:2
```
Note: it appears that when the keyboard is detached, a terminal in UIVM receives a continuous stream of "Enter" keystrokes.

The correct fix would be for xenmgr to listen for surfman's DBus start_service signal and re-issue the surfman and input DBus calls.


eric-ch:
Building: [$1466](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1466)

eric-ch:
So this will work only with measured launch (when using `init.root-ro`). Otherwise, the initramfs will set `DEFINIT=/sbin/init` which will not receive the runlevel (see [init.sh](https://github.com/OpenXT/xenclient-oe/blob/master/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh#L195)).

The runlevel should probably be parsed in the `init.sh` script and passed to `switch_root` as argument for `init`.

jandryuk:
You got me - I didn't try a non-ML install.  Using init.sh looks good, and I can just extend the existing read_args parsing.

I also missed openxt.cfg for UEFI and removing no-graphics from surfman's initscript.  I'll add those.

jandryuk:
The UEFI change is untested.

eric-ch:
Build [$1476](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1476)

eric-ch:
Will merge soon.

Can you submit a PR against stable-8 as well please? As described in OXT-1300, this is required in 8.0.

PR #862, opened 2018-04-17
OXT-1304: v4v: Fix SMAP & HVM guest use

jandryuk:
When SMAP is enabled, v4v_send fails from an HVM domain.  The guest page table walk fails with an emulated SMAP violation.  SMAP emulation needs to be disabled to allow the send.

This is a forward port of a stable-6 patch implementing the same functionality.  It required changing the Xen function used to disable SMAP.  Integrating this into a normal build tests fine with the PV domains.  I haven't actually tested with HVM on a master build at this time.  This is needed for the HVM NDVM, which is still on my to-do list.


eric-ch:
Build [$1465](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1465), the patch does not seem right.

recipes-extended/xen/files/v4v.patch
```diff
@@ -130,7 +130,7 @@ PATCHES
      int port;
 --- /dev/null
 +++ b/xen/common/v4v.c
-@@ -0,0 +1,1964 @@
+@@ -0,0 +1,1985 @@
```
eric-ch: Should be `@@ -0,0 +1,1979 @@`

jandryuk:
Not only was the patch mangled, it wasn't even the right one!  The tested commit was still on my build-server, but I pushed the work in progress, broken one to github from my workstation.   I've force pushed the branch with the Xen 4.9 changes mentioned in the original PR message.  Sorry about that.

eric-ch:
No worries!
Building [$1467](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1467)

eric-ch:
Could you PR against stable-8 as well?

PR #863, opened 2018-04-19
OXT-1302, OXT-1303: linux: upgrade to 4.14.34.

eric-ch:
1. OXT-1302: Upgrade to Linux 4.14.34 (refresh patches and defconfig).
2. OXT-1303: Apply work-around patch until a proper fix is implemented. This is to avoid blocking test efforts of other components. The ticket will be left open.
3. Backport @jandryuk patches to fix 32bit PV kernel boot sequence and xen-netfront race with the toolstack.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1474

jean-edouard:
LGTM, merging tomorrow.

PR #868, opened 2018-04-20
OXT-1306: dbus: Fix recipe DEPENDS statement.

eric-ch:
DEPENDS applies for the whole recipe, giving it an `"_<PACKAGE>"` will
have Bitbake ignore it.

DBus does not require xen as depend (libv4v does).
Since libv4v is optional for that recipe, depend on it only for the
target build. DBus recipe provides native tools, and libv4v is not
needed for that.

### DIFF
--- a/recipes-core/dbus/dbus_1.%.bbappend
+++ b/recipes-core/dbus/dbus_1.%.bbappend
@@ -1,7 +1,14 @@
-DEPENDS_${PN} += " \
+# LibSELinux does not extend nativesdk.
+DEPENDS_append_class-native += " \
+    libselinux \
+"
+# DBus will not link with libv4v in the native case.
+# The v4v kernel headers cannot be expected in the native environment since
+# libv4v depends on Xen, and the hypervisor headers are not separated from the
+# main recipe.
+DEPENDS_append_class-target += " \
     libselinux \
     libv4v \
-    xen \
 "
 FILESEXTRAPATHS_prepend := "${THISDIR}/patches:"
 


eric-ch:
Broken with `nativesdk`. Will push a fix soon.

jandryuk:
The first patch looked good if we want to keep the patch.  "class-target" is interesting - I did not know about it.

I was figuring on just dropping v4v since we are often missing it.  My approach is here: https://github.com/jandryuk/xenclient-oe/tree/dbus-drop-v4v

I guess there could be out-of-tree code that uses family=v4v...  I see one location I missed in bvt.git:
```
src/testcases/syncxt_test.py:    dbus_var = 'DBUS_SYSTEM_BUS_ADDRESS=tcp:family=v4v,host=1.0.0.0,port=5556'
```
That could be changed with a similar fix as the other DBUS_SYSTEM_BUS_ADDRESS changes, I think.

eric-ch:
`class-target` will work with `libselinux` being only added to it and `class-native` (since `libselinux` does not extend `nativesdk`, which is pulled as a dependency of `dnf` with a complete build).

Removing the dependency should be fine unless someone downstream has an issue with it. I will still amend the PR to provide an alternative, but removing the patch (hence the dep) is fine by me.

jandryuk:
I emailed the mailing list to see if anyone would break if we drop the patch.  We can merge this now and always drop later.

LGTM

jean-edouard:
Thanks, merging soon.

PR #870, opened 2018-04-20
STABLE-8: OXT-1306: dbus: Fix recipe DEPENDS statement.

eric-ch:
DEPENDS applies for the whole recipe, giving it an "_<PACKAGE>" will
have Bitbake ignore it.

DBus does not require xen as depend (libv4v does).
Since libv4v is optional for that recipe, depend on it only for the
target build. DBus recipe provides native tools, and libv4v is not
needed for that.


eric-ch:
Build [$1479](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1479)

jean-edouard:
Merging soon.

PR #875, opened 2018-04-25
installer-image: Add timeout to UEFI grub.cfg

jandryuk:
Right now, there is no timeout, so grub waits forever.  Add a default
timeout so we boot into the installer and wait there instead.  This
matches the legacy isolinux boot behaviour.


eric-ch:
Will merge soon. Can you submit a PR against stable-8 as well?

eric-ch:
Build [$1498](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1498)

PR #878, opened 2018-04-26
STABLE-8: OXT-1300: Use runlevels to control console login

jandryuk:
This is the stable-8 version of https://github.com/OpenXT/xenclient-oe/pull/858

We only want to run a getty on tty1 in console mode, and suppress the getty for graphical boots. Use init runlevels to differentiate between text console (runlevel 3) and graphical (5) boots.

More discussion is here https://github.com/OpenXT/xenclient-oe/pull/858


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1490
LGTM, merging soon.

PR #879, opened 2018-04-26
OXT-1311: tpm: Detect TPM version from sysfs caps entry.

eric-ch:
Using /sys/class/tpm/tpm0/device/description is unreliable as the entry
might not be populated by the driver.

From:
https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-class-tpm
It is recommended to use /sys/class/tpm/tpmX/device/caps instead. One
caveat though is documented in the TPM module sources:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/drivers/char/tpm/tpm-sysfs.c?h=v4.14.34#n296

TPM 2.0 devices will not expose /sys/class/tpm/tpmX/device/caps (this is
still the case to this day in Linux 4.16 line). So assuming a
TPM is detected, the absence of this file reveals a TPM 2.0 version.

This is an alternative to other propositions:
- https://github.com/OpenXT/xenclient-oe/pull/844 using dmesg
- https://github.com/OpenXT/xenclient-oe/pull/853 finding the "description" sysfs node.

### DIFF
--- a/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
+++ b/recipes-openxt/initrdscripts/initramfs-xenclient/init.sh
@@ -28,7 +28,20 @@ DEFINIT=/sbin/init
 FIRSTBOOT_FLAG=/boot/system/firstboot
 
 is_tpm_2_0 () {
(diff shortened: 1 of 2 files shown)


jean-edouard:
Tested in stable-8. LGTM, merging soon.

PR #880, opened 2018-04-26
OXT-1310: 8th gen SINIT.

eric-ch:
Add SINIT file for 8th generation Intel Coffee Lake machines and amend the relevant configuration files.


jean-edouard:
Tested in stable-8. LGTM, merging soon.

PR #885, opened 2018-04-26
OXT-1314: Update Xen patches: support for SecureBoot and minor fixes

tklengyel:
Make Xen query the firmware for the status of SecureBoot to determine whether to use `shim_lock->Measure` or `shim_lock->Verify` on the dom0 kernel.

Move definition of `elo_active` to the efi-load-option-support.patch where it belongs.

Signed-off-by: Tamas K Lengyel <lengyelt@ainfosec.com>

### DIFF
--- a/recipes-extended/xen/files/efi-load-option-support.patch
+++ b/recipes-extended/xen/files/efi-load-option-support.patch
@@ -1,7 +1,7 @@
-From d51e22b3876c0a392aad2ef58bba62a334e61b28 Mon Sep 17 00:00:00 2001
+From 3def7c280f64dbb6a9646d94fbbed6f97fa58293 Mon Sep 17 00:00:00 2001
 From: Tamas K Lengyel <tamas@tklengyel.com>
 Date: Thu, 28 Dec 2017 18:50:29 +0100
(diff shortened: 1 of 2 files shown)


richpersaud:
Please include JIRA ticket number.

jandryuk:
Generally looks good.

Do we want to always measure Dom0?  Something like:
```
if (secureboot()) {
  verify()
}
measure()
```
Otherwise we'd have measurements of all the other components but now Dom0, right?  Or am I missing something?

The patch description is now a little stale since it does Verify in addition to Measure.

tklengyel:
@jandryuk dom0 kernel always gets measured, `shim_lock->Verify` incorporates measurements since https://github.com/rhboot/shim/commit/555ef92650944e5ec8ccc23d1d892e984b04afd6

jandryuk:
Okay, great.  I wondered if that might be the case.  Thanks for responding.

eric-ch:
Build [$1498](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1498)

Will merge soon.

Can you submit a PR against stable-8 as well please?

PR #886, opened 2018-04-27
OXT-1315: populate-volatile.sh: depend on cryptdisks

jean-edouard:
OXT-1315

### DIFF
--- a/recipes-core/initscripts/initscripts-1.0/populate-volatile.sh
+++ b/recipes-core/initscripts/initscripts-1.0/populate-volatile.sh
@@ -1,8 +1,8 @@
 #!/bin/sh
 ### BEGIN INIT INFO
 # Provides:             volatile
-# Required-Start:       $local_fs
-# Required-Stop:      $local_fs
+# Required-Start:       $local_fs cryptdisks
+# Required-Stop:        cryptdisks $local_fs
 # Default-Start:        S
 # Default-Stop:
 # Short-Description:  Populate the volatile filesystem


recipes-core/initscripts/initscripts-1.0/populate-volatile.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
 ### BEGIN INIT INFO
 # Provides:             volatile
-# Required-Start:       $local_fs
-# Required-Stop:      $local_fs
+# Required-Start:       $local_fs cryptdisks
+# Required-Stop:        cryptdisks $local_fs
```
eric-ch: This is not parsed AFAIK.

Regardless, even for documentation purpose: [Debian LSBInitScripts documentation](https://wiki.debian.org/LSBInitScripts) defines `$local_fs` as "all local filesystems are mounted", which sounds like `mountall.sh` completely succeeded in an OpenXT setup.

Looking at the runlevel priorities for `S`:
```
lrwxrwxrwx. 1 root root 20 May  2 15:43 /etc/rcS.d/S34cryptdisks -> ../init.d/cryptdisks
lrwxrwxrwx. 1 root root 21 May  2 14:44 /etc/rcS.d/S35mountall.sh -> ../init.d/mountall.sh
lrwxrwxrwx. 1 root root 20 May  2 15:43 /etc/rcS.d/S36udev-cache -> ../init.d/udev-cache
lrwxrwxrwx. 1 root root 30 May  2 14:44 /etc/rcS.d/S37populate-volatile.sh -> ../init.d/populate-volatile.sh
```

Looking at sysvinit_2.88dsf `rc`, `mountall.sh` should be run through:
```
. $scriptname
```
So I am not quite sure...

Maybe `device-mapper` through `cryptsetup` do not expose the `xenclient-log` devnode in time, which leads to `mountall.sh` failing to create expected files?

PR #896, opened 2018-05-09
STABLE-8: OXT-1258:  xen: Ammend PGC_pinned_by_tools declaration

eric-ch:
PGC_pinned_by_tools is a flag in page->count_info field, hence
PGC_count_width should be incremented to accomodate for it. Ref-count
has to be quite high to make this fail, which is why no misbehavior was
seen until now.

XenServer maintains an equivalent hunk in one patch:
https://github.com/xenserver/xen-4.7.pg/blob/1b778262c1b4a3c7c8e504d50e0149fe5909d670/master/nvidia-hypercalls.patch#L487

OXT-1258

### DIFF
--- a/recipes-extended/xen/files/xen-translate.patch
+++ b/recipes-extended/xen/files/xen-translate.patch
@@ -488,21 +488,20 @@ PATCHES
          break;
 --- a/xen/include/asm-x86/mm.h
 +++ b/xen/include/asm-x86/mm.h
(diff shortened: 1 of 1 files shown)


eric-ch:
Building [$1508](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1508)

jean-edouard:
LGTM, merging soon.

PR #899, opened 2018-05-10
STABLE-8: Revert "v4v: Fix SMAP & HVM guest use"

jandryuk:
The original change didn't fix v4v for HVMs with SMAP on Xen 4.9.  It
relied on smap_check_policy in guest_walk_tables to skip the SMAP check,
but that was removed between Xen 4.6 and 4.9.  Upstream NACK-ed
re-introducing support in guest_walk_tables and suggested wrapping the
v4v hypercall in the guest kernel with stac/clac.  We'll do that in the
v4v module: https://github.com/OpenXT/v4v/pull/36

This reverts commit 816157744a05a32e206ddf2eddd504bc0cb09587.

OXT-1304

(cherry picked from commit bfd1cf097cc55504cc2c68513ad940e46cae8bbd)

### DIFF
--- a/recipes-extended/xen/files/v4v.patch
+++ b/recipes-extended/xen/files/v4v.patch
@@ -130,7 +130,7 @@ PATCHES
      int port;
 --- /dev/null
 +++ b/xen/common/v4v.c
-@@ -0,0 +1,1980 @@
+@@ -0,0 +1,1964 @@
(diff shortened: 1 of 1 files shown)


eric-ch:
Building [1513](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1513)

PR #910, opened 2018-05-29
customtask: python reduce was moved in functools.

eric-ch:
xenclient-customtask does not work with Python 3 recent versions:
https://docs.python.org/3/library/functools.html

### DIFF
--- a/classes/xenclient-customtask.bbclass
+++ b/classes/xenclient-customtask.bbclass
@@ -6,9 +6,11 @@ __makeclean() {
     oe_runmake clean
 }
 python do_makeclean() {
+    import functools
+
     workdir = d.getVar('B', True)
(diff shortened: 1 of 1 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1539

jean-edouard:
I tried makeclean on various recipes:
- It "succeeded" on rpc-proxy but didn't clean anything
- It failed on libpng and surfman (No rule to make target 'clean')

Note: rpc-proxy does not use the "build" directory

jean-edouard:
Even if this PR doesn't completely fix makeclean, at least now it doesn't crash anymore.
I will merge this PR this afternoon if there's no objection.

PR #912, opened 2018-06-04
OXT-1327: HVM NDVM Support

jandryuk:
This adds support to running NDVMs as HVMs.  A single disk image is usable as either PV or HVM.  The  mode is selected with /xenmgr/ndvm-mode as "pv" or "hvm".  A reboot is required for the change to take affect.  The default is still pv.

The start of the series adds support for a new .disk image conversion to create a partitioned disk .ext3.disk.vhd.   With that NDVM needs a few small changes.  The NDVM linux config is updated to support running as an HVM.  Some uneeded features are also removed along the way.

There are some changes to allow disabling unneeded/unused devices from HVMs.  In particular, we want to disable graphics and input devices.  While in there, I removed passing qemu options to stubdom through the kernel command line and instead fetch them out of xenstore like mini-os stubdom.  Qubes uses the xenstore mechanism.

HVM NDVM must use a stubdom.  Is this okay?  Otherwise, I think it's just a xsm-policy change to get it working.

Depends on:
https://github.com/OpenXT/xsm-policy/pull/20
https://github.com/OpenXT/manager/pull/116


pearsonk:
Why have a db-key as the decision factor as opposed to breaking the NDVM template into 2 templates, a PV variant and an HVM variant?

pearsonk:
Actually, looking at https://github.com/OpenXT/manager/pull/116 shows the 2 templates. At this point, I would wonder what the purpose of the db-key /xenmgr/ndvm-mode would serve as opposed to simply citing which template gets used during NDVM creation.

Probably worth discussing if we want to keep PV as the default, or change the default to HVM and retain PV as a legacy support option.

jandryuk:
Someone (@rneilturner ?) asked for PV/HVM to be configurable.  When I figured out a single disk image works in both modes, making it runtime configurable is nicer than build time.  Especially if you need to fallback on a deployed system.

The two template solution came about from the xenmgr implementation.  In enumServiceVmTags https://github.com/OpenXT/manager/blob/master/xenmgr/Vm/Templates.hs#L124 you can see xenmgr just grabbing everything named "service-*".  My changed version here: https://github.com/jandryuk/manager/blob/hvm-ndvm/xenmgr/Vm/Templates.hs#L134 filter the ndvm templates so only one is run.  So the ndvm-mode is choosing the template.

Just changing the 6-7 db keys is also an option.  But given my limited Haskell experience, that is more daunting.

While I plan to use HVM, I didn't want to get ahead of myself changing the default, especially when I was asked to make it configurable.

jean-edouard:
+1 @jandryuk.
As a user, I'm not sure I see the point of replacing [a pv guest] with a [a pv guest + an HVM guest], but since there is a need for it, I like the idea of introducing the option while keeping pv as the default.

jandryuk:
Oh, I forgot to wire up the build system.  openxt.git will require a PR.

pearsonk:
I'm not arguing against the run-time configurability and think it is great. It provides a fallback in the case that for some reason a particular NIC doesn't work with an HVM NDVM.

I think the only thing I am not understanding is why we are using both a template approach and a xenmgr DB key. I see that the VM type defined in both templates is 'ndvm', which could point to why the additional logic would be necessary. 

Would it make sense to put the new 'type' as 'ndvm-hvm' and 'ndvm-pv' with legacy support logic (or upgrademgr transition) to change 'ndvm' to 'ndvm-pv'?

The problem that could potentially arise, particularly for multiple-NDVM systems, is that this forces all NDVMs to be the same type. If device support requires running one NDVM as PV where it would be preferable and functional to run all the other NDVMs as HVM, then all the NDVMs would have to be run as PV instead of just the single instance. Unless I am misinterpreting this, which I admittedly haven't had time to dredge deeply into the code and was mostly reacting to the commit message documentation.

cjp256:
With the above changes, will it be easy enough to switch to PVH(v2) once PCI passthrough is supported? PVH seems to be the best fit, but not quite yet ready until PCI passthrough is supported.

jandryuk:
The template approach is how xenmgr works, but I needed a way to control it.  By default, xenmgr grabs everything matching `/usr/share/xenmgr-1.0/templates/default/service-*` to run as service vms.  For that file list, it populates dbd by overwriting each `/vm/${UUID}` dbd entry with the file contents according to the UUID in the file.  Two ndvm entries will conflict over the PCI NICs, so we can only choose one.  Even with different types, we'd need to pick between the two somehow.

To prevent clobbering a VM config, you have to `db-write /xenmgr/overwrite-ndvm-settings false` .  The name (ndvm) is taken from the filename service-ndvm.  So it already is a combination of files and dbd entries.  Maybe it would be better to have a single on-disk template.  Then at runtime tweak the VM configuration as appropriate.

With the submitted PR, you can run both PV and HVM side by side.  I think you must always write `/xenmgr/overwrite-ndvm-settings false` to run multiple NDVMs so you can control the PCI device assignment.  You can create new ndvms, either pv or hvm, with `create_ndvm ${NAME} <pv|hvm>` (defaulting to pv).

@cjp256 Yes, PVH would be great if it were available.  I don't think anything I did precludes it.  The kernel config might need updating, and the vm config would need to be change to PVH.  Some other tweaks might be needed, but without trying out PVH w/ PCI passthrough, I don't know.

cjp256:
@jandryuk thanks

classes/openxt-image-disk.bbclass
```diff
@@ -0,0 +1,110 @@
...
+# in functionality for the limited use case.
+
+BOOTDD_EXTRA_SPACE ?= "2048"
+KERNEL_CMDLINE     ?= "root=/dev/xvda2 ro console=hvc0"
+
+CONVERSIONTYPES_append = " disk"
+
+CONVERSION_DEPENDS_disk = "syslinux \
+                           syslinux-native \
+			   dosfstools-native \
```
jean-edouard: Indent

classes/openxt-image-disk.bbclass
```diff
@@ -0,0 +1,110 @@
...
+			   mtools-native \
+			   "
+
+CONVERSION_CMD_disk() {
+	BOOTDIR="${S}/bootp/boot"
+	BOOTIMG="${S}/bootp.image"
+	ROOTIMG=${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}
+	IMAGE=$ROOTIMG.disk
+
+	#populate_kernel $BOOTDIR
```
jean-edouard: Is this a comment or code that's commented out? This one and the syslinux one below seem unclear to me...

classes/openxt-image-disk.bbclass
```diff
@@ -0,0 +1,110 @@
...
+	install -d $BOOTDIR
+	bbnote "Trying to install ${DEPLOY_DIR_IMAGE}/${KERNEL_IMAGETYPE} as $BOOTDIR/vmlinuz"
+	if [ -e ${DEPLOY_DIR_IMAGE}/${KERNEL_IMAGETYPE} ]; then
+		install -m 0644 ${DEPLOY_DIR_IMAGE}/${KERNEL_IMAGETYPE} $BOOTDIR/
+	else
+		bbwarn "${DEPLOY_DIR_IMAGE}/${KERNEL_IMAGETYPE} doesn't exist"
+	fi
+
+	# initrd is made of concatenation of multiple filesystem images
```
jean-edouard: Do we need this? The new NDVM doesn't seem to have an initrd (nor should it need one).

recipes-extended/xen/files/libxl-display-manager-support.patch
```diff
@@ -46,26 +46,40 @@ PATCHES
...
-+    if (display && display->kind)
-+        flexarray_append_pair(dm_args, "-display", display->kind);
-+    else
-+        flexarray_append_pair(dm_args, "-display", "surfman");
+@@ -1048,11 +1057,6 @@ static int libxl__build_device_model_arg
+         flexarray_append(dm_args, vncarg);
+     } /* OpenXT: no else here, we don't support "-vnc none" */
  
+-    /*
+-     * OpenXT: the default display backend is Surfman
```
jean-edouard: Seems like you're removing code from another patch here, wouldn't it make more sense to edit that other patch instead?

recipes-kernel/linux/4.14/defconfigs/xenclient-ndvm/defconfig
```diff
@@ -377,11 +381,13 @@ CONFIG_SCHED_OMIT_FRAME_POINTER=y
...
 CONFIG_PARAVIRT=y
 CONFIG_PARAVIRT_DEBUG=y
-# CONFIG_PARAVIRT_SPINLOCKS is not set
+CONFIG_PARAVIRT_SPINLOCKS=y
+# CONFIG_QUEUED_LOCK_STAT is not set
 CONFIG_XEN=y
 CONFIG_XEN_PV=y
 CONFIG_XEN_PV_SMP=y
-# CONFIG_XEN_PVHVM is not set
+CONFIG_XEN_PVHVM=y
```
jean-edouard: Why is this needed?

recipes-extended/xen/files/libxl-display-manager-support.patch
jandryuk: libxl-display-manager-support.patch was already touching code on top of libxl-openxt-qemu-args.patch.  My patches were developed on top of both, so they naturally modify the topmost one.

I could redo libxl-openxt-qemu-args.patch to move the hunks and then redo libxl-display-manager-support.patch on top.  Or rework the patch queue further to first add "display->kind" and then add the desired final hunk. That would be weird since the patch would add a field without a user.  If we always make display->kind default to "surfman", then we can avoid the conditional.

Do you want one of these changes, and do you have a preference which one?

classes/openxt-image-disk.bbclass
jean-edouard: Fair enough.

recipes-extended/xen/files/libxl-display-manager-support.patch
jean-edouard: Maybe we can rework that in a later PR, this alone is not enough of a concern to delay merging the whole thing.

recipes-kernel/linux/4.14/defconfigs/xenclient-ndvm/defconfig
jean-edouard: You're right, sorry, I read it as "PVH VM" :)

classes/openxt-image-disk.bbclass
jean-edouard: That's fine, it just looks odd to those who are not familiar with the original code.

jandryuk:
One other small item: NDVM cannot mount the FAT boot partition; it doesn't have the driver.  With a read-only disk, there is no real use.  Also, the kernel is installed both in the /dev/xvda1 partition, and in /boot on root partition /dev/xvda2.

jandryuk:
Re-pushed with https://github.com/OpenXT/xenclient-oe/pull/912/commits/83c48e3404f4bc65b5b77bf2728bdfed590526c4 changing the comments and fixing the whitespace in classes/openxt-image-disk.bbclass per @jean-edouard

eric-ch:
Build [$1558](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1558)

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1562

jean-edouard:
LGTM, merging soon.

PR #914, opened 2018-06-05
STABLE-8: OXT-1333, OXT-1346: Xen stable branch and latest XSA.

eric-ch:
- Fetch Xen sources from the git repository at the latest revisions required to backport the latest XSA. This also add the latest XSA up to XSA-262, as well as fixes pushed to the stable branch.
- Add XSA-263 to the patch-queue until it makes its way to 4.9.3-pre, or until RELEASE-4.9.3 is tagged.

Relates to:
- https://github.com/OpenXT/openxt/pull/313

### DIFF
--- a/recipes-extended/xen/files/libxl-openxt-qemu-args.patch
+++ b/recipes-extended/xen/files/libxl-openxt-qemu-args.patch
@@ -90,15 +90,7 @@ PATCHES
          if (libxl_defbool_val(b_info->u.hvm.gfx_passthru)) {
              switch (b_info->u.hvm.gfx_passthru_kind) {
              case LIBXL_GFX_PASSTHRU_KIND_DEFAULT:
-@@ -665,6 +657,7 @@ static const char *qemu_disk_format_stri
-     case LIBXL_DISK_FORMAT_QED: return "qed";
-     case LIBXL_DISK_FORMAT_HOST_CDROM: return "host_cdrom";
-     case LIBXL_DISK_FORMAT_HOST_DEVICE: return "host_device";
-+    case LIBXL_DISK_FORMAT_FILE: return "file";
-     default: return NULL;
-     }
- }
-@@ -930,19 +923,20 @@ static int libxl__build_device_model_arg
+@@ -915,19 +907,20 @@ static int libxl__build_device_model_arg
                        "-xen-domid",
                        GCSPRINTF("%d", guest_domid), NULL);
  
@@ -127,7 +119,7 @@ PATCHES
      for (i = 0; i < guest_config->num_channels; i++) {
          connection = guest_config->channels[i].connection;
          devid = guest_config->channels[i].devid;
-@@ -982,7 +976,7 @@ static int libxl__build_device_model_arg
+@@ -967,7 +960,7 @@ static int libxl__build_device_model_arg
      }
  
      if (c_info->name) {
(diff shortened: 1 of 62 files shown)


eric-ch:
@jandryuk Considering [Staging branches](https://wiki.xenproject.org/wiki/Xen_Project_Repositories) have not been put through the automated test system, it seems fair to assume they would be subject to rebases from time to time (which could trigger OpenXT build failure) in case of breakage. Back-porting the advertised XSA changes until pushed in the stable branch should avoid that issue. Ideally, I would like to follow stable-X.Y to have releases pointing to a tag to RELEASE-X.Y.Z as much as possible, but I do not mind switching to staging if that is a general consensus.

Also it seems fairly easy to script the back-port of the XSA changes using the .meta file to make sure we do not skip a version. It seems harder tracking the staging branches.

jandryuk:
You are right that we should stay on stable for the benefit of the automated testing.  There is some discussion on xen-devel about XSAs applying to staging, so they require backporting to point releases.  I had that in mind when I asked my question.  Your "stable branch + XSAs" may be the best way forward.  Thanks.

recipes-extended/xen/files/libxl-openxt-qemu-args.patch
```diff
@@ -90,15 +90,7 @@ PATCHES
          if (libxl_defbool_val(b_info->u.hvm.gfx_passthru)) {
              switch (b_info->u.hvm.gfx_passthru_kind) {
              case LIBXL_GFX_PASSTHRU_KIND_DEFAULT:
-@@ -665,6 +657,7 @@ static const char *qemu_disk_format_stri
-     case LIBXL_DISK_FORMAT_QED: return "qed";
-     case LIBXL_DISK_FORMAT_HOST_CDROM: return "host_cdrom";
-     case LIBXL_DISK_FORMAT_HOST_DEVICE: return "host_device";
-+    case LIBXL_DISK_FORMAT_FILE: return "file";
```
jean-edouard: Not needed?

jean-edouard:
Do you plan on doing something similar in master, or is this stable-8-specific?

recipes-extended/xen/files/libxl-openxt-qemu-args.patch
eric-ch: No, the function was moved to a different file which I did not refresh properly it seems. Good catch.

jean-edouard:
Makes sense.

eric-ch:
@jean-edouard https://github.com/OpenXT/xenclient-oe/pull/914/commits/753cadab6da41eddc27b6b58060eedcd2bcabd5c#diff-6e777734ac8ca862f146cf5dd681ddacR210, much better.

Refreshed the patch-queue as some white-spaces were still in there.

eric-ch:
Rebased again with some changes:
- Added `xen-version.inc` to get Xen's source versions from a common place between `xen-common.inc` and `xen-xsm-policy.bb`.
- Changed `xen-xsm-policy.bb` and `xen-common.inc` to use `xen-version.inc` to find Xen sources named version, source revision and branch.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1556

jean-edouard:
LGTM, merging soon.

PR #915, opened 2018-06-05
STABLE-8: customtask: python reduce was moved in functools.

eric-ch:
xenclient-customtask does not work with Python 3 recent versions:
https://docs.python.org/3/library/functools.html

(cherry picked from commit 1e0472694326152af785cae12999d13fc2a7f52c)


jean-edouard:
Merging soon.

PR #917, opened 2018-06-07
OXT-1319: usbback: Fix dom0 crash when assigning greedy devices.

eric-ch:
usbback can be led to BUG() if allocated pages to handle requests from the frontend get exhausted. This is easy to achieve using isochronous devices (webcam, microphones?) that will use indirect requests.
The BUG() crash can be avoided altogether. Increasing the pool of pages should take care of most devices. This is already a module parameter, these changes double the default value.

### DIFF
--- a/recipes-kernel/linux/4.14/patches/usbback-base.patch
+++ b/recipes-kernel/linux/4.14/patches/usbback-base.patch
@@ -1364,7 +1364,7 @@ PATCHES
 +
 +#include "common.h"
 +
-+static int usbif_reqs = 64;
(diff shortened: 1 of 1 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1551

jean-edouard:
LGTM, fixed my camera with Windows 10 guests, merging soon.
@eric-ch stable-8 PR?

eric-ch:
Done: https://github.com/OpenXT/xenclient-oe/pull/919

PR #918, opened 2018-06-08
bash: move root history to /tmp

jean-edouard:

### DIFF
--- a/recipes-extended/bash/bash/root.bashrc
+++ b/recipes-extended/bash/bash/root.bashrc
@@ -1,0 +1,1 @@
+export HISTFILE=/tmp/.bash_history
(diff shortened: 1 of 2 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1562

jean-edouard:
For some reason, `/root/.bashrc` doesn't get labeled properly, maybe because the relabeling skips hidden files?...

jean-edouard:
Yeah it's not in the policy should it should inherit the domain of the parent directory.

jean-edouard:
Doesn't work, closing.

PR #921, opened 2018-06-12
recipes-extended: fix seabios URL

Jafaral:
Somebody decided to move where they host their files

Signed-off-by: Jafar Al-Gharaibeh <jafar@atcorp.com>


eric-ch:
This is getting old... the URL last failed Dec 6 14:52:00 2017 reading the git logs.
People without an updated cache will get impacted, maybe we should add the https://openxt.ainfosec.com/mirror/ mirrors to avoid periodic failures.

Jafaral:
@eric-ch  pushed stable-8 PR.  

I like the idea of a mirror if the url ontinues to break.

PR #922, opened 2018-06-12
STABLE-8: recipes-extended: fix seabios URL

Jafaral:
Signed-off-by: Jafar Al-Gharaibeh <jafar@atcorp.com>


eric-ch:
master: https://github.com/OpenXT/xenclient-oe/pull/921

PR #925, opened 2018-06-14
OXT-1351: Change PCR Sealing selection on upgrade

jandryuk:
Currently, /config/config.pcrs is only written during installation. If we are upgrading, we must write out /config/config.pcrs to ensure new values are used if the PCR selection changed.

configure_measured_launch() is only called on install, so changes to the
pcr sealing set don't take affect on upgrate.  Split out writing
/config/config.pcrs so it can be used on upgrade.

OXT-1351

### DIFF
--- a/recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
+++ b/recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
@@ -511,6 +511,19 @@ mle_success() {
     return $ret
 }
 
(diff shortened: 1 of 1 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1570

jean-edouard:
LGTM, will merge after the stable-8 PR is created.

PR #933, opened 2018-06-19
OXT-1361: dbus: Avoid segfault when stopping dbus-daemon.

eric-ch:
Running /etc/init.d/dbus-1 stop will segfault consistently on OpenXT:
```
dbus-daemon[728]: segfault at 0 ip b7eb170a sp bf8468a8 error 6 in libdbus-1.so.3.14.9[b7e83000+5a000]
```

Here is a trace:
```
Thread 1 "dbus-daemon" received signal SIGSEGV, Segmentation fault.
0xb7ee970a in _dbus_list_unlink (list=0x8c78580, link=0x8c8e200) at /usr/src/debug/dbus/1.10.14-r0/dbus-1.10.14/dbus/dbus-list.c:510
510           link->next->prev = link->prev;
(gdb) bt
```

It seems that bus_context_unref can re-use a cleared list in its path:
```
-> bus_context_unref
  -> bus/connection.c:544: _dbus_list_clear (&connections->monitors);
  -> bus/connection.c:555: bus_connection_disconnected (connection);
    -> bus/connection.c:305: _dbus_list_remove_link (&d->connections->monitors, d->link_in_monitors);
         d->connections->monitors is NULL since _dbus_list_clear since connections == d->connections apparently.
```

From the previous gdb session:
```
305           _dbus_list_remove_link (&d->connections->monitors, d->link_in_monitors);
(gdb) p d->connections
$19 = (BusConnections *) 0x8c78558
555               bus_connection_disconnected (connection);
(gdb) p connections
$22 = (BusConnections *) 0x8c78558
```

`d->link_in_monitors` in this case is a stray pointer to a list that has
been cleared.

### DIFF
--- a/recipes-core/dbus/patches/fix-segfault-bus_connection_disconnected.patch
+++ b/recipes-core/dbus/patches/fix-segfault-bus_connection_disconnected.patch
@@ -1,0 +1,13 @@
+--- a/bus/connection.c
++++ b/bus/connection.c
+@@ -302,7 +302,9 @@ bus_connection_disconnected (DBusConnect
+       if (mm != NULL)
+         bus_matchmaker_disconnected (mm, connection);
+ 
+-      _dbus_list_remove_link (&d->connections->monitors, d->link_in_monitors);
++      if (d->connections->monitors != NULL)
++        _dbus_list_remove_link (&d->connections->monitors, d->link_in_monitors);
++
+       d->link_in_monitors = NULL;
+     }
+ 
--- a/recipes-core/dbus/dbus_1.%.bbappend
+++ b/recipes-core/dbus/dbus_1.%.bbappend
@@ -16,6 +16,7 @@ SRC_URI += " \
     file://0001-Make-the-default-DBus-reply-timeout-configurable.patch \
     file://add-domid-authentication.patch \
     file://v4v.patch \
+    file://fix-segfault-bus_connection_disconnected.patch \
 "
 
 do_install_append() {


recipes-core/dbus/patches/fix-segfault-bus_connection_disconnected.patch
```diff
@@ -0,0 +1,13 @@
+--- a/bus/connection.c
++++ b/bus/connection.c
+@@ -302,7 +302,9 @@ bus_connection_disconnected (DBusConnect
+       if (mm != NULL)
+         bus_matchmaker_disconnected (mm, connection);
+ 
+-      _dbus_list_remove_link (&d->connections->monitors, d->link_in_monitors);
++      if (d->connections->monitors != NULL)
++        _dbus_list_remove_link (&d->connections->monitors, d->link_in_monitors);
```
jandryuk: I think this will now leak link_in_monitors.  `_dbus_list_remove_link` does `_dbus_list_unlink` and `free_link`.  So maybe
```
if (d->connections->monitors != NULL)
    _dbus_list_unlink (&d->connections->monitors, d->link_in_monitors);

free_link (d->link_in_monitors);
d->link_in_monitors = NULL;
```

I looked at upstream dbus, and they haven't changed this code.  Why are we triggering it but not them?

eric-ch: Yes, I checked upstream DBus too when debugging this and could not find a fix for that problem between 1.10.14 and 1.10.26. None of the patches applied in OpenXT seem relevant to that behavior either... This may be an assumption that is not honored, but I have not been able to figure out where.

- `_dbus_clear_list` will call `free_link` on each link of the list;
- `bus/connection.c:544: _dbus_list_clear (&connections->monitors)` will release `connections->monitors` list;
- Since `d->connections->monitors` in `bus_connection_disconnected` is `connections->monitors` in `bus_connections_unref`) and `d->link_in_monitors` is an element of that list it should be free'ed when run `bus_connection_disconnected` I think?

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1574

recipes-core/dbus/patches/fix-segfault-bus_connection_disconnected.patch
jandryuk: Yes, your analysis looks correct.  So it seems like link_in_monitors should be NULLed when removing its entry from connections->monitors at bus/connection.c:544.  But that would mean bus_matchmaker_disconnected doesn't get run.

eric-ch:
stable-8 PR: https://github.com/OpenXT/xenclient-oe/pull/938

PR #934, opened 2018-06-20
STABLE-8: OXT-1351: Change PCR Sealing selection on upgrade

jandryuk:
This is the stable-8 version of https://github.com/OpenXT/xenclient-oe/pull/925

Currently, /config/config.pcrs is only written during installation. If we are upgrading, we must write out /config/config.pcrs to ensure new values are used if the PCR selection changed.

configure_measured_launch() is only called on install, so changes to the
pcr sealing set don't take affect on upgrate.  Split out writing
/config/config.pcrs so it can be used on upgrade.

OXT-1351

(cherry picked from commit a9e0ffaa20a5a52d29acfcc71bd0163da3705208)

### DIFF
--- a/recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
+++ b/recipes-openxt/openxt/openxt-measuredlaunch/ml-functions
@@ -511,6 +511,19 @@ mle_success() {
     return $ret
 }
 
+write_config_pcrs() {
+    local root="${1:-/}"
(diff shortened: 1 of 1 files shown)


jean-edouard:
Merging soon.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1573

PR #939, opened 2018-06-21
OXT-136{4,5,6,7}: avc denials. 

eric-ch:
Backport one upstream modification to the refpolicy-mcs and attend a few other AVC that regularly trigger on OpenXT.


eric-ch:
Nothing breaking afaict, but some programs might use fallback mechanisms because of the EPERM:
- `lvmetad.socket` would have lvm-utils fallback to not using the daemon, which is not critical, disks and logical volumes will be scanned again.
- `updatemgr`, just create log spam as it seems to be assuming default values.
- `input_server`, not quite sure, this may be in the way of input detection, but I did not observe any failure.
- `kmod_t` this is just noise.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1579

jean-edouard:
Merging soon.

PR #940, opened 2018-06-22
OXT-1363: linux: Allow disabling txt event log in Kconfig

jandryuk:
Without text, the tristate option is not displayed in menuconfig and
cannot be disabled. Add text to allow disabling, but also default to
disabled.  Also, slightly improve the help text.

The Dom0 and Installer kernels already select this in their respective
defconfigs, so it doesn't change for them.

OXT-1363


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1579

eric-ch:
Will merge soon.

PR #945, opened 2018-06-28
STABLE-8: seal-system: fix check for all-zeros PCRs

jean-edouard:
(cherry picked from commit e787b872bb4a35f64225806163c3fe8568ce22a1)

### DIFF
--- a/recipes-openxt/openxt/openxt-measuredlaunch/seal-system
+++ b/recipes-openxt/openxt/openxt-measuredlaunch/seal-system
@@ -23,6 +23,14 @@ err() {
     exit 1
 }
 
+contains_only() {
+    local s="$1"
+    local c="$2"
+
+    # Remove all instances of c from s and check if the result is empty
+    [[ -z "${s//$c}" ]]
+}
+
 # Globals
 TPM_DEV="$(find /sys/class -name tpm0)/device"
 
@@ -134,7 +142,7 @@ forward)
         pcr8=$(grep PCR-08 ${TPM_DEV}/pcrs | cut -f2 -d: | tr -d ' ')
     fi
 
-    if [[ "${pcr8}" = *[!0]* ]]; then
+    if ! contains_only "${pcr8}" 0; then
         ESP=$(sfdisk -l 2>/dev/null | grep "EFI System" | awk '{print $1}')
         mkdir -p /tmp/esp
         mount -o ro $ESP /tmp/esp || err "Failed to mount ESP ${ESP}"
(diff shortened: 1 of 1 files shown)


jean-edouard:
Nope. This is just something I found while reading seal-system to try and figure out why forward sealing fails. This PR on its own doesn't fix forward sealing, just a globbing expression that didn't match its comment :)

recipes-openxt/openxt/openxt-measuredlaunch/seal-system
```diff
@@ -134,7 +141,7 @@ forward)
         pcr8=$(grep PCR-08 ${TPM_DEV}/pcrs | cut -f2 -d: | tr -d ' ')
     fi
 
-    if [[ "${pcr8}" = *[!0]* ]]; then
```
tklengyel: This does actually work as expected:

```
#!/bin/bash

pcr8="000000000000000000000000000000000000000"

if [[ "${pcr8}" = *[!0]* ]]; then
    echo "not all 0"
else
    echo "all 0"
fi

pcr8="000000000000000000000000000000000000001"

if [[ "${pcr8}" = *[!0]* ]]; then
    echo "not all 0"
else
    echo "all 0"
fi
```
Produces:
```
all 0
not all 0
```

recipes-openxt/openxt/openxt-measuredlaunch/seal-system
```diff
@@ -188,7 +195,7 @@ forward)
     pcr_params="${pcr_params} -p 15:${pcr15}"
 
     # Calculate DRTM PCRs only when booting legacy (ie. PCR8 is all 0)
-    if [[ "${pcr8}" = *[0]* ]]; then
```
tklengyel: This indeed is buggy, should have been `if [[ ! "${pcr8}" = *[!0]* ]]; then`

eric-ch:
Build: [$1590](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1590)

recipes-openxt/openxt/openxt-measuredlaunch/seal-system
eric-ch: If there is no strong feeling for one ( `-z "${var//0}"`) or the other (`[[ ! "${var}" = *[!0]* ]]`) and since this PR is out both in master and stable-8, I think this is good to be merged soon?

tklengyel: TBH would be more useful to introduce a function that checks if a string is all a given character, not just 0, so that it could be used to check if drtm PCRs are set (which are all `f`s by default).

jean-edouard:
The test function is now generic.

eric-ch:
Will merge soon.

PR #947, opened 2018-06-28
STABLE-8: Use .inputrc for bash/readline shortcuts

jandryuk:
Stable-8 version of https://github.com/OpenXT/xenclient-oe/pull/943

OE packages /etc/inputrc in the readline recipe, so we overrode that
file to implement the bash keyboard shortcuts.  While readline was
installed in xenclient-dom-image at the time of submission, it was only
as an implicit dependency on python's readline support.  When semodule
was fixed to no longer install python, readline and /etc/inputrc were no
longer installed.

We could bbappend bash to install /etc/inputrc, but that would
eventually conflict if readline were installed.  We could install
readline explicitly, but we don't need it.  Or we could avoid the issue
by just installing to /root/.inputrc.  We'll use the user-specific
option since that is good enough for our use case.

Fix some whitespace in xenclient-dom0-tweaks.inc while at it.

OXT-1309

(cherry picked from commit a570b92075a2c81708c780477adb641ce2363959)


eric-ch:
Will merge soon.

PR #948, opened 2018-06-29
STABLE-8: OXT-1370: XSA-264, XSA-265, XSA-266

eric-ch:
- Update stable-4.9 tracked revision as it now includes XSA-267
- Backport XSA-264 & XSA-265
- Backport XSA-266. It will enforce read-write configuration for IDE disk assigned to the guest as QEMU emulation does not support read-only. OpenXT has a [patch](https://github.com/OpenXT/xenclient-oe/blob/master/recipes-openxt/qemu-dm/qemu-dm/readonly-ide.patch) to handle that case, so remove the `assert()` introduced for IDE disks configuration in that XSA.


eric-ch:
Build [1590](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1590)

jean-edouard:
Merging soon.

PR #950, opened 2018-07-02
STABLE-8: OXT-1329: Refactor UIVM process management scripts.

eric-ch:
Master PR: https://github.com/OpenXT/xenclient-oe/pull/949

### nm-applet
There is one nm-applet per NetworkManager running in a Network domain. Since requests between nm-applet and NetworkManager are exchanged between UIVM and Network domain over V4V, each nm-applet has to follow the life-cycle of its backend.

This worked (somehow) before the upgrade to Pyro with a couple of `while true` loops.

Instead of relying on processes to stop, which nm-applet no longer seem to do, and restart in a changed environment, factorise the process management in one agent script that will listen for DBus events over V4V and take appropriate actions for the processes it tracks.

### uim-toolbar
Before refactoring, two scripts handled locale synchronisation with changes made in XenMgr. This can be done in an agent script like nm-applet, just simpler, since there is only one `uim-toolbar` process in UIVM.

Some functions can be shared with other agent scripts monitoring events on DBus or reading configuration from XenMgr in dom0.

### keymap
Like uim-toolbar, make a dedicated agent script to manage the process life-cycle and updates.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1601

jean-edouard:
Merging soon.

PR #951, opened 2018-07-02
sshd: disable "none" authentication module

jean-edouard:
The auth2-none is responsible for handling the PermitEmptyPasswords option.
Whenever that option is set, sshd will try to login all users without a passwords.
For users that do have a password, that means a failed login attempt and a 2 seconds retry timeout.
auth-passwd can handle empty password, clients can just press enter when prompted for a passsword.
Without auth2-none, users that do have a password won't have to wait anymore.

### DIFF
--- a/recipes-connectivity/openssh/openssh/disable-auth2-none.patch
+++ b/recipes-connectivity/openssh/openssh/disable-auth2-none.patch
@@ -1,0 +1,11 @@
+diff -u a/auth2.c b/auth2.c
+--- a/auth2.c	2018-07-02 14:36:45.795787892 -0400
++++ b/auth2.c	2018-07-02 14:36:57.175788121 -0400
+@@ -74,7 +74,6 @@
+ #endif
+ 
+ Authmethod *authmethods[] = {
+-	&method_none,
+ 	&method_pubkey,
+ #ifdef GSSAPI
+ 	&method_gssapi,
--- a/recipes-connectivity/openssh/openssh_7.5p1.bb
+++ b/recipes-connectivity/openssh/openssh_7.5p1.bb
@@ -25,6 +25,7 @@ SRC_URI = "http://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-${PV}.tar
            file://openssh-7.1p1-conditional-compile-des-in-cipher.patch \
            file://openssh-7.1p1-conditional-compile-des-in-pkcs11.patch \
            file://fix-potential-signed-overflow-in-pointer-arithmatic.patch \
+           file://disable-auth2-none.patch \
            file://sshd_check_keys \
            "
 


jean-edouard:
I didn't think "none" was considered an authentication method but I'll check!

jean-edouard:
Very interesting...
First, AuthenticationMethods did not appear in the `man sshd_config` I was looking at, but it is indeed understood by sshd.
Second, when using `AuthenticationMethods password`, if in `/etc/shadow` root has a hash for an empty password, like `$1$aIKg1Jne$qMOTnqQ8VxV5md/xTPx.V/`, then any password will log you in, except for the empty string! Something is seriously wrong there...

jandryuk:
That is bizarre.  If it is actually hashing the input, how can that work?

When do you see the 2 second delay?  I do not see it, but I always have a password set.

jean-edouard:
I see the 2 second delay with a passworded vanilla install of OpenXT.
I tried 3 different distros and the same thing happens: as soon as PermitEmptyPassword is set to yes in `/etc/ssh/sshd_config`, a failed login attempt is logged on the server side before the client gets a password prompt (which takes 2 seconds to show up because of the retry delay).

eric-ch:
There should not be any immediate risk to disabling "none" authentification method.
Since this is working-around an upstream quirk, that LGTM for now, but that should be investigated further.

Tracked by: OXT-1389

jandryuk:
On a master build with https://github.com/OpenXT/xenclient-oe/pull/973, I don't see any delay or failed login attempts.  I installed interactively and set a password.

jean-edouard:
@jandryuk, could you please make sure you're running OpenSSH 7.5p1 and paste your config file (/etc/ssh/sshd_config_readonly) somewhere? Or repro with a build from openxt.ainfosec.com?

This is quite baffling to me, since I tried our config on multiple Linux distributions, got the same result everywhere, and confirmed by looking at the code that auth_none is always the first module, and when PermitEmptyPasswords is set to yes, auth_none tries a password-less login...

eric-ch:
@jandryuk I only see the delay if `PermitEmptyPasswords` is set to `yes`. This is not default if `allow-empty-password` or `debug-tweaks` is not set in the `IMAGE_FEATURES`.

I believe the public builder sets `allow-empty-password` explicitly.

jean-edouard:
Correct, the whole issue is triggered by PermitEmptyPassword, which we need for unattended installs.

jandryuk:
Oh, so, yes, I get the delay and "pam_unix(sshd:auth): authentication failure" for ssh over TCP.  Over v4v, I do not see a delay.

snippet of `diff -u /etc/ssh/sshd_config_readonly{,_v4v}`
```
@@ -62,11 +53,11 @@
 #IgnoreRhosts yes
 
 # To disable tunneled clear text passwords, change to no here!
-#PasswordAuthentication yes
+PasswordAuthentication no
 PermitEmptyPasswords yes
 
 # Change to no to disable s/key passwords
-ChallengeResponseAuthentication no
+ChallengeResponseAuthentication yes
 
 # Kerberos options
 #KerberosAuthentication no
```
Why do these differ?

jandryuk:
Using `PasswordAuthentication no` & `ChallengeResponseAuthentication yes` eliminates the delay and authentication failure message.

jean-edouard:
Nice find @jandryuk! That seems to fix the issue, without having to patch the service.
Let's use that instead, let me know if you want me to PR it.

jandryuk:
@jean-edouard, please do the PR.  Thanks!

jean-edouard:
Replaced by https://github.com/OpenXT/xenclient-oe/pull/979. Closing.

PR #958, opened 2018-07-06
OXT-1364: lvm: lvmetad context is lvm_exec_t

eric-ch:
This will allow proper transition when creating pidfile.

Silence `lvmetad.socket` related AVCs.


jean-edouard:
LGTM, merging soon.

PR #960, opened 2018-07-09
refpolicy: allow modprobe to search the securityfs to load txt

jean-edouard:
During its initialization, our "txt" Linux module searches the securityfs.
Therefore, modprobe (insmod) needs to be able to do that.


jandryuk:
You commit message mentions search, but the patch adds both search and mount.  Is mount still needed after https://github.com/OpenXT/xenclient-oe/pull/962 ensures securityfs is mounted?

I'd also be nice to include the actual avc denial in the commit message.

jean-edouard:
@jandryuk the "mount" line you saw is just context, I did not add that.
The avc is really hard to get since it's dontaudit-ed and for some reason doesn't always show up even with dontaudits disabled (even if the module always fails to load)... But it's something like kmod_t trying to search a security_t directory.

jandryuk:
Ah, okay.  Yup, I mis-read the diff of a diff.

PR #962, opened 2018-07-09
fstab.early: mount the securityfs for the txt module

jean-edouard:


recipes-core/base-files/files/fstab.early
```diff
@@ -14,3 +14,6 @@ tmpfs       /tmp                    tmpfs       defaults,rootcontext=system_u:ob
 tmpfs       /var/volatile           tmpfs       defaults,rootcontext=system_u:object_r:var_t:s0,size=2M         0 0
 tmpfs       /var/cache              tmpfs       defaults,rootcontext=system_u:object_r:var_t:s0,size=100M       0 0
 tmpfs       /var/lock               tmpfs       defaults,rootcontext=system_u:object_r:var_lock_t:s0,size=1M    0 0
+
+# OpenXT: modutils.sh loads before mountall.sh, and modutils.sh will load the txt module, which needs this:
+securityfs                     /sys/kernel/security    securityfs  defaults                                                                        0 0
```
eric-ch: Just as a note, although the comment is in the [/etc/fstab](https://github.com/OpenXT/xenclient-oe/blob/master/recipes-core/base-files/files/xenclient-dom0/fstab#L8), this line has to match the equivalent existing one in the fstab (or be removed from the fstab).

eric-ch:
Will merge soon.

PR #966, opened 2018-07-11
OXT-1378: installer: Remove tapback-daemon initscript.

eric-ch:
OpenXT installer does not have CONFIG_BLK_DEV_TAP required to start the
daemon. This is not fatal, and will only log at startup:
```
tapback: tapback.c:612 Unable to find blktap2 device major
```
[...] and shutdown:
```
Stopping tapback: no /usr/bin/tapback found; none killed
```

### DIFF
--- a/recipes-core/images/xenclient-installer-image.bb
+++ b/recipes-core/images/xenclient-installer-image.bb
@@ -99,6 +99,7 @@ ROOTFS_POSTPROCESS_COMMAND += "post_rootfs_shell_commands; "
 # Remove initscripts pulled-in by dependencies or not required for operation.
 remove_nonessential_initscripts() {
     remove_initscript "blktap"
+    remove_initscript "tapback-daemon"
     remove_initscript "sshd-v4v"
 }
 ROOTFS_POSTPROCESS_COMMAND += "remove_nonessential_initscripts; "


jandryuk:
Should we just not install the blktap packages, or is there a dependency issue?

jandryuk:
With the 64bit openxt-installer image, blktap3 is broken, right?  Is blktap2 also broken?  If so, that means only "vhdgz" vm disk images can be installed: https://github.com/OpenXT/installer/blob/master/part2/stages/Functions/install-main#L761

eric-ch:
> Should we just not install the blktap packages

I thought the installer might need some of the tap utils, but these would also need the `CONFIG_BLK_DEV_TAP`... So yes that sounds like a better idea.

> With the 64bit openxt-installer image, blktap3 is broken, right? Is blktap2 also broken? 

It surely seems like it, since `CONFIG_BLK_DEV_TAP` is only set in `xenclient-dom0` image.

> If so, that means only "vhdgz" vm disk images can be installed.

That would be right.

This goes deeper than I initially thought (just a warning). I'll create a new ticket for it.

jean-edouard:
This is only temporary, we'll be able to use blktap in the installer again once haskell is gone and we switch to full 64bit mode. Therefore, I suggest we don't waste more time on this.

The PR LGTM, I built and tested it on stable-8.
I will merge it on Monday if that works for everybody.

jandryuk:
Isn't the installer already 64bit?  I didn't think the installer included any Haskell components.

openxt-installer required moving tap-ctl operations to firstboot because "32-bit toolstack has trouble interacting with 64-bit kernel."  Would enabling `CONFIG_BLK_DEV_TAP` just fix the issue if we are already 64bit?

jean-edouard:
@jandryuk the installer kernel is 64 bits, but the rootfs is 32 bits (except for a couple EFI tools).

jandryuk:
That was chagned by @eric-ch in d2591307e8d49c645540a8bc392d0388ef212030
```
openxt-installer: Make the machine core2-64.
    
Since it builds a 64bit kernel, it might as well be 64bit entirely and
provide the right libc for 64bit userland components.
    
```

jean-edouard:
Oh I completely forgot about that...
So yeah, maybe blktap can actually be used in the installer now!

jandryuk:
I tested out enabling `CONFIG_BLK_DEV_TAP` in openxt-installer, and it seems to work.  `tap-ctl create` created `/dev/xen/blktap-2/tapdev0` which `mkswap` did not complain about.

eric-ch:
@jandryuk Thanks for testing it! :)

Can you submit the PR? I will close this one.

eric-ch:
@jandryuk Thanks again!

Closing this PR in favor of https://github.com/OpenXT/xenclient-oe/pull/969

PR #967, opened 2018-07-11
STABLE-8: OXT-1353: installer: Remove tapback-daemon initscript.

eric-ch:
OpenXT installer does not have CONFIG_BLK_DEV_TAP required to start the
daemon. This is not fatal, and will only log at startup:
tapback: tapback.c:612 Unable to find blktap2 device major
[...] and shutdown:
Stopping tapback: no /usr/bin/tapback found; none killed


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1614

eric-ch:
Closing in favor of https://github.com/OpenXT/xenclient-oe/pull/967

PR #968, opened 2018-07-16
OXT-1381: Remove dm-agent

jandryuk:
dm-agent has been unused since the xl conversion - qemu is launched directly now.  svirt-interpose is hardcoded to call qemu-dm-wrapper, a script from the dm-agent.git repo.  qemu-dm-wrapper set up alsa configuration for qemu, so move it into the qemu-dm recipe.  After that, we can remove dm-agent and it's SELinux policy.


jean-edouard:

jean-edouard:
Merging soon.

PR #969, opened 2018-07-16
OXT-1378: Restore installer blktap operations

jandryuk:
blktap tap-ctl operations were moved from the installer to firstboot.sh when enabling UEFI host support. tap-ctl broke when running as a 32bit compat process on a 64bit kernel.  In the meantime,
the openxt-installer machine was converted to a 64bit userspace, so the compatibility issue is no longer present.

Set CONFIG_BLK_DEV_TAP=y so the blktap tap-ctl command works in the
installer again, and move the tap-ctl operations back to the installer.  

Depends on https://github.com/OpenXT/installer/pull/94


jean-edouard:
LGTM, merging soon.

jandryuk:
With the merge of https://github.com/OpenXT/xenclient-oe/pull/982 to switch to a linux installer, I'll just close this and https://github.com/OpenXT/xenclient-oe/pull/970.

PR #971, opened 2018-07-17
qemu-dm: Avoid segfault in dmbus_send

jandryuk:
acpi.patch adds this call chain:
acpi_pm1_cnt_write
  xenstore_update_power
    xen_input_send_shutdown
      dmbus_send

When xen_input has not been initialized, dmbus_send segfaults since
input.service is NULL.  If we aren't using xen-input, then we don't need
to notify input_server.  That way we avoid the segfault.

Maybe this should be moved to xen_input_deinit?  In that case, we
wouldn't have the shutdown reason on hand - it would need to be stashed
somewhere.

OXT-1360

### DIFF
--- a/recipes-openxt/qemu-dm/qemu-dm/switcher.patch
+++ b/recipes-openxt/qemu-dm/qemu-dm/switcher.patch
@@ -115,7 +115,7 @@ PATCHES
  x_keymap.o-cflags := $(X11_CFLAGS)
 --- /dev/null
 +++ b/ui/xen-input.c
-@@ -0,0 +1,427 @@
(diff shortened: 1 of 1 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1622

jean-edouard:
LGTM, merging soon.

eric-ch:
@jandryuk This call chain seems to be present in stable-8 QEMU as well. If that is correct, could you please send a PR against stable-8 QEMU 2.6.2 as well?

PR #972, opened 2018-07-19
OXT-1383: xen: Bypass ACPI SLIC support for PVH

jandryuk:
This is a fix for libxl to support PVH.  xenmgr requires changes to generate a PVH xl.cfg.

acpi-slic-support.patch changes acpi_build_tables to modify the OEM ID
so a passed through SLIC table works for Windows activation.  It does so
by modifying the ACPI tables on the fly.  This works fine when run as
part of hvmloader, but it segfaults when the code runs in the xl process
for a PVH domain.

The xl segfault is in is_slic() trying to accessing memory in the range
0xfc00xxxx, ACPI_INFO_PHYSICAL_ADDRESS.  acpi_build_tables calls v2p to
convert pointers to the destination physical address.  In hvmloader
this is a no-op, so continued access through those pointers works.  With
xl for a PVH domain, it does an actual conversion to the destination
ACPI memory range which is invalid for the userspace process.

A SLIC entry can only be present if it is passed in, and it will never
be passed in for a PVH.  So conditionalize execution on whether a table
is passed in or not to avoid the fault.

While here, switch to using nr_secondaries to limit access to only the
populated entries of secondary_tables.

OXT-1383

### DIFF
--- a/recipes-extended/xen/files/acpi-slic-support.patch
+++ b/recipes-extended/xen/files/acpi-slic-support.patch
@@ -117,22 +117,25 @@ PATCHES
  
      acpi_info = (struct acpi_info *)config->infop;
      memset(acpi_info, 0, sizeof(*acpi_info));
-@@ -640,6 +702,24 @@ int acpi_build_tables(struct acpi_ctxt *
+@@ -640,6 +702,27 @@ int acpi_build_tables(struct acpi_ctxt *
      if ( nr_secondaries < 0 )
          goto oom;
  
-+    /* Check to see if one of the secondary tables is a SLIC. */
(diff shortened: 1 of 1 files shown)


jandryuk:
I'm using this patch to enable PVH in the short term: https://github.com/jandryuk/manager/commit/a7bee5490777922c46fdfeda151abb4691e72d15
Make sure you set stubdom to false to avoid fireworks.

jean-edouard:

eric-ch:
I could not find a regression with the fix, since the changes address futur PVH support, it may as well be merged ahead.

jean-edouard:

PR #973, opened 2018-07-19
[efi] Enable the xen efi path for the dom0 kernel

crogers1:
OXT-1307


eric-ch:
@crogers1 Can you open a PR against stable-8 as well please?

Initial test in build [$1610](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1610)

### Requires:
- https://github.com/OpenXT/xctools/pull/47
- https://github.com/OpenXT/installer/pull/96

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1623

crogers1:
@jandryuk I can look into it.  One of my first approaches tried messing with Kconfig dependencies, but maybe I was just looking at the wrong ones.  Tomorrow at the earliest is when I can get to this unfortunately.

jandryuk:
I'm trying out a Kconfig approach now.

jandryuk:
I used a patch that only changed the Kconfig to enable XEN_EFI.  efi.o was built, so it was effective in that respect.  EFI_64BIT is enabled unconditionally in xen_efi_init, so dropping the changes in drivers/firmware/efi/efi.c is okay.  I guess that enables other EFI_64BIT changes that we may not want - I haven't fully investigated.

Unfortunately, it did not work.  On an HP laptop with 8GB ram, I see "efi: EFI data located above 4GB, disabling EFI."  But that failure would happen even with your patch, right?

recipes-kernel/linux/4.14/patches/efi-tables-for-dom0.patch
```diff
@@ -0,0 +1,156 @@
...
+-obj-$(CONFIG_XEN_EFI)		+= efi.o
+ obj-$(CONFIG_XEN_PVH)	 	+= xen-pvh.o
+--- a/arch/x86/xen/efi.c
++++ b/arch/x86/xen/efi.c
+@@ -127,9 +127,8 @@ void __init xen_efi_init(void)
+ 	strncpy((char *)&boot_params.efi_info.efi_loader_signature, "Xen",
+ 			sizeof(boot_params.efi_info.efi_loader_signature));
+ 	boot_params.efi_info.efi_systab = (__u32)__pa(efi_systab_xen);
+-	boot_params.efi_info.efi_systab_hi = (__u32)(__pa(efi_systab_xen) >> 32);
++	boot_params.efi_info.efi_systab_hi = 0;
```
jandryuk: This looks bogus.  efi_systab_xen is a linux data structure.  In a 32bit build, it can't live above 4GB, so it should be unnecessary.

recipes-devtools/dmidecode/dmidecode/use-oe-vars.patch
```diff
@@ -0,0 +1,12 @@
+--- a/Makefile	2018-04-25 15:44:51.431420129 -0400
++++ b/Makefile	2018-04-25 15:36:35.327601950 -0400
+@@ -12,8 +12,7 @@
+ #   (at your option) any later version.
+ #
+ 
+-CC      = gcc
```
jandryuk: Changing to ?= would allow our CC override to occur without breaking a normal build, if the patch is desired for upstreaming.

This dmidecode change should be in a separate commit with a rational.  Why do we need it?

crogers1:
There are EFI_64BIT cases we definitely do _not_ want, one example is in arch/x86/platform/efi/efi.c:efi_systab_init.  We ask xen for the efi systab through a hypercall, which means the struct is addressable in dom0 because of copy_to_user, so we want the efi_systab_init to take the 32 bit path in the else-case.  If EFI_64BIT is set unilaterally, you'll get the "**EFI data located above 4GB disabling EFI**" message because of the #ifdef when in reality, the boot_params.efi_info.efi_systab (from Xen) is not above 4GB.

When we check each individual table in drivers/firmware/efi/efi.c:efi_config_parse_tables, if the table itself is above 4GB (which is entirely possible), we disable EFI at that point, and there's a slightly different message: "**Table located above 4GB, disabling EFI.**"

So to answer your question, I think if you tried my patch, you would _not_ see the "EFI data located above 4GB, disabling EFI" message, and the kernel would at least try to enumerate the tables, which may or may not work depending on where they are located in the firmware.

recipes-kernel/linux/4.14/patches/efi-tables-for-dom0.patch
crogers1: On a 32 bit arch, shifts of 32 bit size are undefined, sometimes they return the original variable unshifted. (ie. 5 >> 32 == 5).  I set it to 0 just to be safe.

recipes-devtools/dmidecode/dmidecode/use-oe-vars.patch
crogers1: It's using the host machine's gcc when it should be using the Open Embedded gcc built through the toolchain (ie. i686-oe-linux-gcc)

recipes-kernel/linux/4.14/patches/efi-tables-for-dom0.patch
jandryuk: Ah, I naively assumed __pa was returning a 64bit value.  So it's a bug in linux.

recipes-devtools/dmidecode/dmidecode/use-oe-vars.patch
jandryuk: Right.  Does compiling with the host gcc cause a particular issue that i686-oe-linux-gcc avoids?

jandryuk:
I'll try your patch to see how it works.

recipes-kernel/linux/4.14/patches/efi-tables-for-dom0.patch
crogers1: Yea I suppose they never considered the case of running the xen_efi_init codepath on a 32-bit dom0.

recipes-devtools/dmidecode/dmidecode/use-oe-vars.patch
crogers1: I actually wrote this patch awhile ago, so I can't remember off the top of my head, but if you give me a few minutes to track it down I can definitely get an answer for ya.

jandryuk:
With your patch, the laptop boots and succeed sealing on first boot.  However, on reboot, it hits 
"Measured Launch Unseal FAILED".  Resealing then boots successfully.

crogers1:
@eric-ch were you guys seeing the same thing with measured launch failing?

eric-ch:
@crogers1 No, at least, not on the machines we had: https://openxt.atlassian.net/wiki/spaces/~eric-ch/pages/620789765/Measurement+test. Your changes were in $1610.

I am currently running build 1624 looking at measurement and forward sealing.
There is a strange behavior with the txt module. SELinux AVC that does not make sense (first seen by @jean-edouard) that could explain measure launch failure.

I'll update the page asap.

jandryuk:
I went to re-install with some additional debugging.  When I git F9 for the boot prompt, I see OpenXT safe graphics as the first (hard drive) boot option.  Maybe the boot selection somehow changes between first boot sealing and reboot?

This HP only shows one of the openxt.cfg EFI options in the boot menu.  It was showing regular OpenXT previous.

jean-edouard:
Interesting, I think I hit something similar there: https://openxt.atlassian.net/browse/OXT-1300?focusedCommentId=36796&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-36796

eric-ch:
@jandryuk Are you still reviewing this PR? If so what is the status on it?

jandryuk:
The dmidecode is distinct from the Linux one, so it should be split out.

Reviewing the Linux EFI changes: When arch/x86/xen/efi.c is compiled as 32bit, not settings EFI_64BIT is correct since the 32bit efi_system_table_t is effectively efi_system_table_32_t.  And hence parsing it as 32bit in efi_systab_init is correct.

xenpf_efi_info.cfg.addr is a uint64_t stored into the 32bit ulong efi_systab_xen.tables.  We'll lose any high bits, so the checks in drivers/firmware/efi/efi.c are potentially pointed at the wrong location.

I'm testing out a combination of the Kconfig change to compile XEN_EFI and changing efi_systab_xen to an explicit efi_system_table_64_t.

jandryuk:
The efi_system_table_64_t conversion works for unsealing.  efi though is disabled though `efi: Setup done, disabling due to 32/64-bit mismatch` which means efivars are no longer accessible.  With @eric-ch other PRs they will be inaccessible anyway.

jandryuk:
My alternate patch is https://github.com/jandryuk/xenclient-oe/commit/d119f60849cfed4311edce31a37db43c40a12862 in this branch https://github.com/jandryuk/xenclient-oe/commits/efi-64bit-oxt1307 with @crogers1 's dmidecode patch split out.

jean-edouard:

jean-edouard:
@jandryuk could you please PR your version(s)? Thanks

eric-ch:
https://github.com/OpenXT/xenclient-oe/pull/997 is merged. This PR can be closed.

PR #975, opened 2018-07-20
STABLE-8: qemu-dm: Avoid segfault in dmbus_send

jandryuk:
This is the stable-8 version of https://github.com/OpenXT/xenclient-oe/pull/971

acpi.patch adds this call chain:
acpi_pm1_cnt_write
--xenstore_update_power
----xen_input_send_shutdown
------dmbus_send

When xen_input has not been initialized, dmbus_send segfaults since
input.service is NULL.  If we aren't using xen-input, then we don't need
to notify input_server.  That way we avoid the segfault.

Maybe this should be moved to xen_input_deinit?  In that case, we
wouldn't have the shutdown reason on hand - it would need to be stashed
somewhere.

OXT-1360

(cherry picked from commit 2ed4eeae384615f16e4d97c2fbb287fdfd054d4c)

### DIFF
--- a/recipes-openxt/qemu-dm/qemu-dm-2.6.2/switcher.patch
+++ b/recipes-openxt/qemu-dm/qemu-dm-2.6.2/switcher.patch
@@ -89,7 +89,7 @@ Index: qemu-2.6.2/ui/xen-input.c
 ===================================================================
 --- /dev/null
 +++ qemu-2.6.2/ui/xen-input.c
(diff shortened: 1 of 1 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1625

jean-edouard:
LGTM, merging soon.

PR #976, opened 2018-07-20
OXT-1378: linux: Work-around noisy i2c hid device.

eric-ch:
Seen on Dell E7480.

Every action on the hid device will log a couple of:
```
[ 1538.596775] i2c_hid i2c-DLL07A0:01: i2c_hid_get_input: incomplete report (83/2)
```

Log was already reduced by upstream commit:
```
ef6eaf27274c HID: i2c-hid: Fix "incomplete report" noise
```
Work-around the problem by logging the issue only once.


eric-ch:
See: https://github.com/OpenXT/xenclient-oe/pull/977

PR #981, opened 2018-07-24
OXT-1388: libselinux: Compile with _FILE_OFFSET_BITS=64

jandryuk:
restorecon fails with EOVERFLOW (Value too large for defined data type)
when calling lstat on files >2GB in size.  Setting _FILE_OFFSET_BITS=64
makes compilation use 64bit struct stat64 and stat64 behind the scene
without changing the source to avoid file size issues.

This is needed to support restorecon on /storage/xc-reserved and other
large VM image files.

OXT-1388

### DIFF
--- a/recipes-security/selinux/libselinux/libselinux_2.%.bbappend
+++ b/recipes-security/selinux/libselinux/libselinux_2.%.bbappend
@@ -2,3 +2,6 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/patches:"
 SRC_URI += " \
      file://libselinux-mount-procfs-before-check.patch;patch=1 \
 "
+
+# We need to support stat on files >2GB in size.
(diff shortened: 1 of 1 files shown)


jean-edouard:

jandryuk:
I used TARGET_CFLAGS for this fix, and that works for our VM images.  But maybe this should have been just straight CFLAGS?  That way it would also be in effect for the build host tools.  Thoughts?

jean-edouard:
LGTM, fixes restorecon, merging soon.

jandryuk:
I tested just using `CFLAGS += "-D_FILE_OFFSET_BITS=64"` and it works.  I'd prefer re-pushing with that change instead.

jean-edouard:
Sounds good! Please close this PR then.

PR #982, opened 2018-07-24
OXT-1377: Work around sensitive EFI firmwares

eric-ch:
When using OpenXT UEFI installer, boot only Linux. The current main graph of the UEFI installer does not use Xen.
When booting into OpenXT, pass efi=no-rs to Xen to disable EFI runtime services. Runtime services are currently only used during the installation. Add attr=uc as well, some firmwares flag their memory incorrectly.

This allows OpenXT to install, Boot and setup MLE (including performing forward-seal) on HP 800 G1 and HP 850 G2 with UEFI, as surely other machines with similar firmware.

Other work-around have been attempted before and are documented in the ticket.

### DIFF
--- a/recipes-core/images/xenclient-installer-image/grub.cfg
+++ b/recipes-core/images/xenclient-installer-image/grub.cfg
@@ -2,19 +2,16 @@ set timeout=10
 
 menuentry --hotkey=i 'OpenXT Install' {
     set background_color=black
-    multiboot2 /isolinux/xen.gz placeholder flask=disabled console=com1 com1=115200,8n1,pci mbi-video vga=current loglvl=debug guest_loglvl=debug sync_console
-    module2    /isolinux/vmlinuz placeholder quiet root=/dev/ram rw start_install=new eject_cdrom=1 answerfile=/install/answers/uefi.ans console=/dev/tty2 selinux=0
-    module2    /isolinux/rootfs.gz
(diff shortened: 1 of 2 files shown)


rneilturner:
Great find! We will do some parallel testing with this commit and report the results.

recipes-core/images/xenclient-installer-image/grub.cfg
```diff
@@ -2,9 +2,8 @@ set timeout=10
 
 menuentry --hotkey=i 'OpenXT Install' {
     set background_color=black
-    multiboot2 /isolinux/xen.gz placeholder flask=disabled console=com1 com1=115200,8n1,pci mbi-video vga=current loglvl=debug guest_loglvl=debug sync_console
-    module2    /isolinux/vmlinuz placeholder quiet root=/dev/ram rw start_install=new eject_cdrom=1 answerfile=/install/answers/uefi.ans console=/dev/tty2 selinux=0
-    module2    /isolinux/rootfs.gz
+    linux   /isolinux/vmlinuz placeholder quiet root=/dev/ram rw start_install=new eject_cdrom=1 answerfile=/install/answers/uefi.ans console=/dev/tty2 selinux=0
+    initrd  /isolinux/rootfs.gz
```
rneilturner: @eric-ch how about the other menu options, should they also be given the same treatment?

jean-edouard:
LGTM, merging soon.

PR #984, opened 2018-07-26
OXT-1393: refpolicy-mcs: Allow installer part2 to mount ESP.

eric-ch:
While performing OTA upgrade, part2 installer scripts will fail in
seal-system because of an AVC:
avc:  denied  { mounton } for  pid=3949 comm="mount" path="/tmp/esp" dev="tmpfs" ino=26094 scontext=system_u:system_r:mount_t:s0-s0:c0.c1023 tcontext=system_u:object_r:xc_installer_tmp_t:s0 tclass=dir permissive=0

### DIFF
--- a/recipes-security/refpolicy/refpolicy-mcs-2.%/patches/xc-installer-interfaces.diff
+++ b/recipes-security/refpolicy/refpolicy-mcs-2.%/patches/xc-installer-interfaces.diff
@@ -1,10 +1,19 @@
-Index: refpolicy/policy/modules/system/lvm.te
-===================================================================
---- refpolicy.orig/policy/modules/system/lvm.te
-+++ refpolicy/policy/modules/system/lvm.te
+--- a/policy/modules/system/lvm.te
++++ b/policy/modules/system/lvm.te
 @@ -427,4 +427,5 @@ optional_policy(`
  	updatemgr_dontaudit_rw_fifo_files(lvm_t)
  	updatemgr_dontaudit_rw_stream_sockets(lvm_t)
  	updatemgr_dontaudit_search_storage(lvm_t)
 +	xc_installer_read_tmp_files(lvm_t)
  ')
+--- a/policy/modules/system/mount.te
++++ b/policy/modules/system/mount.te
+@@ -275,3 +275,8 @@ optional_policy(`
+ optional_policy(`
+ 	xc_dontaudit_rw_v4v_chr(mount_t)
+ ')
++
++# Installer part2 needs mounton for ESP partition.
++optional_policy(`
++	xc_installer_mounton_tmp_dirs(mount_t)
++')
--- a/recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/xc-installer.if
+++ b/recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/xc-installer.if
@@ -85,3 +85,21 @@ interface(`xc_installer_read_tmp_files',`
 
 	read_files_pattern($1, xc_installer_tmp_t, xc_installer_tmp_t)
 ')
+########################################
+## <summary>
+##	Allow the calling domain to mount filesystem on all tmpfs directories.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`xc_installer_mounton_tmp_dirs',`
+	gen_require(`
+		type xc_installer_tmp_t;
+	')
+
+	allow $1 xc_installer_tmp_t:dir mounton;
+')
+


recipes-security/refpolicy/refpolicy-mcs-2.%/patches/xc-installer-interfaces.diff
```diff
@@ -1,10 +1,19 @@
...
-===================================================================
---- refpolicy.orig/policy/modules/system/lvm.te
-+++ refpolicy/policy/modules/system/lvm.te
+--- a/policy/modules/system/lvm.te
++++ b/policy/modules/system/lvm.te
 @@ -427,4 +427,5 @@ optional_policy(`
  	updatemgr_dontaudit_rw_fifo_files(lvm_t)
  	updatemgr_dontaudit_rw_stream_sockets(lvm_t)
  	updatemgr_dontaudit_search_storage(lvm_t)
 +	xc_installer_read_tmp_files(lvm_t)
```
jandryuk: Why this change?  Only the `mounton` one is directly explained by the commit message.  Does an LVM tool poke at the partition after it is mounted?

eric-ch: This is context change only due to `quilt refresh`. I am using the same quiltrc (see [wiki](https://openxt.atlassian.net/wiki/spaces/DC/pages/20905986/Dealing+with+quilt) for the OpenXT layer, so Index and the separation line are gone, and `-ab` is used.
`xc_installer_read_tmp_files(lvm_t)` is already applied earlier.

jandryuk: Oh, I misread the diff of a diff again.  Sorry for the noise.

jandryuk:
An alternative is to re-work seal-system.  Currently, seal-system is called from the installer [install-main:seal_system()](https://github.com/OpenXT/installer/blob/master/part2/stages/Functions/install-main#L431) only on upgrade.  It is executed as `chroot ${DOM0_MOUNT} /usr/sbin/seal-system -f -r ${ROOT_DEV}` 
[](url)

seal-system is mounting ESP to grab all the files needed for forward sealing: [seal-system](https://github.com/OpenXT/xenclient-oe/blob/master/recipes-openxt/openxt/openxt-measuredlaunch/seal-system#L152).  These were copied from the new dom0 image on `$DOM0_MOUNT` into ESP in the first place [copy_to_esp](https://github.com/OpenXT/installer/blob/master/part2/stages/Functions/install-main#L606).  So seal-system could be changed to use the original source files instead of requiring a mount of ESP.

Are there still forward seal issues?  Not sure if they should be, but BOOT.CSV and microcode_intel.bin are not included in seal-system's calculation.

eric-ch:
@jandryuk Sure, measuring the files in the root partition should work too.
We are still having forward-sealing failures with UEF/TPM1.2 platforms.

TPM2.0 platforms manage to boot just fine without measuring `BOOT.CSV` and `microcode_intel.bin` so I would think these are not measured in pcr4/pcr8.

jean-edouard:
LGTM, merging soon.

dpsmith:
I understand why you want to measure microcode, but why measure sinit when finer grained measurements will be stored in 17/18 by senter?

tklengyel:
I understand that the sinit module should get measured into PCR17 but how exactly that is accomplished is a bit murky. If the code in the init module measures itself - and AFAICT that's what happens - then getting that code also getting measured as part of SRTM would be warranted. Plus it's also pretty easy to loop it in, so even if it gets measured 2x, that's not really a bad thing by any means.

jandryuk:
I would measure anything loaded, even if it *should* be measured later on.

dpsmith:
It is not murky, it is measured by the CPU and not self referential integrity like UEFI. Jason and I discussed and in line with his comment that measuring it when it is load provides consistence and no security at the cost of performance when the system uses a discrete TPM.

tklengyel:
@dpsmith thanks, I was not sure about how exactly it is measured. Can you give us the reference where that is described in the manual? I looked but other then that it says it is measured I didn't find more detail about how that measurement is made exactly.

EugeneDMyers:
Ref:  Intel TXT Software Development Guide (SDG)
        TCG PC Client Specific TPM Interface Specification (TI)

A very brief description of what happens during one part of the Intel D-RTM process (SENTER) as it pertains to SINIT AC measurement

(1) The TPM is set into Locality 4.  (Note: Localities 3 and 4 can only be set by the processor microcode.)  During this phase PCR's 17-20 are reset and the measurement of the SINIT AC module is placed into PCR 17.  Section 1.10.2.1 of the SDG describes what is placed there.  (note: this measurement is done by the processor microcode and is considered as the root of D-RTM, ref TI)

(2) Once step one is completed, then the TPM is set into Locality 3.  At this time, the SINIT AC module is then executed where is does the measurement of the MLE, etc.

(3) At the completion of step two, the TPM is set into locality 2 and the MLE is executed.

tklengyel:
@EugeneDMyers Thanks!

tklengyel:
@EugeneDMyers looking through the documentation and it is pretty hard to parse.

> PCR 17 is initialized using the _TPM_HASH_* sequence.  The _HASH_DATA provided 
in this sequence is the concatenation of digest of the SINIT ACM that was used in the launch process and the 4 byte value of the SENTER parameters (in the EDX register 
and also in SinitMleData.EdxSenterFlags).  As part of this sequence, PCRs 17-23 are 
reset to 0 before _TPM_HASH_START.  The digest of SINIT is also stored in the 
SinitMleData.SinitHash field of the SINIT to MLE data table (SinitMleData see Appendix C.5).

This is what I found before in the SDG but it doesn't specify what does the measurement of SINIT. However, as you point out, PCR17 is resettable from Locality 4 only, which implies the microcode:

> Locality 4 (Hardware or microcode access only)

It is still not perfectly clear whether it is actually done by baked-in non-mutable hardware or as you say by microcode (or if its perhaps implementation specific). Microcode is problematic IMHO as it is mutable firmware. I would like to point out a relevant passage in the draft of the PCIe security enhancement document (https://www.intel.com/content/dam/www/public/us/en/documents/reference-guides/pcie-device-security-enhancements.pdf):

> 3.2.1 DIGEST Register
Each Device shall implement one or more DIGEST register(s) that expose the cryptographic identities of all of its firmware components. Every mutable firmware component that the Device executes must  be  measured  in  at  least  one  DIGEST  register. In  this  context, cryptographic identity  implies that 
a  secure  hash  function is applied  to  the  firmware  images. The Device manufacturers should select the algorithm based on the prevailing  recommendations, such as those from NIST (NIST's Policy on Hash Functions). The DIGEST register itself is considered part of the Root-of-Trust for Measurement (RTM) and therefore the underlying implementation of this register cannot have any dependency on 
mutable firmware. Additionally this register must be implemented such that the mutable portion of the 
firmware cannot modify the register value or intercept Host accesses to this register.

EugeneDMyers:
For a good read on how the processor authenticates SINIT, read pages 197-203 of The Intel Safer Computing Initiative by David Grawrock.  Though not an Intel Manual, this book provides the best (and readable) blow by blow description on how the Intel D-RTM process works W/O an NDA.  David was one of the principals in the TXT design.

The Author does not state if this process is in microcode or hardware and I suspect that how it is implemented is dependent on the processor family.

As for immutability, Intel has a very secure microcode update process and I would be more concerned about how the platform vendors define immutability for their BIOS's, as that affects the SRTM.

I like the idea of signing the microcode, that would be a question that Intel would have to answer.

tklengyel:
@EugeneDMyers thanks! Will give that a read!

PR #985, opened 2018-07-26
STABLE-8: OXT-1393: refpolicy-mcs: Allow installer part2 to mount ESP.

eric-ch:
While performing OTA upgrade, part2 installer scripts will fail in
seal-system because of an AVC:
avc:  denied  { mounton } for  pid=3949 comm="mount" path="/tmp/esp" dev="tmpfs" ino=26094 scontext=system_u:system_r:mount_t:s0-s0:c0.c1023 tcontext=system_u:object_r:xc_installer_tmp_t:s0 tclass=dir permissive=0

(cherry picked from commit 4f29bf573e588e42e95f03fd9033c1306f730329)


jean-edouard:
LGTM, merging soon.

PR #986, opened 2018-07-27
OXT-1396: seal-system: change TPM1.2 EV_SEPARATOR to the sha1 of 0xFFFFFFFF

jean-edouard:
(instead of 0x00000000)

### DIFF
--- a/recipes-openxt/openxt/openxt-measuredlaunch/seal-system
+++ b/recipes-openxt/openxt/openxt-measuredlaunch/seal-system
@@ -144,7 +144,9 @@ forward)
 
         # PCR4 is first extended with the digest of EV_SEPARATOR
         # See TCG EFI Protocol Specification 5.2 Crypto Agile Log Entry Format
-        ev_separator="9069ca78e7450a285173431b3e52c5c25299e473"
+        # The specs says that ev_separator is the hash of 0x00000000
+        # However, TPM1.2 implementations use the sha1 of 0xffffffff...
+        ev_separator="d9be6524a5f5047db5866813acf3277892a7a30a"
         [ "${tpm2}" -eq 0 ] && ev_separator="df3f619804a92fdb4057192dc43dd748ea778adc52bc498ce80524c014b81119"
 
         pcr4=$(hash_extend 0 $ev_separator $hashalg)


dpsmith:
Actually below is the exact wording from the spec for the value of EV_SEPARATOR,

> Per Section 2.3.2, used to indicate an error occurred recording
> the CRTM, POST BIOS, or Embedded Option ROMs. For this
> situation, the digest field MUST contain the tagged digest of the
> value 00000001h. The event field is arbitrary to allow platform
> manufacturers to include an indication of what error occurred or
> other useful troubleshooting information.
> Per Sections 2.3.4 and 7.2, used to delimit actions taken during
> the pre-OS and OS environments. For this situation, the digests
> field MUST contain the tagged hash of the event data for each
> PCR bank. The event data size MUST be 4. The event field
> MUST contain the hex value 00000000h or FFFFFFFFh.
 
Therefore when it is used as a delimiter, it can be either 0x00000000 **or** 0xFFFFFFFF, depends on what the firmware implementer decided to do.

eric-ch:
@dpsmith I still find that unexpected. AFAICT, you are quoting [TCG PC Client Platform Firmware Profile Specification](https://trustedcomputinggroup.org/wp-content/uploads/PC-ClientSpecific_Platform_Profile_for_TPM_2p0_Systems_v51.pdf) which applies to family 2.0.

This fixes 1.2 TPM family only.

... [some googling research further]
```
1.1 Interoperability
Although this specification is for TPM 2.0 devices, it contains nothing that actively
prevents the use of the specified protocol with TPM 1.2 devices.
```

So we should calculate against extending both 00000000h or FFFFFFFFh?

eric-ch:
In case this slipped under the radar: 
- Build [$1635](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1635) *has* that change and *succeed consistently* forward sealing on TPM 1.2 plaftorms (HPs and Dells alike).
- Build [$1637](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1637) *does not have* that change and *fails consistently* forward sealing on TPM 1.2 plaftorms (HPs and Dells alike).

dpsmith:
This is why its a problem, the spec a based on whether you are coming from a "Conventional BIOS" or from "Platform Firmware", e.g. UEFI. The "TCG PC Client Specific Implementation Specification for Conventional BIOS; For TPM Family 1.2; Level 2" states,

> Used for PCRs[0, 1, 2, 3, 4, 5, 6 and 7].
> Per Section 3.3.2.2, used to indicate an error
> occurred recording the CRTM, POST BIOS, or
> Embedded Option ROMs. For this situation, the
> digest MUST contain the value 01h. The event field is
> arbitrary to allow platform manufacturers to include
> an indication of what error occurred or other useful
> troubleshooting information.
> Per Section 8.2.3, used to delimit actions taken
> during the pre-OS and OS environments. For this
> situation, the digest value contains the SHA-1 hash of
> the event data. The event data size MUST be 4. The
> event field MUST contain the hex value
> 0xFFFFFFFF.

Thus if you are starting from CSM, then it should only use 0xFFFFFFFF. If you start up in EFI, then it can go either way. You can thank Microsoft for this one.

eric-ch:
> You can thank Microsoft for this one.

[Thanks Microsoft](https://sd.keepcalm-o-matic.co.uk/i-w600/keep-calm-we-re-all-mad-here.jpg)

eric-ch:
Ticket: [OXT-1396](https://openxt.atlassian.net/browse/OXT-1396)

dpsmith:
Unless someone objects, IMHO this should go in as is and document in the new ticket it is possible for TPM 1.2 to use 0x00000000 in EFI. That way if it does break on someone's platform, it will be easier to determine why.

eric-ch:
@jandryuk Using a Dell E7450 (TPM 1.2) in UEFI I have the 0xffffffff value:
```
root@xenclient-dom0:~# cat /sys/kernel/security/tpm0/ascii_bios_measurements | awk '$1 == 4 && $3 == 04 { print $2 }'
d9be6524a5f5047db5866813acf3277892a7a30a
```

dpsmith:
Just to clarify, should != does. The spec also says that after the separator event occurs the pre-boot environment measurements should be complete but the Dell that @jandryuk  is testing with has a PCR1 event post separator event.

tklengyel:
On my Toshiba with UEFI:
```
# cat /sys/kernel/security/tpm0/ascii_bios_measurements | awk '$1 == 4 && $3 == 04 { print $2 }'
9069ca78e7450a285173431b3e52c5c25299e473

tklengyel:
I think the suggestion by @jandryuk is something to consider, read the value at runtime from the logs instead of hard-coding it.

richpersaud:
If further work is planned, please create a 9.0 ticket to collect potential improvements.

eric-ch:
@richpersaud See: https://github.com/OpenXT/xenclient-oe/pull/992

PR #988, opened 2018-07-27
OXT-1395: tpm2-tools: reset hash_set variable during parsing

eric-ch:
Do not leak previous values on N+1 parse operation.


jean-edouard:
:grinning:

jean-edouard:
LGTM, merging soon.

PR #989, opened 2018-07-27
STABLE-8: OXT-1395: tpm2-tools: reset hash_set variable during parsing

eric-ch:
Do not leak previous values on N+1 parse operation.

(cherry picked from commit 6f1a94167540c7b9e10362c5fd62ac67df22b0ce)


crogers1:
Wow great catch.  I'm upset I missed that detail when I originally submitted this patchset.  Thanks

eric-ch:
Set `hash_set` in the function since it is also called for `tpm2_unsealdata`. The value of `hash_set` should not be dependent on its input anyway.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1644

jean-edouard:
Merging soon.

PR #990, opened 2018-07-27
installer: Add ttyS0 console back to the cmdline

eric-ch:
The console used to be set on the Xen cmdline. Pass it to Linux.


jean-edouard:
LGTM, merging soon.

PR #992, opened 2018-07-27
OXT-1396: openxt-ml: read ev_separator in sysfs, if possible

eric-ch:
@jean-edouard's fix including @jandryuk's recommendations.

EV_SEPARATOR can vary between firmwares with TPM1.2 families.
This value can be read from the sysfs in all observed cases, still keep
a fallback just in case.

### Alternative to:
https://github.com/OpenXT/xenclient-oe/pull/986

### DIFF
--- a/recipes-openxt/openxt/openxt-measuredlaunch/seal-system
+++ b/recipes-openxt/openxt/openxt-measuredlaunch/seal-system
@@ -144,10 +144,19 @@ forward)
 
         # PCR4 is first extended with the digest of EV_SEPARATOR
         # See TCG EFI Protocol Specification 5.2 Crypto Agile Log Entry Format
-        ev_separator="9069ca78e7450a285173431b3e52c5c25299e473"
-        [ "${tpm2}" -eq 0 ] && ev_separator="df3f619804a92fdb4057192dc43dd748ea778adc52bc498ce80524c014b81119"
+        ev_separator=""
+        if [ "${tpm2}" -ne 0 ]; then
+            # Read the currently used separator from tpm0 sysfs.
+            if [ -e "/sys/kernel/security/tpm0/ascii_bios_measurements" ]; then
+                ev_separator="$(awk '$1 == 4 && $3 == 04 { print $2 }' /sys/kernel/security/tpm0/ascii_bios_measurements)"
+            fi
+            # Pick `printf "\xff\xff\xff\xff" | sha1sum` as it looks popular.
+            ev_separator=${ev_separator:-"d9be6524a5f5047db5866813acf3277892a7a30a"}
+        else
+            ev_separator="df3f619804a92fdb4057192dc43dd748ea778adc52bc498ce80524c014b81119"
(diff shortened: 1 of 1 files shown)


recipes-openxt/openxt/openxt-measuredlaunch/seal-system
```diff
@@ -144,8 +144,17 @@ forward)
 
         # PCR4 is first extended with the digest of EV_SEPARATOR
         # See TCG EFI Protocol Specification 5.2 Crypto Agile Log Entry Format
-        ev_separator="9069ca78e7450a285173431b3e52c5c25299e473"
-        [ "${tpm2}" -eq 0 ] && ev_separator="df3f619804a92fdb4057192dc43dd748ea778adc52bc498ce80524c014b81119"
+        if [ "${tpm2}" -ne 0 ]; then
+            # Read the currently used separator from tpm0 sysfs.
+            if [ -e "/sys/kernel/security/tpm0/ascii_bios_measurements" ]; then
+                ev_separator="$(awk '$1 == 4 && $3 == 04 { print $2 }' /sys/kernel/security/tpm0/ascii_bios_measurements)"
+            else
```
jandryuk: If awk fails to find a string, `ev_separator` will be empty.  Do we want a fall back for that case, or would things be so broken it's not worth bothering?

recipes-openxt/openxt/openxt-measuredlaunch/seal-system
```diff
@@ -144,8 +144,17 @@ forward)
...
+                # Pick sha1sum $(printf "\xff\xff\xff\xff" as it looks popular.
+                ev_separator="d9be6524a5f5047db5866813acf3277892a7a30a"
+            fi
+        else
+            ev_separator="df3f619804a92fdb4057192dc43dd748ea778adc52bc498ce80524c014b81119"
+        fi
 
         pcr4=$(hash_extend 0 $ev_separator $hashalg)
```
jandryuk: `ev_separator` is unquoted, so this will become `hash_extend 0 [sha1|sha256]` with hash alg defaulting to sha1.  The literal strings sha1 or sha256 will fail the length check and hash_extend will return 1.  `pcr4` then becomes an empty string.

Maybe if ev_separator is unset we should be skipping this extension?  But again, maybe things are too broken to bother with that case.

recipes-openxt/openxt/openxt-measuredlaunch/seal-system
eric-ch: I do not see any drawback with a fallback if `awk` fails to find the string.

eric-ch:
Amend commit with Jason's recommendation:
- Set `ev_separator` from `/sys/kernel/security/tpm0/ascii_bios_measurements` if the file is present
- Should `ev_separator` be empty, fallback on the default value.
- Add quoting.

jean-edouard:
LGTM, merging soon.

PR #994, opened 2018-07-30
installer: add physical serial support to the kernel

jean-edouard:


eric-ch:
Restore serial output from the installer.
Will merge soon.

PR #996, opened 2018-08-03
OXT-1403: Signed kernel modules

jandryuk:
Configure linux kernels to only load signed modules.  The default is to create and sign with an auto-generated keypair.  It is also possible to specify a keypair.

For external keys, two variables can be set in local.conf: KERNEL_MODULE_SIG_CERT and KERNEL_MODULE_SIG_KEY.  KERNEL_MODULE_SIG_CERT can point to the absolute filepath of a public key for signed module verification.  KERNEL_MODULE_SIG_KEY can point to the absolute filepath of a private key for signing kernel modules.

If KERNEL_MODULE_SIG_CERT is defined, it will be embedded in the generated linux kernel.  If KERNEL_MODULE_SIG_KEY is defined, it will be used to sign the modules during the build.  If undefined, modules must be signed manually.  Images would need to be extracted, all the .ko files signed in /lib/modules, and then the image re-packed.  The general signing commands are found in classes/module-signing.bbclass

Keypairs can be generated according to Documentation/admin-guide/module-signing.rst in the linux source tree along with more information about signed kernel modules.

OXT-1403


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1654

jean-edouard:
LGTM, merging soon.

recipes-kernel/linux/4.14/defconfigs/xenclient-dom0/defconfig
```diff
@@ -304,7 +304,15 @@ CONFIG_MODULE_UNLOAD=y
 CONFIG_MODULE_FORCE_UNLOAD=y
 CONFIG_MODVERSIONS=y
 CONFIG_MODULE_SRCVERSION_ALL=y
-# CONFIG_MODULE_SIG is not set
+CONFIG_MODULE_SIG=y
+CONFIG_MODULE_SIG_FORCE=y
```
eric-ch: It is my understanding that enabling this will prevent loading an unsigned module.
Yet, with build [$1653](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1653), unsigned modules are loaded in dom0  (e.g, `fbtap`, its recipe does not inherit `kernel-signing.bbclass`).

I must be missing something or the signing is not performed while `CONFIG_SIG_FORCE` is set.

jandryuk:
Sight modules are enabled. Without doing anything, a build-time keypair will be generated and used for signing. So you don't have to do anything by default. Support is also there for specifying, and signing with, a keypair.

If you do `strings /lib/modules/.../...ko | tail -n1` you'll see "~Module*~" showing the module is signed.

eric-ch:
@jandryuk ... I just realized that I used build [$1653](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1653) and not [$1654](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1654)...

```bash
root@xenclient-dom0:~# strings $(find /lib/modules/4.14.57/ -name '*.ko') | tail -n1
~Module signature appended~
```
... match the [doc](https://www.kernel.org/doc/html/v4.17/admin-guide/module-signing.html) as you mentioned.

Using psmouse.ko from [$1653](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1653) on [$1654](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1654):
```bash
root@xenclient-dom0:~# insmod /storage/psmouse.ko 
insmod: ERROR: could not insert module /storage/psmouse.ko: Required key not available
```
... confirms things are working as expected.

jean-edouard:
\o/ thanks, merging very soon now :)

PR #997, opened 2018-08-06
OXT-1307: Convert XEN_EFI support to 64bit

jandryuk:
Allow enabling EFI support for 32bit Xen dom0.  This is needed to
populate sysfs DMI and ACPI entries needed by dmidecode, xcpmd, and
get-config-key.

This is an alternative to https://github.com/OpenXT/xenclient-oe/pull/973

CONFIG_XEN_EFI depends on CONFIG_X86_64, but this PR changes Kconfig
to allow X86 to compile for 32bit Dom0.

The Xen EFI data is 64bit, so we need to switch efi_systab_xen to an explicit
64bit efi_system_table_64_t for the 32bit Dom0.  The data returned from the
hypercall is 32-bit addressable, the only unknown is whether the EFI tables
themselves are 32-bit addressable. The EFI initialization code protects against
this case by checking to see if each table is below 4GB in memory, and if not,
it disables EFI.

With the 64bit conversion, the casts make the runtime and boottime pointers
into 64bit ~0 values.  This trips the "EFI data located above 4GB, disabling
EFI." error in efi_systab_init.  Dropping the cast leaves them as 32bit which
passes the check.

While setup succeeds, EFI is disabled with:
"efi: Setup done, disabling due to 32/64-bit mismatch"

This is enough to populate the sysfs DMI and ACPI entries, which dmidecode,
xcpmd, and get-config-key rely upon.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1654

jean-edouard:
@crogers1, could you please review?

crogers1:
LGTM, sorry for the delay.  Much cleaner than my first attempt.

eric-ch:
@crogers1 Thanks!

Will merge soon.

jandryuk:
Thanks, @crogers1 .  You did the hard part of actually solving the problem. :)

PR #998, opened 2018-08-06
OXT-1388: libselinux: Compile with _FILE_OFFSET_BITS=64

jandryuk:
restorecon fails with EOVERFLOW (Value too large for defined data type)
when calling lstat on files >2GB in size.  Setting _FILE_OFFSET_BITS=64
makes compilation use 64bit struct stat64 and stat64 behind the scene
without changing the source to avoid file size issues.

This is needed to support restorecon on /storage/xc-reserved and other
large VM image files.

OXT-1388

This version changes CFLAGS as opposed to TARGET_CFLAGS in https://github.com/OpenXT/xenclient-oe/pull/981


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1655

jean-edouard:
Merging soon.

PR #1001, opened 2018-08-08
OXT-1403: Signed modules fix

jandryuk:
This fixes module signing on a clean build when
${STAGING_KERNEL_BUILDDIR}/scripts/sign-file isn't populated.

On a clean build, do_sign_modules fails when using a specified
$KERNEL_MODULE_SIG_KEY when it cannot find sign-file.  do_sign_modules
runs before sign-file has been installed into $STAGING_KERNEL_BUILDDIR.
module-base.bbclass provides do_make_scripts which is responsible for
the installation, but that is only run when an out-of-tree module is
built.

module-base cannot be inherited into the kernel build because it creates
a dependency loop.

The kernel build dir ${B} already has sign-file, so just check for it
there in addition to the $STAGING_KERNEL_BUILDDIR.


eric-ch:
Will merge soon.

PR #1002, opened 2018-08-08
linux: Remove Remote Control & Infrared drivers

jandryuk:
A change to 'default y' meant that upgrading some old linux configs
pulled these in.  We don't need them, so remove them.


eric-ch:
Will merge soon.

PR #1003, opened 2018-08-08
xenclient-installer: Make package arch-independent

jandryuk:
xenclient-installer is arch-independent since it is a bunch of shell
scripts.  Mark it as such by inheriting allarch.


eric-ch:
Will merge soon.

PR #1004, opened 2018-08-09
pre-install: switch to bash and allow sshd to be already stopped

jean-edouard:

### DIFF
--- a/recipes-core/images/xenclient-installer-image/network.ans
+++ b/recipes-core/images/xenclient-installer-image/network.ans
@@ -1,5 +1,5 @@
 <interactive>false</interactive>
-<preinstall>#!/bin/ash -e
+<preinstall>#!/bin/bash
 touch /tmp/preinstall.touch
 mkdir -p /config/etc/ssh
 touch /config/etc/ssh/enabled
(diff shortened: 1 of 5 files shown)


jandryuk:
Why switch to bash?

Are you dropping "-e" to avoid error-ing out when sshd isn't running?

jean-edouard:
Since we now have bash everywhere, we might as well switch to it...

Yes, no need to fail the whole install if any of these few lines fail. The most likely one being sshd stop as you guessed :)

jandryuk:
It would be good to include a little more explanation in the commit message.  /etc/init.d/sshd stop doesn't appear in the diff context.  Another option would be to switch to `/etc/init.d/sshd stop || :` to only bypass the error-exit for that one command.  For these preinstall commands, removing "-e" is fine though.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1669

richpersaud:
Please create a ticket for the proposed change, with details on the motivation, benefits and risks of the change.

jean-edouard:

jean-edouard:

eric-ch:
Commit message is amended, is there any other concern about this PR?

richpersaud:
Is there a JIRA ticket for the change proposed in this PR?

PR #1006, opened 2018-08-13
OXT-1410: seabios: Change source URL

eric-ch:
Change premirror, tarballs are now at:
https://www.seabios.org/downloads/


eric-ch:
Building [$1669](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1669)

jean-edouard:
LGTM, merging soon.

PR #890, opened 2018-05-03
OXT-1318: Launch tboot from Xen.efi

tklengyel:
Add Xen patches to enable launching tboot from xen.efi

Signed-off-by: Tamas K Lengyel <lengyelt@ainfosec.com>

PR #1011, opened 2018-08-15
OXT-1397: openxt-ml: Do not mount ESP in seal-system.

eric-ch:
From initial discussion on https://github.com/OpenXT/xenclient-oe/pull/984.

seal-system is called from the installer install-main:seal_system() only
on upgrade. It is executed as chroot ${DOM0_MOUNT} /usr/sbin/seal-system
-f -r ${ROOT_DEV}. seal-system is mounting ESP to grab all the files
needed for forward sealing. These were copied from the new dom0 image on
$DOM0_MOUNT into ESP in the first place copy_to_esp.

Knowning that, assume default path for the measured files and provide
two new options to seal-system:
```
-4  [object[,objects]]  List, comma ',' separated, of objects measured
                        in pcr4 for UEFI installations. The hash of the
                        first object in the list is calculated by
                        pesign, according to the specification.
-8  [object[,objects]]  List, comma ',' separated, of objects measured
                        in pcr8 for UEFI installations. Order is
                        significant.
```

### DIFF
--- a/recipes-openxt/openxt/openxt-measuredlaunch/seal-system
+++ b/recipes-openxt/openxt/openxt-measuredlaunch/seal-system
@@ -31,6 +31,29 @@ contains_only() {
     [[ -z "${s//$c}" ]]
 }
 
+usage() {
+    cat <<EOF >&2
+Usage: seal-system [-h] [-s|-u|-f] [-ctpgbr path] -[48 object[,objects]]"
+  -h      Display this usage.
+  -s      Seal operation.
+  -u      Unseal operation.
+  -f      Forward sealing operation.
+  -c path Path to configuration key.
+  -t path Path to sealed key.
+  -p path Path to pcr list file.
+  -g path Path to "good" pcrs log file.
+  -b path Path to "bad" pcrs log file.
+  -r path Path to root block device to be measured.
+
+  -4 object[,objects] List, comma ',' separated, of objects measured in pcr4
+                      for UEFI installations. The hash of the first object
+                      in the list is calculated by pesign, while the following
+                      will extend as raw-buffers the initial Authenticode hash.
+  -8 object[,objects] List, comma ',' separated, of objects measured in pcr8
+                      for UEFI installations. Order is significant.
+EOF
+}
+
 # Globals
 TPM_DEV="$(find /sys/class -name tpm0)/device"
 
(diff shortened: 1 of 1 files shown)


jandryuk:
Thanks for doing this.

https://github.com/OpenXT/xenclient-oe/commit/0c92e3b4dd2a2eae47ad3217cb62058dc1cab84a from https://github.com/OpenXT/xenclient-oe/pull/890 Moves a umount, so there will be a conflict. Other fixups will probably be needed.

recipes-openxt/openxt/openxt-measuredlaunch/seal-system
```diff
@@ -47,8 +47,20 @@ hashalg="sha1"
 good_pcrs="/config/good.pcrs"
 bad_pcrs="/boot/system/tpm/bad.pcrs"
 root_dev="/dev/mapper/xenclient-root"
-
-while getopts ":sufc:t:p:g:b:r:" opt; do
+# First file is parsed and signed with pesign, following object with extend the
+# hash.
```
tklengyel: This comment perhaps needs a bit more explanation from my end. According to the UEFI specification PCR4 should contain measurement of UEFI executables using the Authenticode hash, which is what pesign produces. Since the shim gets loaded by the firmware directly it gets measured as Authenticode. Due to a firmware bug on Dell systems the other executables (xen/dom0 kernel) can't be measured with Authenticode, so they are measured as raw buffers. Ideally once Dell fixes their issue (they have been notified) and all firmwares uniformly can measure with Authenticode we would make this UEFI spec compliant. But it's arguably a low priority.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1677

jean-edouard:
Smoke test failed, got the following error on OTA upgrade:
```
Aug 16 17:50:54.467354 updatemgr: pesign: Error opening input: No such file or directory
Aug 16 17:50:54.481341 updatemgr: seal-system: failed to calculate pcr4
```

eric-ch:
@jean-edouard Thanks, looking into it.

eric-ch:
@jean-edouard ahem... my bad...

##### v2:
- Fixed dumb typo...
- Added Tamas comment inline and in the commit message to hopefully prevent futur loss of sanity,
- Added a (too simple I am sure) usage for short options. since this is getting hairy.

jean-edouard:

jean-edouard:
LGTM, merging soon.

PR #1013, opened 2018-08-17
OXT-1323, OXT-1389: tboot: Upgrade to 1.9.6 and backports.

eric-ch:
Patchqueue forward ports:
- TB_POLTYPE_WARN_ON_FAILURE
* Upstream introduced TB_ERR_PREV_TXT_ERROR, which seems equivalent to
our TB_ERR_PREV_LAUNCH_FAILURE. The new replace the old in the ported
patch.
* Split 0012-tboot-Propagate-failure-to-map_tboot_pages.patch from
former warn-on-failure-policy.patch as it does not seem to relate
directly. More disagregation could be done later, documentation is left
in the patch for futur work.

Backports (patches 0001-0009):
- 0006-Fix-security-vulnerabilities-rooted-in-tpm_if-struct.patch
  This is OXT-1398.

Patches upstream:
- increase-cmdline-in-linux-real-mode-boot-header-to-1024.patch
- linux-kernel-cmdline-buffer-overflow.patch


jean-edouard:

jean-edouard:
LGTM, merging tomorrow.

PR #1017, opened 2018-08-17
xfwm: start as daemon instead of background

jean-edouard:
Not sure why that changes anything, but this is what xfce does by default,
and it ensures terminals always spawn in the foreground.

### DIFF
--- a/recipes-openxt/xenclient/xenclient-uivm-xsessionconfig/xinitrc
+++ b/recipes-openxt/xenclient/xenclient-uivm-xsessionconfig/xinitrc
@@ -111,7 +111,7 @@ if test x"$DBUS_SESSION_BUS_ADDRESS" = x""; then
     fi
 fi
 
-xfwm4 &
+xfwm4 --daemon
 resized &
 xfsettingsd --no-daemon --replace &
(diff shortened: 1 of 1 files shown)


jean-edouard:
I think this should go into stable-8 as well. I'll PR there if nobody disagrees.

eric-ch:
Terminal no longer open in background (...why?).

Ending the UI (Ctrl+Shift+q) will terminate `X`, which will close the socket read by `xfwm4` and end the process, so restarting the UI is not a problem either with this change.

jean-edouard:

jandryuk:
The terminal in the background was annoying.  Thanks for fixing.

Out of curiousity, I looked up xfwm4 daemon mode, and it was removed in upstream https://github.com/xfce-mirror/xfwm4/commit/43e76c7bdfa6db6bf95f576723dc5e41cc15f59e  So this may come back in the future, but this is fine for now.

eric-ch:
@jandryuk quoting the commit "This is causing all sort of weird bugs and issues." :D. Hopefully we will not have to backport it. Thanks for tracking the change!

eric-ch:
Will merge soon.

PR #1018, opened 2018-08-17
OXT-1410: seabios: Remove PREMIRRORS statement.

eric-ch:
First, PREMIRRORS is a little too strong to be defined at the recipe
level like so. I imagine it is more intended at providing a local.conf
override to fetch sources from a local cache. Having PREMIRRORS_prepend
makes that much more difficult.

Also, in this case, the upstream recipe has been fixed (at least was
yesterday).

This source is quite volatile. It should be recommended to fetch from a
stable mirror.

Maybe put the tarball on mirror.openxt.org and add a {PRE,}MIRRORS
statement in the default configuration file.


jean-edouard:

jean-edouard:
LGTM, merging soon.

PR #1026, opened 2018-08-22
OXT-1408: dmidecode: bbappend upstream recipe.

eric-ch:
Use the upstream recipe and apply the use-oe-var.patch modification to
inherit the build environment.


recipes-devtools/dmidecode/dmidecode_3.%.bbappend
```diff
@@ -0,0 +1,5 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/patches:"
+
+SRC_URI += " \
+    file://use-oe-vars.patch;patch=1 \
```
jandryuk: When a file ends in .patch, Bitbake/OE treats it as a patch, so patch=1 isn't necessary.

eric-ch: Yes, I took the original line, will amend.
(Btw, we should probably remove that everywhere when we can).

jean-edouard:

PR #1028, opened 2018-08-22
OXT-1365: refpolicy-mcs: give updatemgr overcommit read perm

eric-ch:
Use the correct interface to allow read permission on
/proc/sys/vm/overcommit_memory.

While kernel_read_vm_sysctls() will give appropriate permission to find
the overcommit_memory procfs node, that last one is labelled
sysctl_vm_overcommit_t.

kernel.if defines kernel_read_vm_overcommit_sysctl() macro which allow
both kernel_search_vm_syctls() and sysctl_vm_overcommit_t
read_file_perms required by updatemgr.


eric-ch:
@jandryuk Not directly, I would think it ends up being the glibc given the right memory allocation options passed to, say `malloc(3)` or `mmap(3)`? Could be the Haskell 6.x memory management trying to allocate memory in a bulk?

jandryuk:
glibc contains the string.

jean-edouard:

jean-edouard:
LGTM, merging soon.

PR #1030, opened 2018-08-23
module-signing: Fix SIGN_FILE error detection

jandryuk:
Use a lockfile to avoid racing between make_scripts and sign_modules tasks.

Also fix the error condition for the sign-file detection.


jean-edouard:

eric-ch:
Will merge soon.

PR #1037, opened 2018-09-06
OXT-1431: Guest UEFI support

tklengyel:
Build and install OVMF to support UEFI mode in guests. The OVMF version produced from the upstream openembedded recipe doesn't work with Xen. The version of OVMF tagged by Xen doesn't support recent versions of Windows 10. Because of this we pick a recent revision of OVMF from upstream that is known to work (same revision is picked by Debian buster).

Guest UEFI mode has been tested with Debian Stretch and various version of Windows 10. Booting Windows 10 installer with OVMF may take several minutes during which only the "TianoCore" splash screen is displayed but afterwards Windows installs and boots without any issues.

### DIFF
--- a/recipes-core/ovmf/ovmf/0003-BaseTools-makefile-adjust-to-build-in-under-bitbake.patch
+++ b/recipes-core/ovmf/ovmf/0003-BaseTools-makefile-adjust-to-build-in-under-bitbake.patch
@@ -1,0 +1,25 @@
+diff --git a/BaseTools/Source/C/Makefiles/header.makefile b/BaseTools/Source/C/Makefiles/header.makefile
+index db436773cf..79e554b790 100644
+--- a/BaseTools/Source/C/Makefiles/header.makefile
++++ b/BaseTools/Source/C/Makefiles/header.makefile
+@@ -68,15 +68,15 @@ $(error Bad HOST_ARCH)
+ endif
+ 
+ INCLUDE = $(TOOL_INCLUDE) -I $(MAKEROOT) -I $(MAKEROOT)/Include/Common -I $(MAKEROOT)/Include/ -I $(MAKEROOT)/Include/IndustryStandard -I $(MAKEROOT)/Common/ -I .. -I . $(ARCH_INCLUDE) 
+-BUILD_CPPFLAGS = $(INCLUDE) -O2
++BUILD_CPPFLAGS += $(INCLUDE) -O2
+ ifeq ($(DARWIN),Darwin)
+ # assume clang or clang compatible flags on OS X
+-BUILD_CFLAGS = -MD -fshort-wchar -fno-strict-aliasing -Wall -Werror -Wno-deprecated-declarations -Wno-self-assign -Wno-unused-result -nostdlib -c -g
++BUILD_CFLAGS += -MD -fshort-wchar -fno-strict-aliasing -Wall -Werror -Wno-deprecated-declarations -Wno-self-assign -Wno-unused-result -nostdlib -c -g
+ else
+-BUILD_CFLAGS = -MD -fshort-wchar -fno-strict-aliasing -Wall -Werror -Wno-deprecated-declarations -Wno-stringop-truncation -Wno-restrict -Wno-unused-result -nostdlib -c -g
++BUILD_CFLAGS += -MD -fshort-wchar -fno-strict-aliasing -Wall -Werror -Wno-deprecated-declarations -Wno-stringop-truncation -Wno-restrict -Wno-unused-result -nostdlib -c -g
+ endif
+-BUILD_LFLAGS =
+-BUILD_CXXFLAGS = -Wno-unused-result
++BUILD_LFLAGS = $(LDFLAGS)
++BUILD_CXXFLAGS += -Wno-unused-result
+ 
+ ifeq ($(HOST_ARCH), IA32)
+ #
--- a/recipes-core/ovmf/ovmf_git.bbappend
+++ b/recipes-core/ovmf/ovmf_git.bbappend
@@ -1,0 +1,53 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/ovmf:"
+
+SRC_URI = "git://github.com/tianocore/edk2.git;branch=master \
+    file://0002-ovmf-update-path-to-native-BaseTools.patch \
+    file://0003-BaseTools-makefile-adjust-to-build-in-under-bitbake.patch \
+    "
+
+SRCREV="dd4cae4d82c7477273f3da455084844db5cca0c0"
+
+FILES_${PN} += "\
+    /usr/share/firmware/ovmf.bin \
+    "
+
+# This is mostly a copy-paste from the upstream recipe but we need to force
+# OVMF_ARCH to be X64 and there is no easier way to do that. We omit a bunch
+# of symlinks and SecureBoot related logic that are just not applicable.
+do_compile_class-target() {
+    export LFLAGS="${LDFLAGS}"
+    OVMF_ARCH="X64"
+
+    # The build for the target uses BaseTools/Conf/tools_def.template
+    # from ovmf-native to find the compiler, which depends on
+    # exporting HOST_PREFIX.
+    export HOST_PREFIX="${HOST_PREFIX}"
+
+    # BaseTools/Conf gets copied to Conf, but only if that does not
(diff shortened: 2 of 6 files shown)


recipes-core/ovmf/ovmf_git.bbappend
```diff
@@ -0,0 +1,78 @@
...
+      *)
+        FIXED_GCCVER=GCC5
+        ;;
+    esac
+    echo ${FIXED_GCCVER}
+}
+
+do_compile_class-target() {
+    export LFLAGS="${LDFLAGS}"
+    OVMF_ARCH="X64"
```
jandryuk: This is a copy and paste of the oe-core version, but hardcoding OVMF_ARCH and dropping some commands at the end after the build.  I want to recommend trying to override TARGET_ARCH, but I can't think of a way.  It's a shame you have to duplicate this.

recipes-core/ovmf/ovmf_git.bbappend
```diff
@@ -0,0 +1,78 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/ovmf:"
+
+SRC_URI = "git://github.com/tianocore/edk2.git;branch=master \
+    file://0002-ovmf-update-path-to-native-BaseTools.patch \
+    file://0003-BaseTools-makefile-adjust-to-build-in-under-bitbake.patch \
+    "
```
jandryuk: Are the other patches from oe-core no longer needed with the SRCREV specified below?  Usually, a .bbappend just modifies SRC_URI to add or remove patches instead of redefining the entire variable.

This is fine if you want to remove them - it's just a little unconventional.

tklengyel: The patches are not relevant to x86_64 builds and they don't apply cleanly to the revision we picked.

recipes-core/ovmf/ovmf/0003-BaseTools-makefile-adjust-to-build-in-under-bitbake.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/BaseTools/Source/C/Makefiles/header.makefile b/BaseTools/Source/C/Makefiles/header.makefile
```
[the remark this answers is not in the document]
tklengyel: I couldn't get the patch from upstream openembedded to apply because there are some weird non-ascii characters littered in the source so I manually made the changes and did a git diff. Without this patch OVMF doesn't build under openembedded.

recipes-core/ovmf/ovmf_git.bbappend
tklengyel: The curse of having multiple target archs in a single build.

eric-ch: Maybe the following would not cause too much mayhem and do the trick:
```sh
do_compile_prepend_class-target() {
    TARGET_ARCH=x86_64
```

It does not get rid of the following though
```sh
ln ${build_dir}/${OVMF_ARCH}/EnrollDefaultKeys.efi ${WORKDIR}/ovmf/
```

tklengyel:
> Just to be clear, it's only the first boot of Windows 10 that takes a long time to boot? Subsequent boots are fine?

Yes, it's only the installer that takes a while to get started during which nothing is displayed on the screen, just the TianoCore splash, so it can appear frozen. Not sure if it's because of OVMF or not, just something I noticed (didn't try Win10 install with Legacy BIOS so it could be that it's the same).

recipes-core/ovmf/ovmf_git.bbappend
jandryuk: I considered TARGET_ARCH prepend idea, but I think it won't work.  bitbake will expand variables written as ${..} (${TARGET_ARCH}) when it writes the script file run.do_compile, so we can't play games to override the variable at runtime.

tklengyel: Getting rid of those symlinks is not important, it was just something we don't need so I got rid of them. If we could just use `do_compile_class-target` as-is by overriding the target arch that would suffice. I will give your suggestion a shot.

tklengyel: It keeps building IA32 version even with
```
do_compile_prepend_class-target() {
    export TARGET_ARCH="x86_64"
}

**Comment by richpersaud** [comment]
Thanks for the submission.  

Could you identify the portions of this PR which could be submitted upstream? E.g. is this compatible with the Q35 work that has been discussed on xen-devel?

Do you know why OpenXT needs legacy ROM support in seabios, e.g. which OpenXT drivers create the requirement?

**Comment by tklengyel** [comment]
> Could you identify the portions of this PR which could be submitted upstream? E.g. is this compatible with the Q35 work that has been discussed on xen-devel?

This is largely based on upstream version of support for OVMF in Xen. Only difference compared to stock Xen is the version of OVMF we build (we build a newer one) and that we don't build OVMF using the Xen build scripts (because of openembedded tweaks being necessary). I don't think the Q35 should conflict, KVM already supports Q35 and it works fine with this version of OVMF. I have it on my TODO to ping xen-devel notifying the maintainer that the version of OVMF currently tagged has issues with new Windows 10 images so they should do an uprev at some point.

> Do you know why OpenXT needs legacy ROM support in seabios, e.g. which OpenXT drivers create the requirement? Is that needed for non-Windows guests?

We haven't looked at the SeaBIOS more in-depth, I just noticed it too that it has some custom Option ROMs. Not sure what that's for either. With OVMF we haven't noticed any issues running Windows 10 or Linux as guests.

**Comment by eric-ch** [comment]
> Do you know why OpenXT needs legacy ROM support in seabios, e.g. which OpenXT drivers create the requirement? Is that needed for non-Windows guests?

OpenXT exposes two option ROMs to guests by default, IPXE and VGABIOS, for guest legacy boot. Ipxe is used for in-guest PXE boot. VGABios initializes the emulated VGA card and add a couple of VBE functions used to handle the display without PV drivers (early display && initial installation).

**Comment by jandryuk** [comment]
```
TARGET_ARCH_task-compile = "x86_64"
HOST_PREFIX_task-compile = "i686-oe-linux-"
STAGING_BINDIR_TOOLCHAIN_task-compile = "${STAGING_DIR_NATIVE}${bindir_native}/i686${TARGET_VENDOR}-${TARGET_OS}"
```
Overriding `TARGET_ARCH` makes the build set OVMF_ARCH to X64, but then the build fails when it can't find the x86_64-oe-linux-gcc.  Overriding `HOST_PREFIX` switches to i686-oe-linux-gcc, but we still fail to build since we can't find it in `PATH`.  Overriding `STAGING_BINDIR_TOOLCHAIN` fixes up `PATH` so i686-oe-linux-gcc can be found and the build succeeds.

While this works, it's hard-coding variables and will break if the assumption is invalid.  Therefore I think it's better to leave your current `do_compile_class-target`.  Maybe add a comment saying it's a copy & paste to force the 64bit build for the 32bit dom0?  That was people know it can be removed in the future if dom0 becomes 64bit.

deepin125:
I was wondering if this has been tested with UEFI PXE boot?  Are the UEFI firmware drivers present for the e1000 emulated device that is presented via QEMU?

tklengyel:
@deepin125 OVMF does support PXE boot but we haven't tested it.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1739

jean-edouard:
This PR builds fine, along with https://github.com/OpenXT/idl/pull/31, https://github.com/OpenXT/manager/pull/126 and https://github.com/OpenXT/toolstack-data/pull/50.
Guest UEFI is not enabled by default and seems to work pretty well (the latest Windows 10 installs and boots).
The code has been reviewed and approved by multiple people.
I think it's ready to merge, and I will do so tomorrow if there's no objection.
Thanks!

tklengyel:
Following @deepin125 question I looked into PXE boot with OVMF and it is not provided by default under Xen. For some reason Xen specifically disables PCI enumeration and consequently the `efi-e1000.rom` is not loaded as an option rom either though QEMU serves it when available. Workaround is to copy that rom to a filesystem accessible by the vm and run `loadpcirom` from the OVMF provided EFI shell. Afterwards the PXE boot options become available in the firmware menu.

jean-edouard:
Thanks for looking into it. Probably worth a ticket.

PR #1038, opened 2018-09-13
OXT-1432: xen: Amend xen.efi LoadOptions handling.

eric-ch:
A more thorough sanity check is required in get_argv() to determine the
content of LoadOptions and its relevance to the cmdline:

Here is an example of LoadOptions returned by EFI boot-services on two
different platforms:

On a Dell 7450:
>00000000  01 00 00 00 62 00 4f 00  70 00 65 00 6e 00 58 00  |....b.O.p.e.n.X.|
>00000010  54 00 00 00 04 01 2a 00  01 00 00 00 00 08 00 00  |T.....*.........|

This content would have get_argv() return argc = 1, while nothing is
really usable to the cmdline.
In turn, efi_arch_handle_cmdline() would not prepend "xen" to the
cmdline set in the MBI structure.

On the other hand, on an HP 800 G1:
>00000000  d0 9b 00 f0 d0 9b 00 f0  c3 e2 00 f0 d0 9b 00 f0  |................|
>00000010  d0 9b 00 f0 54 ff 00 f0  47 66 00 f0 d0 9b 00 f0  |....T...Gf......|

This content would have get_argv() return argc = 0, which is legitimate.
In turn, efi_arch_handle_cmdline() would prepend "xen" to the cmdline
set in the MBI structure then passed to TBoot, in OpenXT.

This behavior will result in TBoot measurement failing on PCR18 when
performing forward sealing as the command-line used by the forward
sealing scripts will not include the prepended "xen " statement.

Per the author of the patch, efi-load-option-support.patch was an early
iteration of a patch now upstreamed, which can be backported in its
place and handles the LoadOptions with more consistency.

Now efi_arch_handle_cmdline() will, in both cases, prepend "xen" to the
MBI cmdline. Since this is not the desired behavior on OpenXT EFI/DRTM
bootstrap, the image name is not set like that, a second patch will only
prepend the image name to the command-line if it is set in LoadOptions,
and not rely on a default value.

OXT-1432


eric-ch:
Built in [$1731](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1731)

jean-edouard:
LGTM.
@tklengyel, as the author of the patch, do you approve this PR?
Was there a reason for not including it into OpenXT earlier?
Thanks

tklengyel:
@jean-edouard the upstream version of the patch was not backported to #890 because there was no observable difference on sane platforms between that and the earlier iteration of the patch, so the idea was to simply drop the patch once Xen 4.12 is released and OpenXT is upreved. Now that we have an obviously buggy firmware on the G1 it makes sense to backport it as it is doing even more sanity checking then the earlier version did. I would still not recommend using UEFI on that platform though as it's very likely that other things are broken on it too.

@eric-ch are the changes in the other patches due to rebasing on top of the new patch?

jean-edouard:
Merging soon.

PR #1034, opened 2018-08-27
OXT-1421: XSA-270 Linux netback driver OOB access in hash handling 

jandryuk:
https://xenbits.xen.org/xsa/advisory-270.html

This is not upstream yet.

OXT-1421

This is on top of https://github.com/OpenXT/xenclient-oe/pull/1031.  Once that is merged, it is just a single commit.

PR #1039, opened 2018-09-13
OXT-1303: linux: Proper fix for OOPS __radix_tree_lookup

eric-ch:
Backport upstream fixes for the OOPS:
"NULL pointer dereference EIP: __radix_tree_lookup"

As suggested on the thread upstream, this was most likely a race when
clearing PTE using 32bit writes instead of an atomic 64bit operation. As
it happens this was also resulting in an L1TF vulnerable PTE.

OXT-1303, OXT-1418


eric-ch:
@jandryuk Since I did not find the new changeset in 4.14.69 I thought it may have been interesting to backport it... If anything it will end up in more build and receive more test.

I have had that patch, without the bandaid, for some time now and couldn't reproduce the issue, with the reboot loop or in other uses. I do not know any other way to reproduce it with better odds.

The explanation from Juergen on the LKML sounds fit IMHO. It corroborates not seeing the behavior with the new changes as well.

My bad I meant to git rm the bandaid, thanks!

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1735

eric-ch:
Rebased post #1034 merge.

PR #1042, opened 2018-09-20
OXT-1436: xenclient-idl: Add recipe version

jandryuk:
xenclient-idl lacked a version since it was a git recipe, so bitbake
assigned it version git-r0.  When idl.git changed, dependency tracking
was thrown off since the version remained the same.  This would break
rebuilds of dependant packages since their sysroots would not be updated
with the newer version.

Set PV so the recipe is versioned and dependencies can be tracked.

OXT-1436

### DIFF
--- a/recipes-openxt/xenclient/xenclient-idl_git.bb
+++ b/recipes-openxt/xenclient/xenclient-idl_git.bb
@@ -2,6 +2,8 @@ DESCRIPTION = "XenClient IDL definitions + rpc stubs generation mechanism"
 LICENSE = "GPLv2"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/GPL-2.0;md5=801f80980d171dd6425610833a22dbe6"
 
+PV = "0+git${SRCPV}"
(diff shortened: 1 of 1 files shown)


eric-ch:
@jandryuk Thanks for fixing this, looking at the logs I removed it during the haskell layer refactoring during the Pyro upgrade, my bad.

jandryuk:
@eric-ch no worries :).  I didn't even look at the history - I just realized 'git-r0' was an unusual version.

eric-ch:
@jandryuk I think there may be other instances of that mistake.

Examples I could find with a quick grep:
- xenclient-rpcgen.inc
- dbd_git.bb
- uid_git.bb

jandryuk:
Good thinking to grep.  I also just checked in the tmp-glibc/work and heimdallr and ovmf also show up with recipe version "git-r0".  I'm a little surprised the upstream ovmf lacks a PV definition

jean-edouard:

jean-edouard:
Merging soon.

PR #1043, opened 2018-09-20
OXT-1441: libxl patch cleanup

jandryuk:
There are some unneccesary changes to tools/libxl/libxl_dm.c. This removes them.  We only use qemu-xen, so we don't need to touch libxl__build_device_model_args_old used by qemu-traditional.


jandryuk:
I re-pushed after removing a few stray comments.

jean-edouard:

jean-edouard:
LGTM, merging soon.

PR #1044, opened 2018-09-21
OXT-1437: Fix Guest 802.1x

jandryuk:
From the description, linux patch break-8021d.patch is supposed to forward 802.1d link local packets like 802.1x PAE from the guest down to NDVM. Switches are not supposed to forward these link local packets - hence the name break_8021d.

break_8021d is itself broken in current linux since it only controls Spanning Tree (STP/RSPT/MSTP) (01-80-C2-00-00-00) - 802.1x PAE packets (01-80-C2-00-00-03) would not be forwarded. Upstream linux has since gained group_fwd_mask to forward more link local traffic - though a few packet types will not be forwarded. For the stubdom, we really just want to be a blind passthrough, so we should just pass everything we can.

The restricted link-local packet types are:
01-80-C2-00-00-00 Spanning Tree (STP/RSPT/MSTP)
01-80-C2-00-00-01 Ethernet Flow Control (pause frames)
01-80-C2-00-00-02 Link Aggregation Control Protocol (LACP)

Disabling STP on a bridge will forward STP packets leaving just the pause and LACP.

Network slave disables stp on bridges it creates, but does not use break_8021d. 802.1x would only make sense on brbridged with a single guest running.

Options:
1) Drop support
2) Only allow passing 802.1x packets
3) Pass all link local packets allowable by linux
4) Patch linux to forward all link local packets

Working on option 3 since it doesn't require patching linux.

Works with: https://github.com/OpenXT/network/pull/12


jean-edouard:

rneilturner:
+1 LGTM

jean-edouard:
Thanks @rneilturner, merging soon.

PR #1046, opened 2018-09-25
syncvm: Specify read-only-rootfs IMAGE_FEATURES

jandryuk:
The syncvm image is read-only, so specify that.  Most visibly, this
removes a bunch of warnings about populating the udev cache when the
rootfs is read-only.


jean-edouard:
LGTM, merging soon.

PR #1047, opened 2018-09-25
OXT-1445: QEMU 3.0

jandryuk:
This uprevs to QEMU 3.0.  Changes are pretty straightforward, but two notable items.

libxl is [changed](https://github.com/OpenXT/xenclient-oe/commit/f1412cfa16d84650ffb8a1db2143737926ec6a63) to leave cdroms as format=host_cdrom to avoid this warning:

> warning:  Opening a block device as a file using the 'file' driver is deprecated

@jean-edouard I think you introduced the patch to set format=file.  Changing ISO images works, but I'm not sure why this patch was originally introduced.  Do you think this is okay?

atapi passthrough is untested, so testing is appreciated.

PR #1048, opened 2018-10-01
Xen: upgrade to 4.11.0

jean-edouard:


jean-edouard:
Needs https://github.com/OpenXT/xsm-policy/pull/21

jandryuk:
The upgrade is a massive 3000+ line commit.  Would you please split out patches that require a code change into their own commit?  Then use a separate commit for the context refresh of the rest of patch queue - like I've structured the [qemu 3.0 patchset](https://github.com/OpenXT/xenclient-oe/pull/1047).

The revert of "x86/boot: Make alternative patching NMI-safe" needs explanation of *why* it is needed.  Having said that, I don't think just reverting the patch is acceptable.  Upstream Xen wants NMI-safe alternative patching, so we'll need a fix to make Xen under tboot compatible.

jean-edouard:
@jandryuk the only patch that actually has real code changes is libxl-vwif-support.patch, and even there it's just a handful of lines after expanding 2 macros...
About the revert, I don't have more information... I just noticed that it broke xen-on-tboot so I reverted it... If you have a better solution I'll take it.

jandryuk:
What are the details of "broke xen-on-tboot"?  What exactly broke?  Those need to be included in the patch file.  Vanilla Xen is fine and it's only an issue when tboot launches Xen?

Identifying the commit is great and it'll help find the best fix.

rossphilipson:
@jean-edouard are you saying that the  "x86/boot: Make alternative patching NMI-safe" patch breaks TBOOT+Xen combination? I think that is what you are saying. The problem may be that TBOOT does not explicitly re-enable NMIs that the ACM disabled. I guess it assumes some other interrupt will come along and enable it at some point. So maybe things are getting jacked up because NMIs are still disabled? You could try to explicitly re-enable them in TBOOT by doing an IRET to self. Our Trenchboot code does that here:

https://github.com/rossphilipson/travail/blob/master/trenchboot/tb_stub/0002-x86-Trenchboot-kernel-early-boot-stub.patch#L144

Anyway just a possibility that is what is going on...

richpersaud:
Please create a JIRA epic for Xen 4.11 upgrade, then tickets below that for specific issues such as tboot.

jean-edouard:
+1 to that, props to @rossphilipson! :)

jean-edouard:
But let's not forget the most important thing: creating a ticket :)

jandryuk:
Nice.  Thanks @rossphilipson and @eric-ch 

Is/was this only a problem under UEFI?  A legacy Ivy Bridge machine booted fine.  I tried sticking an iret into Xen's 32bit boot code, but a UEFI test machine still wouldn't boot.  I may not have covered the correct entry point through...

eric-ch:
@jandryuk I tried a Dell 7470 on both Legacy and UEFI without problem. I do not have an Ivy Bridge at hand. Can you try [$1760](https://openxt.ainfosec.com/builds/custom/master/1760/iso/) on Ivy Bridge?

jandryuk:
@eric-ch the Ivy Bridge only boots legacy... at least as far as I am concerned.  It's a Skylake that failed UEFI boot.  I'll test that one.  (It also doesn't have a serial port or AMT, so early boot issues like this are a pain.)

jandryuk:
The Skylake machine booted up (UEFI) with build 1760.

eric-ch:
@jandryuk Thanks!

PR #1050, opened 2018-10-04
OXT-1446: tboot: Unmask NMI for Xen 4.11 patching.

eric-ch:
At boot time, [Xen 4.11 patches alternatives in NMI context](https://github.com/xen-project/xen/commit/5191c1ef51) to avoid
NMI or MCE happening while altering their code path. If the self-NMI has
been masked during an SENTER (and apparently it is in some cases), Xen
will panic() with:
    "Timed out waiting for alternatives self-NMI to hit"

Patch TBoot to unmask NMI that may have been masked during SENTER.

Remove the reverted change introducing patching in NMI context from the
Xen patch-queue.

Thanks to Ross Philipson at Oracle for his detailed explanations on how
to solve this.

### DIFF
--- a/recipes-openxt/tboot/tboot-1.9.6/0023-tboot-Unmask-NMI-potentially-masked-during-SENTER.patch
+++ b/recipes-openxt/tboot/tboot-1.9.6/0023-tboot-Unmask-NMI-potentially-masked-during-SENTER.patch
@@ -1,0 +1,89 @@
+From 4412c46d5ce1f392e1893c443ef0a6d4a8441cc0 Mon Sep 17 00:00:00 2001
+From: Eric Chanudet <chanudete@ainfosec.com>
+Date: Wed, 3 Oct 2018 14:22:09 -0400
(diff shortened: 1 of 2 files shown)


eric-ch:
Actually we might want to unmask NMI later (e.g, where SMI are re-enable) to protect the MLE.

rossphilipson:
And unmask it after AP startup too. I agree that doing it where SMI is re-ennabled is a good place.

rossphilipson:
Updated patch looks correct to me. I am not able to test it but I will give you my R-b
Reviewed-by: Ross Philipson <ross.philipson@oracle.com>

eric-ch:
@rossphilipson Thanks! Commit amended.

eric-ch:
Building: [$1765](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1765)

jandryuk:
I tested this PR on top of the Xen 4.11 changes from https://github.com/OpenXT/xenclient-oe/pull/1048 and it boots.

jean-edouard:
LGTM, merging soon.

PR #1059, opened 2018-11-14
xen: upgrade to 4.11

jean-edouard:

### DIFF
--- a/recipes-extended/xen/files/allow-stubdoms-cacheattr-control.patch
+++ b/recipes-extended/xen/files/allow-stubdoms-cacheattr-control.patch
@@ -22,14 +22,11 @@ attributes configured in the new pinned region when mappings are introduced.
 Patch changes:
 
 * Reorder the EPT page table entry logic to respect memory regions that have
-  been pinned via the XEN_DOMCTL_pin_mem_cacheattr operation, even for guests
+  been pinned via the XEN_DMOP_pin_memory_cacheattr operation, even for guests
   that do not have I/O privileges.
 
-* Update the dummy XSM policy, active when XSM is disabled, to grant
-  stubdomains permission to use XEN_DOMCTL_pin_mem_cacheattr on their guest.
-
 If appropriate for your trust model, guests themselves can be prevented from
-using XEN_DOMCTL_pin_mem_cacheattr via XSM policy.
+using XEN_DMOP_pin_memory_cacheattr via XSM policy.
 
 ################################################################################
 CHANGELOG
@@ -68,26 +65,16 @@ INTERNAL DEPENDENCIES
 ################################################################################
 
 Recommended: XSM policy should grant stubdomains access to
-XEN_DOMCTL_pin_mem_cacheattr applied to their target guests.
+XEN_DMOP_pin_memory_cacheattr applied to their target guests.
 It should disallow guests themselves access to it unless needed for passed
 through devices.
 
 ################################################################################
 PATCHES
 ################################################################################
---- a/xen/include/xsm/dummy.h
-+++ b/xen/include/xsm/dummy.h
-@@ -127,6 +127,7 @@ static XSM_INLINE int xsm_domctl(XSM_DEF
-     case XEN_DOMCTL_memory_mapping:
-     case XEN_DOMCTL_bind_pt_irq:
-     case XEN_DOMCTL_unbind_pt_irq:
-+    case XEN_DOMCTL_pin_mem_cacheattr:
-         return xsm_default_action(XSM_DM_PRIV, current->domain, d);
-     case XEN_DOMCTL_getdomaininfo:
-         return xsm_default_action(XSM_XS_PRIV, current->domain, d);
 --- a/xen/arch/x86/hvm/mtrr.c
 +++ b/xen/arch/x86/hvm/mtrr.c
-@@ -798,12 +798,6 @@ int epte_get_entry_emt(struct domain *d,
+@@ -803,12 +803,6 @@ int epte_get_entry_emt(struct domain *d,
          return MTRR_TYPE_UNCACHABLE;
      }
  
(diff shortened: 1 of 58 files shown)


jean-edouard:
Replaces https://github.com/OpenXT/xenclient-oe/pull/1048
Depends on https://github.com/OpenXT/xsm-policy/pull/21

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1790

jean-edouard:
No ticket.
The vwif comments are self explanatory and I don't want to have a discussion about that patch.
I can remove the mentions of deprecation if they pose a problem.

recipes-extended/xen/files/libxl-vwif-support.patch
```diff
@@ -1,36 +1,61 @@
...
+A small app should be written for every supported operating system.
+That app would have a tray icon that shows the signal strength, while allowing
+the NIC to be a regular VIF, driven by the standard netfront.
+With such an app, this patch could be discarded, and frontend driver(s) could be
+thrown away.
 
 ################################################################################
 UPSTREAM PLAN
 ################################################################################
+Please don't.
```
jandryuk: :)

richpersaud:
> No ticket.

Please use 
https://openxt.atlassian.net/browse/OXT-1454 for this PR.

> The vwif comments are self explanatory and I don't want to have a discussion about that patch.
> I can remove the mentions of deprecation if they pose a problem.

I have filed
https://openxt.atlassian.net/browse/OXT-1455 to track related topics.

recipes-extended/xen/files/allow-stubdoms-cacheattr-control.patch
```diff
@@ -77,17 +77,17 @@ PATCHES
 ################################################################################
 --- a/xen/include/xsm/dummy.h
 +++ b/xen/include/xsm/dummy.h
-@@ -127,6 +127,7 @@ static XSM_INLINE int xsm_domctl(XSM_DEF
+@@ -141,6 +141,7 @@ static XSM_INLINE int xsm_domctl(XSM_DEF
      case XEN_DOMCTL_memory_mapping:
      case XEN_DOMCTL_bind_pt_irq:
      case XEN_DOMCTL_unbind_pt_irq:
-+    case XEN_DOMCTL_pin_mem_cacheattr:
++    case XEN_DMOP_pin_memory_cacheattr:
```
jandryuk: I think this should be dropped since it's covered by dm_op.  There is a separate xsm_dm_op function in `xen/include/xsm/dummy.h` for the dm_ops.  It is already 
```
    XSM_ASSERT_ACTION(XSM_DM_PRIV);
    return xsm_default_action(action, current->domain, d);
```
so pin_memory_cacheattr is already covered the same way.

Also, XEN_DOMCTL_pin_mem_cacheattr (41) != XEN_DMOP_pin_memory_cacheattr (18), so this is probably allowing XEN_DOMCTL_setdebugging (18)

jean-edouard: Good catch, will drop.
Do you understand the second part of that patch by any chance? Should that be kept?

jandryuk: I can only go from the description.  @dozylynx re-ordered the code - I guess to get to hvm_get_mem_pinned_cacheattr before the `if ( !need_iommu(d) && !cache_flush_permitted(d) )` check.  That way memory can be set to write combine - otherwise it would be write back.  So I guess it's still needed.

jean-edouard:
Thanks for the reviews.
I force-pushed fixes to address them.
Looks like GitHub can now show a force-push diff, pretty cool!

jean-edouard:
Also, built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1793

recipes-extended/xen/files/fix-memcpy-in-x86-emulate.patch
```diff
@@ -83,17 +83,17 @@ PATCHES
...
+@@ -263,6 +263,26 @@ hvm_emulate_insn_fetch(enum x86_segment
+     return X86EMUL_OKAY;
  }
  
 +/*
-+ * Routine to make sh_x86_emulate_write appropriate to use for copying the
++ * Routine to make hvm_emulate_write appropriate to use for copying the
 + * results of instruction emulation back to guest memory - these
 + * typically require 64-bit, 32-bit and 16-bit writes to be atomic
 + * whereas memcpy is only atomic for 64-bit writes. This is still
```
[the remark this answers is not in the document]
eric-ch: See [OXT-1460](https://openxt.atlassian.net/browse/OXT-1460) to address that. This can be done after this PR is merged.

recipes-extended/xen/files/xen-x86-Fix-up-rules-when-forcing-mno-sse.patch
```diff
@@ -36,12 +36,12 @@ PATCHES
...
--CFLAGS += -mno-red-zone -mno-sse -fpic
-+CFLAGS += -mno-red-zone -mno-sse -fpic -mfpmath=387
- CFLAGS += -fno-asynchronous-unwind-tables
- # -fvisibility=hidden reduces -fpic cost, if it's available
- ifneq ($(call cc-option,$(CC),-fvisibility=hidden,n),n)
+-CFLAGS += -mno-red-zone -fpic -fno-asynchronous-unwind-tables
++CFLAGS += -mno-red-zone -fpic -fno-asynchronous-unwind-tables -mfpmath=387
+ 
+ # Xen doesn't use SSE interally.  If the compiler supports it, also skip the
+ # SSE setup for variadic function calls.
```
jandryuk: Looks like '-mno-sse' moved to just after this patch context.  '-mfpmath=387' may make more sense being added there.
```
CFLAGS += -mno-red-zone -fpic -fno-asynchronous-unwind-tables

# Xen doesn't use SSE interally.  If the compiler supports it, also skip the
# SSE setup for variadic function calls.
CFLAGS += -mno-sse $(call cc-option,$(CC),-mskip-rax-setup)
```

jean-edouard: New PR up with this fixed: https://github.com/OpenXT/xenclient-oe/pull/1069

jean-edouard:
Replaced by https://github.com/OpenXT/xenclient-oe/pull/1069, to avoid a potentially destructive force-push. Closing.

PR #1061, opened 2018-11-15
installer: remove console=tty2 on the kernel cmdline

jean-edouard:
In the the installer, we don't want the kernel to print anything on screen.
The kernel still logs to dmesg and /var/log/kern.log.
Additionally, the installer shows /var/log/messages in tty4.
With this fix, the installation dialogs should not be disrupted anymore.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1791

jandryuk:
I thought we have:
tty1 - installer with dialog messages
tty2 - cat installer log
tty3 - root prompt
tty4 - cat /var/log/messages

wouldn't console=tty2 only affect the installer.log VT?

What exactly are you trying to solve?  I think there are some places in the installer that may lack the >2 redirect to avoid flashing the tty1 installer dialog commands.

jean-edouard:
@jandryuk for some reason, "console=tty2" actually seems to print kernel messages in the current tty, at least in UEFI.
Not too sure why, but removing that fixes the issue, but it didn't make sense anyway...

eric-ch:
@jandryuk (about /dev/tty2 in isolinux.cfg) Oh right, I got confused...

eric-ch:
The spurious message I have when booting the UEFI installer from a USB key:
```
sd 2:0:0:0: [sda] No Caching mode page found
sd 2:0:0:0: [sda] Assuming drive cache: write-through
```

It happens on the first screen.

jandryuk:
I ran a USB install on UEFI this morning.  I did not see those particular sd messages, but some udevd messages popped through.  `ls -l /proc/$( pgrep udevd )/fd` shows udev is properly daemonized with fds 0-2 set to /dev/null.  udev does have /dev/kmsg open and will print messages there.

```
udevd[2453]: conflicting device node '/dev/mapper/temporary-cryptsetup-2542' found, link to '/dev/dm-8' will not be created
udevd[2544]: conflicting device node '/dev/mapper/temporary-cryptsetup-2590' found, link to '/dev/dm-8' will not be created
udevd[2453]: conflicting device node '/dev/mapper/temporary-cryptsetup-2590' found, link to '/dev/dm-8' will not be created
udevd[2544]: conflicting device node '/dev/mapper/temporary-cryptsetup-2619' found, link to '/dev/dm-8' will not be created
udevd[2453]: conflicting device node '/dev/mapper/temporary-cryptsetup-2634' found, link to '/dev/dm-8' will not be created
udevd[2453]: conflicting device node '/dev/mapper/config' found, link to '/dev/dm-8' will not be created
udevd[2453]: conflicting device node '/dev/mapper/log' found, link to '/dev/dm-9' will not be created
udevd[2721]: conflicting device node '/dev/mapper/cores' found, link to '/dev/dm-10' will not be created
```
I guess that's https://github.com/OpenXT/xenclient-oe/pull/1065?

Anyway, looking at /proc/console, it shows ttyS0, tty2 & netcon0.  Dropping console=tty2 removes it from /proc/consoles and seems to remove the messages from the screen.

Go to tty3 and log in as root.
`sleep 5 ; echo '<0>foo' > /dev/kmsg`
alt-f1 to switch back to tty1
See 'foo' pop through.  Drop console=tty2 and it does not appear.

https://elixir.bootlin.com/linux/v4.14.95/source/drivers/tty/vt/vt.c#L2519
```
 * By default, the kernel messages are always printed on the current virtual
 * console. However, the user may modify that default with the
 * TIOCL_SETKMSGREDIRECT ioctl call.
```
So I guess that's the magic for the console output.

eric-ch:
@jandryuk Correct for https://github.com/OpenXT/xenclient-oe/pull/1065. Thanks for investigating and explaining it.

eric-ch:
Will merge soon.

PR #1063, opened 2019-01-07
openxt-main: Remove nfs from DISTRO_FEATURES

jandryuk:
packagegroup-base creates a subpackage packagegroup-base-nfs when nfs is
set in DISTRO_FEATURES.  This installs and runs rpcbind in all the
OpenXT images since nfs is set in DISTRO_FEATURES_DEFAULT.

Drop nfs from DISTRO_FEATURES to omit rpcbind from the images.


jean-edouard:
Yes, thanks!
I already tested that, I thought I PRed it but apparently not!
Will merge soon.

PR #1064, opened 2019-01-08
OXT-1463: Xorg recipes cleanup.

eric-ch:
Remove old deprecated recipes from the layer and upgrade to the ones provided in openembedded-core.

`Xenfb2`/`resized` still requires a patch for video driver fbdev to handle UIVM resizing.

Move image specific configuration out of the machine configuration while we're at it.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1810

jean-edouard:
LGTM, merging soon.

jandryuk:
This PR went in before I noticed it.  I like it and in fact I had a work-in-progress branch of the same thing which I never got around to finishing for posting.

The main hold up was I thought the drawing performance on the newer xserver was worse.  Open a terminal, resize it until it covers maybe 1/3 of the screen, and drag it around quickly in circles.  At some point, I could get it to keep redrawing for a second (or two?) after I let go of the mouse.  It may have been somewhat better after some further changes, but I never tracked it down further.  In general, I felt like the drawing was slower.  Maybe it was just the 1920x1080 resolution and software rendering combined with the particular Intel graphics card?  Have you seen anything like this?

If you drop the xorg.conf, xserver will default to screen blanking.  Do we care about that change?  Also, I think dropping DontVTSwitch means it defaults to enabled, but maybe that doesn't matter since no other VTs are running anything?  Also, I think fbdev will default to ShadowFB "on" while the old setting was "off".

eric-ch:
@jandryuk Some dependencies piled up (webkit...), but this was motivated by starting to upgrade the OE layer to sumo a couple of months ago.

I did not notice graphic performances being decreased in general, but I had the feeling that the scaling effect on the VM icons sometimes get "stuck" when the pointer passes over it... It does not always happen and I was not sure this came from that change, but now I am not quite sure.

Xorg does blank the framebuffer following the DPMS settings. It does not blank the screen, so no real power save there.

DontVTSwitch should be "on", you are correct, switch will trigger all sorts of problem with xenfb2.

I believe you are correct about ShadowFB, I wonder if that explains the presumed graphic glitch mentioned above.

eric-ch:
@jandryuk I experimented with the `xorg.conf` configuration.

One thing was very clear, turning off the `ShadowFB` makes everything terribly slow. I interpret it as X will draw directly in the linear-fb region which will generate more events between front & backend, while having the `ShadowFB` will tend to coalesce things. Moving a screen-tall terminal sideways will look like it was before with `ShadowFB`, and be terribly slow and use ~100%CPU in UIVM without.

jandryuk:
@eric-ch "Moving a screen-tall terminal sideways will look like it was before with ShadowFB, and be terribly slow and use ~100%CPU in UIVM without."  To be clear, which is `ShadowFB=on` and which is `ShadowFB=off`?

jandryuk:
Interesting.  The old config had:
```
        Option     "ShadowFB" "Off"
```

Assuming "ShadowFB" "Off" == "ShadowFB" "false", it's interesting you need to set "ShadowFB" "true" to restore performance on the new X server.

PR #1065, opened 2019-01-09
lvm udev rules: avoid trying to create bad symlinks

jean-edouard:
lvm expects /dev/mapper to be populated with symlinks,
but instead it contains actual blk files.
Remove the rule that creates those symlinks.

OXT-1274

### DIFF
--- a/recipes-support/lvm2/lvm2/test-for-blk-file-before-creating-dm-mapper-symlink.patch
+++ b/recipes-support/lvm2/lvm2/test-for-blk-file-before-creating-dm-mapper-symlink.patch
@@ -1,0 +1,26 @@
+--- a/udev/10-dm.rules.in
++++ b/udev/10-dm.rules.in
+@@ -132,7 +132,11 @@ ENV{DM_SUSPENDED}=="Suspended", ENV{DM_S
+ # VSN 2 - add support for synthesized events
+ ENV{DM_UDEV_RULES_VSN}="2"
+ 
+-ENV{DM_UDEV_DISABLE_DM_RULES_FLAG}!="1", ENV{DM_NAME}=="?*", SYMLINK+="(DM_DIR)/$env{DM_NAME}"
++# OpenXT: at this point, /dev/mapper is already populated with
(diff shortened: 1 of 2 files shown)


jandryuk:
If lvm expects to create symlnks, why do we have the blk files in the first place?  Should we just stop creating those?

jean-edouard:
Yeah that's the first thing I tried, but I couldn't figure it out.
Also it's not clear whether those should really be symlinks or blkfiles.
Also changing from one to the other has the potential of creating a whole new series of problems, starting with the SELinux policy.
So in the end I decided to just not try to change the way it works, but just avoid messing up the installer dialogs and dom0 dmesg by adding that check to get rid of the (harmless) errors.

jandryuk:
My workstation has /dev/mapper/fedora-root and /dev/fedora/root as symlinks to /dev/dm-0.  Are we doing something unexpected that creates the blk nodes instead of the symlinks?  I've seen commits from @eric-ch that look like LVM and udev have been troublesome in the past.

Looking quickly, the Fedora Dracut initramfs has udev install and creating the nodes.  OpenXT initramfs uses `vgscan --mknodes`.  Maybe that's the difference?

eric-ch:
@jandryuk That may be it. I am confused by LVM & device-mapper behavior. There are [mentions](https://bugzilla.redhat.com/show_bug.cgi?id=869253) of such behavior with `/dev/mapper` having blkdev instead of symlinks.

Maybe using `vgscan --mknodes` is conflicting with the `switch_root` to a udev managed environment, in which case, removing the `vgscan` and starting udevd in the initramfs could cut it?

`vgscan --mknodes` is presumed automatic with LVM2 (says some [vgscan(8)](https://linux.die.net/man/8/vgscan), but not all [vgscan(8)](http://man7.org/linux/man-pages/man8/vgscan.8.html)).

eric-ch:

eric-ch:
Any issue merging this to silence the warning and move investigation/fix to a futur ticket?

jandryuk:
Yes, this can be merged.

eric-ch:
Logged in [OXT-1520](https://openxt.atlassian.net/browse/OXT-1520).

Will merge soon.

PR #1066, opened 2019-01-10
Removes line that was deleting pkg-config directory/files

tijko:


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1814

jean-edouard:
LGTM, merging soon.

PR #1067, opened 2019-01-14
OXT-1463: xorg: Partially restore former configuration.

eric-ch:
- DontVTSwitch:
    Disable switch to another "virtual terminal" (Ctrl+Alt+Fn) which can
    trigger graphic glitches with xenfb2.
- DontZap:
    Disable XKB action terminating X session (Ctrl+Alt+Backspace).
- BlankTime:
    Set inactivity timeout before blanking the screen, 0 disables the
    phase. This restore past behavior.
- Device/ShadowFB:
    Enable ShadowFB explicitely. It is by default, but left in place as
    it appears to slow down UIVM significantly when disabled.

~- DPMS is not implemented for fbdev, so remove the `StandbyTime`, `SuspendTime`, `OffTime`.~
Actually it is enforced with the fbdev patch... So add back the values to avoid the blanking.

- AutoAddDevices defaults to on and removes the need to set `InputDevice` sections in the configuration based on absolute paths (`/dev/input/eventX`).
- `Device` section for "intel" is legacy.

### DIFF
--- a/recipes-graphics/xorg-xserver/xserver-xf86-config/xenclient-uivm/xorg.conf
+++ b/recipes-graphics/xorg-xserver/xserver-xf86-config/xenclient-uivm/xorg.conf
@@ -1,0 +1,16 @@
+# Xorg configuration for UIVM.
+
+Section "ServerFlags"
+        Option "DontZap"        "true"
+        Option "DontVTSwitch"   "true"
+        Option "BlankTime"      "0"
+        Option "StandbyTime"    "0"
+        Option "SuspendTime"    "0"
+        Option "OffTime"        "0"
+EndSection
+
+Section "Device"
+        Identifier "fbdev0"
+        Driver     "fbdev"
+        Option     "ShadowFB"   "true"
+EndSection
--- a/recipes-graphics/xorg-xserver/xserver-xf86-config_0.%.bbappend
+++ b/recipes-graphics/xorg-xserver/xserver-xf86-config_0.%.bbappend
@@ -1,0 +1,2 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${BPN}:"
+SRC_URI += "file://xorg.conf"


recipes-graphics/xorg-xserver/xserver-xf86-config/xenclient-uivm/xorg.conf
```diff
@@ -0,0 +1,20 @@
...
+        Option "BlankTime"      "0"
+EndSection
+
+Section "Device"
+        Identifier "fbdev0"
+        Driver     "fbdev"
+        Option     "ShadowFB"   "false"
+EndSection
+
+Section "Screen"
```
jandryuk: Looking at my WIP branch, I don't have a Screen section.  I think it is created automatically, so that could be dropped.

eric-ch: Oh nice, I could not find if the `DefaultDepth` was always set at 24, although it seems to be the case everywhere.

eric-ch: ... Sorry I spoke too soon. There is definitely a slowdown without `DefaultDepth 24`, which leads me to believe this is not the default value. [The documentation](https://www.x.org/releases/X11R7.7/doc/man/man5/xorg.conf.5.xhtml) mentions that the default is usually 8.

eric-ch:
v2:
- Switch `Option "ShadowFB"` to `"true"`, as it was initially intended and amend the commit message to reflect the change. With `ShadowFB` disabled, the UIVM graphics become extremely slow.
- Remove the `Screen` Section which is not required. As pointed by the logs, the `Depth` is set to 24 from the framebuffer information:
```
(II) FBDEV(0): Creating default Display subsection in Screen section
        "Default Screen Section" for depth/fbbpp 24/32
```
- Mention OXT-1320 in the commit message.

eric-ch:
v3:
- Disable DPMS again.

jean-edouard:
LGTM, merging soon.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1821

jandryuk:
@jean-edouard Can you give any details on the weird display behavior?

jean-edouard:
I could only repro it on one machine, but what happens is that killing X in UIVM turns off the laptop display completely and forever. But it happens both with and without this PR.

jandryuk:
Ok, thanks, @jean-edouard .  Do you know if surfman crashed?

jean-edouard:
surfman did not crash, and X restarted in UIVM, along with surf and all.
No error anywhere!

eric-ch:
@jean-edouard This points at the Xorg upgrade then?

jean-edouard:
lemme try a pre-upgrade build

jean-edouard:
Yeah, $1807 restarts the UI just fine...

jean-edouard:
Merging soon.

PR #1068, opened 2019-01-15
[RFC] initrdscripts: add openxt framework modules

dpsmith:
This PR is to serve as an RFC to allow others to review and test a refactoring of the initramfs init I need to do for a capability under development.

This PR removes the custom monolithic init script for the initramfs image and adopts the modular OE initramfs-framework init. While the intent was to minimize any modifications to upstream, initramfs-framework's init script did not directly support parsing a runlevel from the kernel parameters. As a result, this PR includes an extended version of the init that handles parsing the runlevel. An alternative approach that does not require altering the init script would be to write a module based on how the upstream init parameter parser works. The current parser would have handled the runlevel parameter by creating the variable $bootparam_{N}='true' where {N} would be the runlevel. The runlevel module would have to iterate through all permutations of $bootparam_{N} checking if it is set and setting the $bootparam_runlevel variable to the last {N} found set.

After review, unless otherwise reconsidered, I will close this PR so it will remain the record and do a new PR with that will incorporate all comments. Thanks.

### DIFF
--- a/recipes-core/initrdscripts/initramfs-framework/bootfs
+++ b/recipes-core/initrdscripts/initramfs-framework/bootfs
@@ -1,0 +1,59 @@
+#!/bin/sh
+# Copyright (C) 2018 Apertus Solutions, LLC
+# Licensed on MIT
+
+bootfs_enabled() {
+	if [ -n "${bootparam_boot}" ]; then
+		return 0
+	else
+		return 1
+	fi
+}
+
+bootfs_run() {
+	BOOTFS_DIR="/${ROOTFS_DIR}/boot/system"
+        if [ -z "$BOOTFS_DIR" ]; then
+                return
+        fi
+        C=0
+	# will use rootfs' delay and timeout values
+        delay=${bootparam_rootdelay:-1}
+        timeout=${bootparam_roottimeout:-5}
+        while [ ! -d $BOOTFS_DIR/grub ]; do
+                if [ $(( $C * $delay )) -gt $timeout ]; then
+                        fatal "boot '$bootparam_boot' doesn't exist or does not contain a /grub."
+                fi
+
+                if [ -n "$bootparam_boot" ]; then
+                        debug "No e2fs compatible filesystem has been mounted, mounting $bootparam_boot..."
+
+                        if [ "`echo ${bootparam_boot} | cut -c1-5`" = "UUID=" ]; then
+                                boot_uuid=`echo $bootparam_boot | cut -c6-`
+                                bootparam_boot="/dev/disk/by-uuid/$boot_uuid"
+                        fi
+
+                        if [ "`echo ${bootparam_boot} | cut -c1-9`" = "PARTUUID=" ]; then
+                                boot_uuid=`echo $bootparam_boot | cut -c10-`
+                                bootparam_boot="/dev/disk/by-partuuid/$boot_uuid"
+                        fi
+
+                        if [ -e "$bootparam_boot" ]; then
+                                flags=""
+                                if [ -n "$bootparam_bootfstype" ]; then
+                                        flags="$flags -t$bootparam_bootfstype"
+                                fi
+                                mount $flags $bootparam_boot $BOOTFS_DIR
+                                if [ -d $BOOTFS_DIR/grub ]; then
+                                        break
+                                else
+                                        # It is unlikely to change, but keep trying anyway.
+                                        # Perhaps we pick a different device next time.
+                                        umount $BOOTFS_DIR
+				fi
+			fi
+                fi
+                debug "Sleeping for $delay second(s) to wait root to settle..."
+                sleep $delay
+                C=$(( $C + 1 ))
+        done
+}
--- a/recipes-core/initrdscripts/initramfs-framework/init
+++ b/recipes-core/initrdscripts/initramfs-framework/init
@@ -1,0 +1,150 @@
+#!/bin/sh
+# Copyright (C) 2011 O.S. Systems Software LTDA.
+# Licensed on MIT
+#
+# Provides the API to be used by the initramfs modules
+#
+# Modules need to provide the following functions:
+#
+# <module>_enabled : check if the module ought to run (return 1 to skip)
+# <module>_run     : do what is need
+#
+# Boot parameters are available on environment in the as:
+#
+# 'foo=value' as 'bootparam_foo=value'
+# 'foo' as 'bootparam_foo=true'
+# 'foo.bar[=value] as 'foo_bar=[value|true]'
+
+# Register a function to be called before running a module
+# The hook is called as:
+#   <function> pre <module>
+add_module_pre_hook() {
+	MODULE_PRE_HOOKS="$MODULE_PRE_HOOKS $1"
+}
+
+# Register a function to be called after running a module
+# The hook is called as:
+#   <function> post <module>
+add_module_post_hook() {
+	MODULE_POST_HOOKS="$MODULE_POST_HOOKS $1"
+}
+
+# Load kernel module
+load_kernel_module() {
+	if modprobe $1 >/dev/null 2>&1; then
+		info "Loaded module $1"
+	else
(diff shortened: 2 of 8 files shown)


recipes-core/initrdscripts/initramfs-framework/init
```diff
@@ -0,0 +1,150 @@
...
+for p in `cat /proc/cmdline`; do
+	opt=`echo $p | cut -d'=' -f1`
+	opt=`echo $opt | tr '.-' '__'`
+	if [ "`echo $p | cut -d'=' -f1`" = "$p" ]; then
+		case $opt in
+		[0123456Ss])
+			eval "bootparam_runlevel=\"${opt}\"";;
+		*)
+			eval "bootparam_${opt}=true";;
+		esac
```
eric-ch: This is the only change from [upstream](https://github.com/openembedded/openembedded-core/blob/pyro/meta/recipes-core/initrdscripts/initramfs-framework/init)

It introduces the runlevel management.

jandryuk: Would it make sense to convert all the bootloader configs to runlevel=N and use the upstream script?

recipes-core/initrdscripts/initramfs-framework/bootfs
```diff
@@ -0,0 +1,59 @@
...
+# Copyright (C) 2018 Apertus Solutions, LLC
+# Licensed on MIT
+
+bootfs_enabled() {
+	if [ -n "${bootparam_boot}" ]; then
+		return 0
+	else
+		return 1
+	fi
+}
```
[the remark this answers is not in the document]
dpsmith: Yes, that would be concise but I agree with your second comment and we should do an info message for true and debug for failure.

recipes-core/initrdscripts/initramfs-framework/bootfs
```diff
@@ -0,0 +1,59 @@
...
+bootfs_run() {
+	BOOTFS_DIR="/${ROOTFS_DIR}/boot/system"
+        if [ -z "$BOOTFS_DIR" ]; then
+                return
+        fi
+        C=0
+	# will use rootfs' delay and timeout values
+        delay=${bootparam_rootdelay:-1}
+        timeout=${bootparam_roottimeout:-5}
+        while [ ! -d $BOOTFS_DIR/grub ]; do
```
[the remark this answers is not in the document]
dpsmith: Ack

recipes-core/initrdscripts/initramfs-framework/bootfs
```diff
@@ -0,0 +1,59 @@
...
+                if [ $(( $C * $delay )) -gt $timeout ]; then
+                        fatal "boot '$bootparam_boot' doesn't exist or does not contain a /grub."
+                fi
+
+                if [ -n "$bootparam_boot" ]; then
+                        debug "No e2fs compatible filesystem has been mounted, mounting $bootparam_boot..."
+
+                        if [ "`echo ${bootparam_boot} | cut -c1-5`" = "UUID=" ]; then
```
crogers1: I would recommend swapping the backticks for $() evaluation syntax here and elsewhere in the PR if possible.  I think most modern bash is preferring the POSIX-style nowadays.

recipes-core/initrdscripts/initramfs-framework/init
```diff
@@ -0,0 +1,150 @@
...
+	[ -n "$bootparam_verbose" ] && echo "$@" >/dev/console
+}
+
+# Prints information if debug bootparam is used
+debug() {
+	[ -n "$bootparam_debug" ] && echo "DEBUG: $@" >/dev/console
+}
+
+# Prints a message and start a endless loop
+fatal() {
```
crogers1: Would this cause the user to hard poweroff the box?

crogers1:
+1 from me as well, looks very well organized.

recipes-core/initrdscripts/initramfs-framework/bootfs
dpsmith: Yes, I generally use $() than ` but this is a duplication from upstreams rootfs module with ROOTFS replaced everywhere with BOOTFS. BTW, POSIX defines both as legitimate.[1](http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html)

recipes-core/initrdscripts/initramfs-framework/init
dpsmith: No, three finger salute will do a warm reboot.

recipes-core/initrdscripts/initramfs-framework/bootfs
```diff
@@ -0,0 +1,59 @@
...
+	if [ -n "${bootparam_boot}" ]; then
+		return 0
+	else
+		return 1
+	fi
+}
+
+bootfs_run() {
+	BOOTFS_DIR="/${ROOTFS_DIR}/boot/system"
+        if [ -z "$BOOTFS_DIR" ]; then
```
jean-edouard: Don't you mean `-d` instead of `-z`?
I don't think that string can be empty...

recipes-core/initrdscripts/initramfs-framework/bootfs
jean-edouard: Or without a subshell, pipe and binary call: `[ "${bootparam_boot:0:5}" = "UUID=" ]` seems to work fine with /bin/sh in OpenXT.
Maybe a bit picky, and maybe too bash-specific, but there are so many subshells with pipes and binary calls in this PR that it could be worth it to even switch to /bin/bash...

recipes-core/initrdscripts/initramfs-framework/tpm
```diff
@@ -0,0 +1,68 @@
...
+
+    load_kernel_module tpm_tis
+
+    local dev_path="$(find /sys/class -name tpm0)"
+    if [ -z "${dev_path}" ]; then
+        info "could not locate a TPM"
+        return 1
+    fi
+
+    if [ -e  "${dev_path}/device/caps" ]; then
```
jean-edouard: Double space :) #picky

recipes-core/initrdscripts/initramfs-framework/bootfs
dpsmith: So interesting point, the rootfs checks for existence because if it was set then in init the directory is created. In this case,  BOOTFS_DIR points at a directory within ROOTFS_DIR. Probably better to switch to directory check.

recipes-core/initrdscripts/initramfs-framework/bootfs
```diff
@@ -0,0 +1,59 @@
...
+                return
+        fi
+        C=0
+	# will use rootfs' delay and timeout values
```
jandryuk: Do we require bootfs to always be on the same device as rootfs?  If we require the same device, then the delay is unneeded since rootfs would already handle it.  However, I can understand leaving it to track the upstream rootfs script.

Also, here and throughout the PR, there is a mix of spaces and tabs.

eric-ch: @jandryuk The mix of spaces and tabs is coming from the upstream script.

dpsmith: I left the bootfs delay as it is in grub.cfg today. Do we want it to go away?

dpsmith:
Here is the outline of changes to be made,
- bootfs_enable: add info and debug message calls
- bootfs_run: rewrite to confirm to our standards, accepting this will cause divergence from upstream origin
- bootfs_run: rework validation for BOOTFS_DIR
- tpm_enable: cleanup double space
- init: drop custom runlevel parsing
- grub.cfg: change lines with runlevels to conform with init's parsing approach

There is one open question broached by @jandryuk, do we drop the timeout parameter? It is currently populated in grub.cfg today, but I can remove it while converting runlevel.

dpsmith:
Closing PR in preparation official PR.

PR #1069, opened 2019-01-16
xen: upgrade to 4.11

jean-edouard:
OXT-1454


jean-edouard:
The only difference with https://github.com/OpenXT/xenclient-oe/pull/1059 is in `xen-x86-Fix-up-rules-when-forcing-mno-sse.patch`, where `-mfpmath=387` was moved to a different line.

jean-edouard:
Depends on https://github.com/OpenXT/xsm-policy/pull/21

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1825

Unfortunately, this PR is incompatible with the latest Xorg version...
Please do not merge, at least for now.

jean-edouard:
This is the problem:
```
root@xenclient-uivm:~# modprobe xenfb2
modprobe: ERROR: could not insert 'xenfb2': Required key not available
```
No idea how this PR can mess with module keys...

jandryuk:
@jean-edouard I wonder if the kernel was rebuilt, but xenfb2 was not?  If you are using the auto-generated keys, then that would lead to a mismatch since xenfb2.ko would still be signed by the old key.  Can you check build logs or tmp-glibc/ to see if file times indicate this scenario?

jean-edouard:
@jandryuk thanks for the suggestion, but all custom builds are made entirely from scratch, without any cache, so I don't think that scenario is possible...

jean-edouard:
I kicked a rebuild, to rule out a race in the build/sign process, and to have a tree I can look at.

jean-edouard:
I kicked another build with the same branches, without any change, and that one boots just fine.
I guess it must be some kind of race. To be investigated if that ever happens again...
Please consider this PR merge-able again.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1828

crogers1:
After some conversation, stable-9 will need 4.12.  Since 4.12 is still in the release process and we shouldn't move that quite yet, 4.11 is a good interim version to park ourselves on until 4.12 drops.  After a rebase I'm +1 on this PR.

jean-edouard:
Fixed conflicts and updated to the tip of the 4.11 branch.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1872

crogers1:
Tested build on Dell 9010, Legacy boot, installed windows HVM guest with stubdom, encountered no errors.

+1

PR #1070, opened 2019-01-18
openxt_image_types: Fix .vhd size calculation

jandryuk:
The VHD size calculation was mistakenly using the wrong units.  This
meant that 2GB VHDs were being generated for our images.

ROOTFS_SIZE is in KB, so ALIGN should be 2 * 1024 for 2MB in KB.  The
TARGET_VHD_SIZE_KB was mislabeled since it is really MB which is the
unit that `vhd convert` expects.  Rename the two TARGET_VHD_SIZE*
variables so the units are clear.

Finally, correct the alignment calculation.  We want to add $ALIGN - 1
to force ROOTFS_SIZE up to the next chunk for any partially chunks.
Adding $ALIGN + 1 always forces up one chunk and potentially two.

A .disk partitioned disk image will be larger that ROOTFS_SIZE
kilobytes.  Now that we are using a more-precise .vhd size, we need to
take into account the extra size of the .disk.

We check the read size against ROOTFS_SIZE and adjust accordingly.
Maybe we should just always use the READ_SIZE_KB?

### DIFF
--- a/classes/openxt_image_types.bbclass
+++ b/classes/openxt_image_types.bbclass
@@ -14,14 +14,23 @@ IMAGE_CMD_ext3_append() {
 
 # OpenXT vhd.
 # Standard vhd image from an existing filesystem.
-# vhd size must be a multiple of 2 MB.
 CONVERSIONTYPES_append = " vhd"
 
 CONVERSION_CMD_vhd() {
-    local ALIGN=`expr 2 \* 1024 \* 1024`
-    local TARGET_VHD_SIZE=`expr \( \( ${ROOTFS_SIZE} + ${ALIGN} + 1 \) / ${ALIGN} \) \* ${ALIGN}`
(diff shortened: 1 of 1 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1830

jandryuk:
@jean-edouard this is the alternate which always uses the read size - https://github.com/jandryuk/xenclient-oe/commit/867c8eac864ba1499f79f96351ab6c68a045e9f7
You want that one force pushed here or in a new PR?

jean-edouard:
@jandryuk I'd say make a new PR, that way people can compare the two options.

jandryuk:
Closing this PR - use https://github.com/OpenXT/xenclient-oe/pull/1072 instead.

PR #1072, opened 2019-01-22
openxt_image_types: Fix .vhd size calculation

jandryuk:
The VHD size calculation was mistakenly using the wrong units.  This
meant that 2GB VHDs were being generated for our images.

ROOTFS_SIZE is in KB, so ALIGN should be 2 * 1024 for 2MB in KB.  The
TARGET_VHD_SIZE_KB was mislabeled since it is really MB which is the
unit that `vhd convert` expects.  Rename the two TARGET_VHD_SIZE*
variables so the units are clear.

Finally, correct the alignment calculation.  We want to add $ALIGN - 1
to force ROOTFS_SIZE up to the next chunk for any partially chunks.
Adding $ALIGN + 1 always forces up one chunk and potentially two.

A .disk partitioned disk image will be larger that ROOTFS_SIZE
kilobytes.  Too handle that, always read the image size of the image in
case it is a .disk.vhd.

Alternative to https://github.com/OpenXT/xenclient-oe/pull/1070

### DIFF
--- a/classes/openxt_image_types.bbclass
+++ b/classes/openxt_image_types.bbclass
@@ -14,14 +14,16 @@ IMAGE_CMD_ext3_append() {
 
 # OpenXT vhd.
 # Standard vhd image from an existing filesystem.
-# vhd size must be a multiple of 2 MB.
 CONVERSIONTYPES_append = " vhd"
 
 CONVERSION_CMD_vhd() {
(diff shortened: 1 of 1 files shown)


eric-ch:
Building: [$1843](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1843)

eric-ch:
Will merge soon.

PR #1073, opened 2019-01-22
OXT-1482 rpc-broker

tijko:
Bringing in rpc-broker is another step towards a Haskell free toolstack in OpenXT.  Rpc-broker will be a stand-in replacement for the current rpc filtering mechanism, rpc-proxy, which was written in Haskell and is no longer actively maintained.

Rpc-broker uses actively maintained libraries (e.g. libwebsockets, libdbus, libxml, libjson-c, etc..) while rpc-proxy was written years ago and rolled out its own Haskell modules in place of these well established and tested projects. 

Currently rpc-broker has travis-ci integration tests, rpc-proxy has no testing integration, or test-suites for that matter (has it ever been rigorously tested?).  In the process of writing the code to replace rpc-proxy, I came across numerous parts of the program that had errors or were outright broken.  For instance, when handling any type of complex data structures from dbus, any nesting whatsoever, results in malformed responses.

Rpc-broker also will replace the NDVM rpc filtering program, dbusbouncer.  While not written in Haskell, dbusbouncer is also very broken.


recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/services/xenpmd.te
```diff
@@ -50,7 +50,7 @@ dbus_connect_system_bus(xenpmd_t)
 dbus_send_system_bus(xenpmd_t)
 uid_dbus_chat(xenpmd_t)
 surfman_dbus_chat(xenpmd_t)
-rpcproxy_websockets_dbus_chat(xenpmd_t)
+rpcbroker_dbus_chat(xenpmd_t)
```
jandryuk: Why does this change from the websockets variant to the plain one?  Your SELinux policy includes rpcbroker_websockets_dbus_chat.

recipes-openxt/xctools/ndvm-rpc-broker-config/rpc-broker-ndvm.rules
```diff
@@ -0,0 +1 @@
+allow all
```
jandryuk: Could this at least be written to only allow dom0?

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/services/xenpmd.te
jean-edouard: Yeah this is confusing... Do you know what the best practice is for allowing 2 domains to "dbus_chat" with each other? Should they both "call" the interface of the other one? Or maybe only the domain that will be the caller should "call" the necessary interface?

recipes-openxt/xctools/dom0-rpc-broker-config.bb
```diff
@@ -0,0 +1,16 @@
+DESCRIPTION = "Dom0 policy file for rpc-broker"
+
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+SRC_URI = " \
+    file://rpc-broker-dom0.rules\
+"
+
```
jandryuk: You can `inherit allarch` since this config file doesn't depend on the architecture.  Naming it rpc-broker-config-dom0 may nicer since then the two recipes will be together in the listing.

recipes-openxt/xctools/ndvm-rpc-broker-config.bb
```diff
@@ -0,0 +1,16 @@
+DESCRIPTION = "NDVM policy file for rpc-broker"
+
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+SRC_URI = " \
+    file://rpc-broker-ndvm.rules\
+"
```
jandryuk: Same comments as the dom0 config.

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/services/rpcbroker.te
```diff
@@ -0,0 +1,99 @@
...
+surfman_dbus_chat(rpcbroker_websockets_t)
+
+# access logging
+logging_send_syslog_msg(rpcbroker_t)
+logging_send_syslog_msg(rpcbroker_websockets_t)
+
+# websockets ssl privileges 
+corecmd_exec_bin(rpcbroker_websockets_t)
+
+# websockets ssl certificates
```
jandryuk: Does rpc-broker actually use SSL on its websockets?

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/services/rpcbroker.te
```diff
@@ -0,0 +1,99 @@
...
+xen_dbus_chat(rpcbroker_websockets_t)
+dbd_dbus_chat(rpcbroker_websockets_t)
+surfman_dbus_chat(rpcbroker_websockets_t)
+
+# access logging
+logging_send_syslog_msg(rpcbroker_t)
+logging_send_syslog_msg(rpcbroker_websockets_t)
+
+# websockets ssl privileges 
+corecmd_exec_bin(rpcbroker_websockets_t)
```
jandryuk: rpcproxy.te has this entry with a comment explaining it is to run openssl.  Is that still applicable to rpc-broker?

recipes-connectivity/libwebsockets/libwebsockets_3.0.0.bb
```diff
@@ -0,0 +1,26 @@
...
+LIC_FILES_CHKSUM = "file://LICENSE;md5=4ce87f3facb6f911c142c8bef9bfb380"
+
+DEPENDS = "zlib"
+
+SRC_URI = "https://github.com/warmcat/libwebsockets/archive/v${PV}.tar.gz"
+SRC_URI[md5sum] = "2a78c8c49bbbfe67c24637398d989958"
+SRC_URI[sha256sum] = "a6b611c212c52f161f70556339fdaa199b7e9b6a167c4638e086d19db75d6290"
+inherit cmake pkgconfig
+
+PACKAGECONFIG ?= "libuv client server http2 ssl"
```
jandryuk: Maybe if you drop 'ssl', then the openssl stuff would go away?  'http2' and maybe 'client' could also be dropped.

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/services/rpcbroker.te
```diff
@@ -0,0 +1,99 @@
...
+xc_files_rw_v4v_chr(rpcbroker_websockets_t)
+
+# Allow xenstore in dom0
+xen_stream_connect_xenstore(rpcbroker_t)
+xen_stream_connect_xenstore(rpcbroker_websockets_t)
+# Allow xenstore outside of dom0
+dev_rw_xen(rpcbroker_t)
+dev_rw_xen(rpcbroker_websockets_t)
+
+dev_write_urand(rpcbroker_t)
```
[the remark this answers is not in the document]
tijko: I believe this is something we had to allow because of how libwebsockets initializes its connections.

https://libwebsockets.org/git/libwebsockets/tree/lib/roles/ws/client-ws.c?h=v3.1-stable#n117

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/services/xenpmd.te
tijko: Got it thanks! I will centralize everything in the rpc-broker policy module.

recipes-openxt/xctools/dom0-rpc-broker-config.bb
tijko: Will do.

recipes-openxt/xctools/rpc-broker/rpc-broker.initscript
```diff
@@ -0,0 +1,63 @@
...
+#
+PIDFILE=/var/run/rpc-broker.pid
+PIDFILE_WEBSOCKETS=/var/run/rpc-broker-websockets.pid
+
+# Make sure the progam exists
+[ -f /usr/sbin/rpc-broker ] || exit 0
+
+start() {
+	echo "Starting rpc-broker"    
+	export LD_PRELOAD=/usr/lib/libv4v-1.0.so.0
```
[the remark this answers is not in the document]
tijko: The reasoning behind this was, we were to either write an extensive patch to convert all libwebsockets calls to v4v or to use LD_PRELOAD.  Also, using the interposer allowed for travis-ci testing.

PR #1076, opened 2019-01-24
openssh: Packaging fixes

jandryuk:
The first change makes openssh-sshd-tcp-init get packaged generically.  With it, any VM could potentially install it, and the openxt-installer will use the core2-64 openssh package.  dom0's openssh is still machine specific since it uses a machine override for the sshd_config file.

The second change just adds the RDEPENDS on libv4v to the sshd, ssh & scp sub-packages since they contain the v4v variants - sshv4v, scpv4v, & sshd-v4v.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1837

jean-edouard:
LGTM, merging soon.

PR #1077, opened 2019-01-25
OXT-1485: Attend build warnings on TSS recipes.

eric-ch:
Noticed on a recent build:
```
WARNING: tpm2-tss-2.0.0-r0 do_package: QA Issue: tpm2-tss: Files/directories were installed but not shipped in any package:
  /usr/share
  /usr/share/man
  /usr/share/man/man7
  /usr/share/man/man3
  /usr/share/man/man7/tss2-tcti-mssim.7
  /usr/share/man/man7/tss2-tcti-device.7
  /usr/share/man/man3/Tss2_Tcti_Mssim_Init.3
  /usr/share/man/man3/Tss2_Tcti_Device_Init.3
  /usr/lib/udev
  /usr/lib/udev/rules.d
  /usr/lib/udev/rules.d/tpm-udev.rules
```

After reading the recipes, it seems they could be simplified and brought more inline with Bitbake/OE:
- Use `tpm2-tss` `./bootstrap` script only to generate the source list for `automake`.
- Do not use `tpm2-tools` `./boostrap` script at all since it calls `autoconf` without OE configuration
- Aggregate `tpm2-tss` library packages under `tpm2-tss` since they are all installed together.
- Let OE package `${libdir}`, `${docdir}` and `${libdir}/udev`

### DIFF
--- a/recipes-core/images/xenclient-initramfs-image.bb
+++ b/recipes-core/images/xenclient-initramfs-image.bb
@@ -14,10 +14,8 @@ IMAGE_FSTYPES = "cpio.gz"
 IMAGE_INSTALL = " \
     busybox \
     lvm2 \
-    libtss2 \
-    libtctisocket \
-    libtctidevice \
     tpm-tools-sa \
+    tpm2-tss \
     tpm2-tools \
     initramfs-module-lvm \
     initramfs-module-bootfs \
--- a/recipes-core/packagegroups/packagegroup-xenclient-dom0.bb
+++ b/recipes-core/packagegroups/packagegroup-xenclient-dom0.bb
@@ -114,9 +114,7 @@ RDEPENDS_${PN} = " \
     linuxfb-surfman-plugin \
     xenmgr \
     xen-xenstore \
-    libtss2 \
-    libtctidevice \
-    libtctisocket \
+    tpm2-tss \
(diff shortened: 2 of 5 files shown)


recipes-openxt/tss/tpm2-tss/leave-autotools-to-oe.patch
```diff
@@ -0,0 +1,16 @@
+--- a/bootstrap
++++ b/bootstrap
+@@ -13,7 +13,6 @@ src_listvar () {
+ }
+ 
+ VARS_FILE=src_vars.mk
+-AUTORECONF=${AUTORECONF:-autoreconf}
```
jandryuk: Instead of patching the file, we could call `AUTORECONF=true ./bootstrap` from do_configure_prepend() to skip executing autoreconf.

eric-ch: Indeed! It is preferable.

eric-ch:
v2:
- Rebased on master.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1887

jean-edouard:
LGTM, merging tomorrow.

PR #1079, opened 2019-01-25
switch to upstream blktap3

mahantesh-ais:
Upstream blktap3 (xapi-project/blktap) recently pulled some of patches (encryption)
from OpenXT/blktap3, it's better to align with upstream rather than having to fork it.
But, upstream code doesn't work as is, some corrections need to be applied.
These corrections are applied in the form of patches.
Currently following the master branch of upstream until a stable/good release.

OXT-1474


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1840
Note: merging this PR implies retiring https://github.com/OpenXT/blktap3

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
```diff
@@ -0,0 +1,270 @@
...
+Index: blktap/include/blktap.h
+===================================================================
+--- blktap.orig/include/blktap.h
++++ blktap/include/blktap.h
+@@ -32,7 +32,7 @@
+ #define _TD_BLKTAP_H_
+ 
+ #define BLKTAP2_SYSFS_DIR              "/sys/class/blktap2"
+-#define BLKTAP2_CONTROL_NAME           "blktap/control"
++#define BLKTAP2_CONTROL_NAME           "blktap-control"
```
jandryuk: Could we add a udev rule to symlink /dev/blktap/control to /dev/blktap-control?

recipes-extended/xen/blktap3/fix-segfault-if-startup-fails.patch
```diff
@@ -0,0 +1,161 @@
...
++++ b/tapback/tapback.c
+@@ -174,9 +174,9 @@ tapback_backend_destroy(backend_t *backend)
+     close(backend->ctrl_sock);
+     unlink(backend->local.sun_path);
+ 
+-	list_del(&backend->entry);
++    list_del(&backend->entry);
+ 
+-	free(backend);
++    free(backend);
```
jandryuk: This is just whitespace, so can we drop this hunk?

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
```diff
@@ -0,0 +1,270 @@
...
++		return err;
+ 	*off = vhd_position(ctx) - sizeof(vhd_footer_t);
+ 	return 0;
+ }
+@@ -2751,7 +2751,7 @@ vhd_initialize_header_parent_name(vhd_co
+ 	 * MICROSOFT_COMPAT
+ 	 * big endian unicode here 
+ 	 */
+-	cd = iconv_open(UTF_16BE, "ASCII");
++	cd = iconv_open("UTF-16BE", "ASCII");
```
jandryuk: This doesn't seem necessary since it's defined in `include/libvhd.h`

recipes-extended/xen/blktap3/use-existing-control-file.patch
```diff
@@ -0,0 +1,59 @@
...
+diff --git a/include/blktap.h b/include/blktap.h
+index a78d7da..1fd0737 100644
+--- a/include/blktap.h
++++ b/include/blktap.h
+@@ -36,7 +36,7 @@
+ #define BLKTAP2_CONTROL_DIR            "/var/run/blktap-control"
+ #define BLKTAP2_CONTROL_SOCKET         "ctl"
+ #define BLKTAP2_DIRECTORY              "/dev/xen/blktap-2"
+-#define BLKTAP2_CONTROL_DEVICE         BLKTAP2_DIRECTORY"/control"
++#define BLKTAP2_CONTROL_DEVICE         "/dev/blktap-control"
```
jandryuk: Can we just use udev to create a symlink instead of patching the file?

recipes-extended/xen/blktap3/OXT-specific-errors-fix.patch
```diff
@@ -0,0 +1,86 @@
...
+ 	int         state;
+Index: blktap/control/Makefile.am
+===================================================================
+--- blktap.orig/control/Makefile.am
++++ blktap/control/Makefile.am
+@@ -9,6 +9,7 @@ AM_CPPFLAGS += -I$(top_srcdir)/include
+ AM_CPPFLAGS += -DTAPDISK_EXEC='"tapdisk"'
+ AM_CPPFLAGS += -DTAPDISK_EXECDIR='"$(libexecdir)"'
+ AM_CPPFLAGS += -DTAPDISK_BUILDDIR='"$(top_builddir)/drivers"'
++AM_CPPFLAGS += -DTAP_CTL_NO_DEFAULT_CGROUP_SLICE
```
jandryuk: Can we use bitbake to pass in this flag?

recipes-extended/xen/blktap3/add-missing-files-to-gitignore.patch
```diff
@@ -0,0 +1,71 @@
+################################################################################
+SHORT DESCRIPTION: 
+################################################################################
+gitignore: Add missing generated files.
```
jandryuk: This patch would be good to upstream, but do we need it in our patch queue?

recipes-extended/xen/blktap3/compiler-errors-fix.patch
```diff
@@ -0,0 +1,409 @@
+################################################################################
+SHORT DESCRIPTION: 
+################################################################################
+Resolve compiler errors.
```
jandryuk: This seems like multiple patches all merged together.  I think you should avoid the whitespace changes - that will cut down on the patch size.  Then maybe separate out the inline changes from the PRIu64 changes into there own patches.  Those should be easily upstreamable.  The remainder would then be pretty small.

recipes-extended/xen/blktap3/compiler-errors-fix.patch
```diff
@@ -0,0 +1,409 @@
...
+Index: blktap/drivers/tapdisk-utils.h
+===================================================================
+--- blktap.orig/drivers/tapdisk-utils.h
++++ blktap/drivers/tapdisk-utils.h
+@@ -100,6 +100,6 @@ shm_create(struct shm *shm);
+ int
+ shm_destroy(struct shm *shm);
+ 
+-inline long long timeval_to_us(struct timeval *tv);
++long long timeval_to_us(struct timeval *tv);
```
jandryuk: Do these want to be static inline instead of dropping inline?

recipes-extended/xen/blktap3/fix-error-checks.patch
```diff
@@ -0,0 +1,357 @@
...
+SHORT DESCRIPTION: 
+################################################################################
+Fix error checks.
+
+################################################################################
+LONG DESCRIPTION: 
+################################################################################
+Replace atoi() with strtol().
+atoi() has limited functionality when it comes to error checks on the input.
+strtol() provides broader error checks on the input.
```
jandryuk: While strtol is certainly better than atoi, do we need to worry about this?  Also, there is a lot of duplication, if you create a parse_int function you could centralize all or it.

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
```diff
@@ -0,0 +1,270 @@
...
++
+ 	fd = open(BLKTAP2_CONTROL_DEVICE, O_RDONLY);
+ 	if (fd == -1) {
+ 		EPRINTF("failed to open control device: %d\n", errno);
+@@ -57,5 +72,10 @@ tap_ctl_free(const int minor)
+ 	err = ioctl(fd, BLKTAP2_IOCTL_FREE_TAP, minor);
+ 	close(fd);
+ 
++	unlink(tapdev_name);
++	unlink(ring_name);
```
jandryuk: Are these `unlink()` calls necessary?  Shouldn't udev clean the device nodes after BLKTAP2_IOCTL_FREE_TAP above?

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
```diff
@@ -0,0 +1,270 @@
...
+Index: blktap/control/tap-ctl-ipc.c
+===================================================================
+--- blktap.orig/control/tap-ctl-ipc.c
++++ blktap/control/tap-ctl-ipc.c
+@@ -64,7 +64,7 @@ tap_ctl_read_raw(int fd, void *buf, size
+ 		if (ret == -1)
+ 			break;
+ 		else if (FD_ISSET(fd, &readfds)) {
+-			ret = read(fd, buf + offset, size - offset);
++			ret = read(fd, (char*)buf + offset, size - offset);
```
jandryuk: I thought you could add an offset to a void * pointer without an issue.  i.e. is this cast needed?

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
```diff
@@ -0,0 +1,270 @@
...
+ 
+ 		printf("block %s: ", conv(hex, blk));
+ 		printf("sec: %s: %d\n", conv(hex, sec), bit);
+@@ -639,8 +642,8 @@ vhd_test_batmap(vhd_context_t *vhd, uint
+ 
+ 	for (i = 0; i < count; i++) {
+ 		cur = block + i;
+-		fprintf(stderr, "batmap for block %s: %d\n", conv(hex, cur),
+-			vhd_batmap_test(vhd, &vhd->batmap, cur));
++		printf("batmap for block %s: %d\n", conv(hex, cur),
```
jandryuk: Why this change?

recipes-extended/xen/blktap3/remove-creation-of-unused-log-files.patch
```diff
@@ -0,0 +1,150 @@
...
+-	tapdisk_logfile_close(logfile);
+-
+-	if (!keep)
+-		tapdisk_logfile_unlink(logfile);
+-}
+-
+-static int
+-tlog_logfile_open(const char *name, int level)
+-{
+-	td_logfile_t *logfile = &tapdisk_log.logfile;
```
jandryuk: To stay closer to upstream, maybe make this code just return when name is NULL?  Something would need to be figured out for the other calls.

recipes-extended/xen/blktap3/OXT-specific-errors-fix.patch
```diff
@@ -0,0 +1,86 @@
...
+Index: blktap/include/tap-ctl.h
+===================================================================
+--- blktap.orig/include/tap-ctl.h
++++ blktap/include/tap-ctl.h
+@@ -85,7 +85,7 @@ int tap_ctl_connect_send_receive_ex(int
+ 				    struct timeval *timeout);
+ char *tap_ctl_socket_name(int id);
+ 
+-typedef struct {
++typedef struct tap_list{
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will change declaration sites to use tap_list_t

recipes-extended/xen/blktap3/OXT-specific-errors-fix.patch
```diff
@@ -0,0 +1,86 @@
...
+No
+
+################################################################################
+INTERNAL DEPENDENCIES 
+################################################################################
+
+################################################################################
+PATCHES 
+################################################################################
+Index: blktap/include/Makefile.am
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will do.

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
```diff
@@ -0,0 +1,270 @@
...
+-            err = 0;
+-            if (!xs_rm(device->backend->xs, XBT_NULL, _path)) {
+-                if (errno != ENOENT) {
+-                    err = errno;
+-                    WARN(device, "failed to remove %s: %s\n", path,
+-                            strerror(err));
+-                }
+-            }
+-		}
+-    } else {
```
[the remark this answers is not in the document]
mahantesh-ais: I think the code removed is dead code. Above this code there is a comment mentioning non-existence of front-end path is not possible as that is what triggers the function call.  Also, we search a device with appropriate front-end state path above. The function returns if device is NULL which could be because of non-existent front-end state path.
However, please let me know if you think this is not dead code. I'll remove this hunk from patch.

recipes-extended/xen/blktap3/remove-creation-of-unused-log-files.patch
```diff
@@ -0,0 +1,150 @@
...
+Mahantesh Salimath <salimathm@ainfosec.com>
+
+################################################################################
+REMOVAL 
+################################################################################
+
+################################################################################
+UPSTREAM PLAN
+################################################################################
+Yes, could be upstreamed.
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will update appropriately.

recipes-extended/xen/blktap3/add-device-string-support-to-tap-destroy.patch
```diff
@@ -0,0 +1,226 @@
...
++{
++	int pid, err;
++	struct list_head list = LIST_HEAD_INIT(list);
++	tap_list_t *entry;
++
++	err = tap_ctl_list(&list);
++	if (err)
++		return err;
++
++	pid = -1;
```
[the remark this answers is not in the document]
mahantesh-ais: Ack.

recipes-extended/xen/blktap3/compiler-errors-fix.patch
mahantesh-ais: This patch should be upstream'd completely, if you suggest splitting them, I'll do so. Thanks

recipes-extended/xen/blktap3/fix-error-checks.patch
mahantesh-ais: I had introduced this change as part of review fix from @eric-ch while updating OpenXT/blktap3. I can either remove this completely or create parse_int function as you suggested.
@eric-ch: do you have any suggestions ?

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
mahantesh-ais: I think so, last time I checked those were not removed. Will check again and update accordingly.

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
mahantesh-ais: I'm pretty sure I got a segfault for not using the cast.

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
mahantesh-ais: fprintf is causing an issue.

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
jandryuk: Oh, I didn't realize this was for the /proc/misc check - You are right - udev would not work for that.

That makes we wonder why we have a different value than xenserver.  Do we have a different version of the driver?

jandryuk:
@mahantesh-ais, also nice work on getting vhd encryption upstream.

recipes-extended/xen/blktap3/compiler-errors-fix.patch
jandryuk: Oh, sorry, I missed that these are only prototypes.

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
mahantesh-ais: I believe this is just a typo, but yes we'll confirm with xenserver

recipes-extended/xen/blktap3/use-existing-control-file.patch
mahantesh-ais: I tried creating symlink using udev rule. When I run udevadm test the logs show that rule should run fine but when I reboot the machine the udev rule seems to not have been triggered. This may be because it is not an actual device, but not sure. Also selinux policy needs to be updated in this case. 
However, I think this not much of concern, this hunk can be dropped without any issue.
Let me know what you think, keep it or drop it ?

recipes-extended/xen/blktap3/OXT-specific-errors-fix.patch
mahantesh-ais: Ack, will do

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
mahantesh-ais: I checked again and the nodes are not removed by default. BLKTAP2_IOCTL_FREE_TAP is used by kernel module only to free its resources.

recipes-extended/xen/blktap3/remove-creation-of-unused-log-files.patch
mahantesh-ais: These functions just create log files where nothing is actually logged. Removing the hunk will create those files and selinux policy needs to be updated. I prefer these files not be created. Let me know if you think otherwise, I can remove the hunk and update selinux policy accordingly.

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
mahantesh-ais: You're right, updating accordingly.

recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
jandryuk: Okay.  Thanks for checking.

recipes-extended/xen/blktap3/remove-creation-of-unused-log-files.patch
jandryuk: Oh, so it always tries to create the files.  Making it configurable would be better for upstreaming, but this is fine for openxt.

recipes-extended/xen/blktap3/use-existing-control-file.patch
jandryuk: Hmm, looks like the code creates BLKTAP2_CONTROL_DEVICE if it doesn't exist.  I like your idea of using /dev/blktap-control, but I would just drop the hunk to minimize the diff.

mahantesh-ais: Ack, will do.

mahantesh-ais:
> Generally I think this looks good. I'd like the two commits squashed together before it goes in.
> 
> Also, compile-errors-fix.patch still has a lot of whitespace changes. I think it's worth cleaning up the whitespace upstream, but I don't think it's needed for OpenXT. Generally I think improvements and cleanups should go upstream, and the xenclient-oe patch queue should only contain patches we need for OpenXT.

@jandryuk : Are you talking about fix-format-specifier-errors.patch and remove-inline-function-declarations.patch to be squashed together ? Just making sure.

The whitespace/indentation fixes in compile-errors-fix.patch are actually needed, it fails to compile without them. This patch is meant to go upstream anyways. 
Agreed, except for the following I'm planning to push every other patch to go upstream:
1. OXT-specific-errors-fix.patch
2. remove-creation-of-unused-log-files.patch
3. blktap3-vhd-icbinn-support.patch

jandryuk:
This PR has two commits to xenclient-oe.  Those should be squashed together.

The other patches are split up well.  Okay, I didn't realize the whitespace changes are needed.  Without them, we have -Werror=misleading-indentation.

mahantesh-ais:
> This PR has two commits to xenclient-oe. Those should be squashed together.
> 
> The other patches are split up well. Okay, I didn't realize the whitespace changes are needed. Without them, we have -Werror=misleading-indentation.

Ack, will do. Thanks

mahantesh-ais:
> Sorry, one last thing. If you create VERSION without a trailing new line `echo -n 3.7.0 > VERSION` it looks like the m4_include(VERSION) works.
> 
> ```
> +--- /dev/null
> ++++ b/VERSION
> +@@ -0,0 +1 @@
> ++3.7.0
> +\ No newline at end of file
> ```

Oh okay, didn't realize new line was the issue. Thanks, will take care of it.

mahantesh-ais:
Built here : http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1848

mahantesh-ais:
Using v3.20.0 as base (tagged on 01/15/2019) instead of master branch.

jean-edouard:
@jandryuk, if your concerns have been addressed, please approve this pull request.
Thank you

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1887

jean-edouard:
Merging tomorrow.

PR #1080, opened 2019-01-28
grub-efi: Don't create grubx64.efi symlink

jandryuk:
grubx64.efi is a convenience symlink to grub-efi-bootx64.efi.
Unfortunately, it can get set to grub-efi-bootia32.efi if
grub-efi-native:do_deploy runs after grub-efi.  This breaks the build
with
cp: cannot stat ‘tmp-glibc/deploy/images/openxt-installer/grubx64.efi’: No such file or directory

openxt.git build scripts will copy the actual file, so don't bother
creating a symlink.

OXT-1484

This should go in after https://github.com/OpenXT/openxt/pull/334


jean-edouard:

PR #1081, opened 2019-01-28
OXT-1481: tboot: upgrade to 1.9.9

eric-ch:
Port the OpenXT patch-queue to release 1.9.9 and amend affected measurement scripts.

As the title mentions, this is mostly RFC to ease the review and discuss the possible implications of https://github.com/OpenXT/xenclient-oe/commit/e9974fb24ff2f44f79069607b9d900c71f7f9163 on some platforms (should someone find one).

The current patch-queue is available as a git repository: https://github.com/eric-ch/tboot/tree/openxt-1.9.9.

Patches 0015, 0016, 0017 are intended to be folded into 0006 once reviewed.

Tested with Dell E7470 (TPM2.0), legacy and UEFI installation, forward-sealed from master to this version.
Tested with Dell E7450 (TPM1.2), legacy and UEFI installation, forward-sealed from master to this version.

Another thing, not related to this upgrade directly, may be considered as well. OpenXT 8 did not implement UEFI/TBoot measurement, making forward sealing from 8 to 9 a possible hassle. Currently the scripts do not support forward sealing a UEFI/SRTM (standard scheme in OpenXT 8) installation to a UEFI/SRTM+DRTM (TBoot) one nor allow to forward seal without using TBoot/TXT. This submission does not address this issue.


jandryuk:
Regarding the evtlog overrides, `acminfo` will print the tcg_event_log_format value from an SINIT.  If we know which ACM we are running, we can check the value and act appropriately.  Would that help?

On a stable-8 box, all the ACMs have `tcg_event_log_format: 0` except /boot/8th_gen_i5_i7-SINIT_76.bin which has `tcg_event_log_format: 1`

You left the tboot-1.9.6 files in place.  Is that intentional?

Xen has a patch upstream to re-enable NMI, so "tboot: Unmask NMI potentially masked during SENTER" can go away as long as we have xen 072e054359a4 "x86/boot: enable NMIs after traps init" or a backport when we go to xen 4.11+.

eric-ch:
> acminfo will print the tcg_event_log_format value [...] Would that help?

Maybe, the hash calculation is up to the SINIT, but it seems predictable. We may be able to use one or the other depending on the `acminfo` returned value.

> On a stable-8 box, all the ACMs have tcg_event_log_format: 0 except /boot/8th_gen_i5_i7-SINIT_76.bin which has tcg_event_log_format: 1

That would mean 8.0.1+ (post Tboot 1.9.6) would fail forward seal on Intel 8th gen. If someone has one to try that would be interesting. Looking at the tests published, I am not sure an Intel 8th gen was tested.

> You left the tboot-1.9.6 files in place. Is that intentional?

I was going to remove them with the patches folded to have less noise in the PR for initial review.

> Xen has a patch upstream to re-enable NMI

Right, I forgot to remove ours, it seems after reading the thread that re-enabling the NMIs in Xen is preferable.

eric-ch:
Looking at [OXT-1438](https://openxt.atlassian.net/browse/OXT-1438) again, it seems we switched back and forth between the `EVTYPE_OSSINITDATA_CAP_HASH` values.

@jandryuk it seems using the `acminfo` should be sufficient if we have both hashes already, at least until some other bit decides to flip.

jandryuk:
EVTYPE_OSSINITDATA_CAP_HASH is a hash of the OsSinitData capabilities dword
$ echo -en '\x02\x00\x00\x00' | sha256sum 
26b25d457597a7b0463f9620f666dd10aa2c4373a505967c7c8d70922a2d6ece  -
$ echo -en '\x02\x02\x00\x00' | sha256sum 
5a3e80a37915b1601c363acd1601df7ef257d5d32c664004a2ec0484a4f60628  -

So 0x2 and 0x202.

tboot has an initial MLE_HDR_CAP of 0x227.  In tboot/txt/txt.c:init_txt_heap() it looks at its configuration and the SINIT and comes up with a value.

I guess the "right" thing would be to refactor the tboot code for use by a standalone executable.  Give it an SINIT module and it'll spit out the future capabilities value.  We'd also need to determine the SINIT - acminfo will do it, but it uses /dev/mem :(

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1870

crogers1:
Tested on Dell 7060, installed Measured Launch, successfully booted into dom0 with good MLE.  Previously saw the tboot boot loop with 1.9.6 on this platform.

eric-ch:
v2:
- Remove RFC tag, this is now tracked under separate Jira tickets (OXT-1506, OXT-1507)
- Remove patch unmasking NMI (OXT-1446), as this is now done in Xen, which is considered preferable
- Fold patches for `pcr-calc` separated initially for review convenience.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1878

jean-edouard:
LGTM, merging soon.

PR #1082, opened 2019-01-28
linux: Fix domU build of usbback

jandryuk:
We get an error in evtchn.h because domid_t is not defined at the time
of inclusion.  include/xen/xen.h automatically includes
xen/interface/xen.h when CONFIG_DOM0 is defined.  That doesn't happen
when CONFIG_DOM0 is not set leading to the domid_t error.  Rearranging
the headers allows domid_t to be already defined when evtchn.h is
included.  interfaces.c also needs an additional include.


jean-edouard:

eric-ch:
Will merge soon.

PR #1083, opened 2019-01-29
init.root-ro: Remove restorecon /storage

jandryuk:
mountall.sh already runs a restorecon on /storage, so we don't need to
do it in this init script.

An encrypted /storage will not be mounted during init.root-ro, so we
want restorecon delayed until /storage is actually available.


jean-edouard:

jandryuk:
Force pushed to handle the rename.

jean-edouard:
Thank you! Merging soon.

PR #1085, opened 2019-02-11
rsyslog avoid duplicate "kernel:" lines and config consolidation

jandryuk:
This PR makes a few different changes to the rsyslog configuration and packaging.

First it corrects the path for the logrotate.rsyslog file removal.

Next, it re-organizes the machine overrides so that we have a single default rsyslog.conf for all VMs and overrides for dom0 and the installer.  The openxt-installer directory is a symlink to xenclient-dom0 so that they are kept in sync - is that okay?

Finally, it drops imklog from the default VM rsyslog.conf.  This eliminates the duplicate "kernel:" lines that rsyslogs prints for all kernel messages.  Only the kernel logging to /dev/hvc0 will make it through.


recipes-extended/rsyslog/files/openxt-installer
```diff
@@ -0,0 +1 @@
+xenclient-dom0
```
eric-ch: I'm fine with that for now, but I would prefer to move toward recipes avoiding to be machine specific for configuration purposes. This is actually tied closer to the image recipe than the targeted machine.

I know of two ways to do so in that kind of case:
- append `rsyslog` recipe to use `update-alternatives.bbclass`; create new recipes with `ALTERNATIVE_TARGET` and `ALTERNATIVE_PRIORITY` providing specific configurations at higher priorities.
- append `rsyslog` recipe and remove the configuration file; create a new recipes providing default and specific configuration files.

Both avoid rebuilding `rsyslog` for each machine just to *maybe* change a configuration file.
The first one has a default configuration with the standalone recipe, but deploys potentially 2 configuration files on target when using a specific configuration.
The second one does not provide a "default" configuration when the recipe is used alone, but deploys only the intended configuration.

eric-ch:
Building: [$1856](http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1856)

recipes-extended/rsyslog/files/openxt-installer
jandryuk: Yes, I never liked building MACHINE-specific packages that just differ in their config files.   But I haven't seen a good OE way to deal with different config files.

I want something like OE packaging CONFFILES into ${PN}-conf with RDEPENDS_${PN}=${PN}-conf.  Recipe ${PN}-conf-dom0 would RPROVIDE ${PN}-conf and those would supersede the default ${PN}-conf package.  This doesn't work.

However, using RRECOMMENDS_${PN}=${PN}-conf might do what I want?

eric-ch: - `RRECOMMENDS` would require setting `NO_RECOMMENDATIONS` so the recommended packages are not installed by default, then adding even recommended package to `IMAGE_INSTALL` or...
- `BAD_RECOMMENDATIONS` to list the "other" recommended packages exhaustively.

I would pick the second option I think. Then `bbappend` the recipe to remove the default configuration files and provide them in a different recipe if the package built does not allow for configuration extension.

I *should* work?

dpsmith: Actually, I think there is a simpler approach. Have a core configuration which is in rsyslog.conf with a final include directive for the /etc/rsyslog.d directory. Then have a set of two digit prefixed (to enable ordered loading) config partials with each partial individually package and have any necessary RDEPENDS linkages between config partials. Then in the image you just include the one or more partials desired to be included.

jandryuk: @eric-ch `RRECOMMENDS` seemed to work without `NO_RECOMMENDATIONS` or `BAD_RECOMMENDATIONS`.  In my brief test, new recipe `rsyslog-conf-dom0` RPROVIDES `rsyslog-conf` and `rsyslog.bb` packages `rsyslog-conf`.  Adding `rsyslog-conf-dom0` to the packagegroup installs that config.  Omitting `rsyslog-conf-dom0`, the default `rsyslog.conf` is installed.  Maybe because `rsyslog-conf` is already provided by `rsyslog-conf-dom0`, it doesn't have to fall back to looking for recommends?

I tried setting `CONFFILES_${PN}-conf = "${CONFFILES}_${PN}"` but it only expanded ${PN} so the package data was incorrect for `rsyslog-conf`.

@dpsmith Yeah, your idea would work here.  It's a little bit of a shame that we'd have to change our config setup just to work around OE.  There may be other places (`sshd_config`?) where we don't have an option to side-step the issue.

dpsmith: The problem is not OE, its due to the incorrect way we use OE's constructs. We are attempting to control distro level configuration with machine based overrides. The correct way would be to use distro_features to pick the config file (or fragment). Then if you want to break the standard distro config, you can just override the variable in your local.conf.

jandryuk: It's not a distro or machine thing.  We have N images that want N different config files.  Since OE packages the binaries and the config files together, we don't have a good OE way to replace just the config files.  In essence, we use machine overrides as image overrides.

dpsmith: Then we can still do the initial suggestion. For the core recipe, package the config file in a different conf package. Then have separate recipes that generate alternative configuration packages and select the respective config package needed by the image.

eric-ch: @dpsmith I wish `/etc/${BPN}.d` was more common, that would solve the problem entirely.

@jandryuk Uh! That is in contradiction with [`RRECOMMENDS`](https://www.yoctoproject.org/docs/2.7/ref-manual/ref-manual.html#var-RRECOMMENDS) documentation as far as I understand it. At least I would expect the `do_rootfs` to moan about overwriting the configuration file from one package with the other. How does the package manage knows which `RRECOMMENDS` recipe to install? Is it the first one provided in `IMAGE_INSTALL` (or through `packagegroup` so `RDEPENDS`)?

jandryuk: rsyslog.bb:
```
PACKAGES = "rsyslog rsyslog-conf"
RECOMMENDS_${PN} = "${PN}-conf"
```

rsyslog-conf-dom0.bb:
```
RPROVIDES_${PN} = "rsyslog-conf"
```

When packagegroup has `IMAGE_INSTALL = "rsyslog rsyslog-conf-dom0"`, it installs rsyslog and rsyslog-conf-dom0, skipping rsyslog-conf.
When packagegroup has `IMAGE_INSTALL = "rsyslog"`, it installs rsyslog and rsyslog-conf.

So I think it only grabs rsyslog-conf since nothing in the installation command provides it.  i.e. it doesn't search for a recommended package since the rsyslog-conf-dom0 RPROVIDES fulfills the dependency.

When I tried initially with RDEPENDS (Basically the same as above but replacing RRECOMMENDS), it failed to install.   opkg doesn't give much useful output, but I guess it's the strictness of RDEPENDS and the conflicting files.

jandryuk:
I pushed two more commits to show the RRECOMMENDS implementation.

I wanted `CONFFILES_${PN}-conf += "${CONFFILES}_${PN}"` to work, but it does not.  Instead I hardcoded `CONFFILES_${PN}-conf += "$(sysconfdir}/rsyslog.conf"`

eric-ch:

eric-ch:
Will merge soon.

PR #1086, opened 2019-02-14
OXT-1499: QEMU 3.1

jandryuk:
This is an uprev of QEMU to version 3.1.  It's pretty straight forward over the previous version.

I don't have ATAPI hardware, so I have not tested ATAPI passthrough.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1865

jean-edouard:
LGTM, merging soon.

PR #1087, opened 2019-02-15
rpc-proxy: initscript: start rpc-proxy even when SELinux is disabled.

jean-edouard:
2 of the 3 rpc-proxy instances are started with runcon, since we want
each process to be in its own SELinux context.
However, runcon fails when SELinux is disabled, and rpc-proxy is not
started in that case.
Fixing this by not running runcon when SELinux is disabled.

OXT-1504


eric-ch:

eric-ch:
Looking at the ticket, there are other possible approaches that can be addressed later (e.g, multiple initscripts to transition without using `runcon`).

Will merge on Monday if no objection is raised.

PR #1090, opened 2019-02-19
Remove iproute2 customization

jandryuk:
The iproute2 .bbappend makes the package machine-specific as it modifies
/etc/iproute2/rt_tables for NDVM and Dom0.  But customization is unnecessary.

NDVM mounts a tmpfs over /etc/iproute2 and network-slave populates
rt_tables, so the customized version is never seen.

Dom0 doesn't do any routing, so there is no reason to modify the
rt_tables file.

Since iproute2 is no longer machine-specific, the xen dependency hack can be removed.


eric-ch:

PR #1091, opened 2019-02-19
udev: Delete remnant file

jandryuk:
This file is unused.  There is no .bbappend to make it apply... and we
use eudev now anyway.

I think this is leftover from the introduction of the 64bit installer
after the pyro uprev.


eric-ch:

PR #1092, opened 2019-02-20
OXT-1514, OXT-1515, OXT-1516, OXT-1517: xen/xl: Misc fixes and refactoring in xl/libxl.

eric-ch:
* xen/xl: Fix memory fault in libxl create domain.
Triggered on 64bit dom0. Refactor `libxl-avoid-creating-unusable-cdrom-vbd-xs-nodes.patch` while at it.

* xen/xl: Refactor `tapdev_is_shared` callsites.
Multiple patches seem to have accumulated some churn, try to avoid multiple calls to `tapdev_is_shared`. It seems that it should be called in the blktap{2,3} implementations.

* xen/xl: Add pid to syslog() redirections for xl.
Amend the patches redirecting logs to `syslog()` to catch more logs from `xl`. Also add the pid of the process prepended to log lines. It can be useful to identify the daemon process.

* xen/xl: Refactor xenmgr sync for firewall rules.
Reduce the timeout spent waiting for `xenmgr` to notify `xl` through Xenstore that the V4V firewall rules are applied before starting the stubdomain. The current timeout will trigger the dm timeout. Also be more conservative with garbage collected allocations by removing them from the active loop.

### DIFF
--- a/recipes-extended/xen/files/libxl-xenmgr-support.patch
+++ b/recipes-extended/xen/files/libxl-xenmgr-support.patch
@@ -170,22 +170,26 @@ PATCHES
  out:
 --- a/tools/libxl/libxl_dm.c
 +++ b/tools/libxl/libxl_dm.c
-@@ -2123,6 +2123,20 @@ retry_transaction:
+@@ -2123,6 +2123,24 @@ retry_transaction:
      libxl__xs_printf(gc, XBT_NULL,
                       DEVICE_MODEL_XS_PATH(gc, dm_domid, guest_domid, "/xen_extended_power_mgmt"), "2");
  
 +    /* OpenXT: We wait on xenmgr writing v4v-firewall-ready.
 +     *         xenmgr waits on us writing image/device-model-domid
 +     */
-+    int32_t timeout = 0;
-+    char * ready = NULL;
++    unsigned int timeout;
++    char *ready = NULL;
++    const char *firewall_path =
++        GCSPRINTF("%s/v4v-firewall-ready",
++                  libxl__xs_get_dompath(gc, guest_domid));
 +    /* Block and wait for v4v firewall rules */
-+    while (timeout < 30) {
-+        ready = libxl__xs_read(gc, XBT_NULL, GCSPRINTF("%s/v4v-firewall-ready", libxl__xs_get_dompath(gc, guest_domid)));
-+        if(ready)
++    for (timeout = 0; !ready && timeout < 4; ++timeout) {
++        ready = xs_read(ctx->xsh, XBT_NULL, firewall_path, NULL);
++        if (ready) {
++            free(ready);
 +            break;
++        }
 +        sleep(1);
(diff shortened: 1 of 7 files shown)


recipes-extended/xen/files/libxl-xenmgr-support.patch
```diff
@@ -170,22 +170,26 @@ PATCHES
...
++    unsigned int timeout;
++    char *ready = NULL;
++    const char *firewall_path =
++        GCSPRINTF("%s/v4v-firewall-ready",
++                  libxl__xs_get_dompath(gc, guest_domid));
 +    /* Block and wait for v4v firewall rules */
-+    while (timeout < 30) {
-+        ready = libxl__xs_read(gc, XBT_NULL, GCSPRINTF("%s/v4v-firewall-ready", libxl__xs_get_dompath(gc, guest_domid)));
-+        if(ready)
++    for (timeout = 0; !ready && timeout < 4; ++timeout) {
```
jandryuk: I agree 30 seconds is excessive, and I agree the stubdom should boot in 4 seconds or less.  But do we want a little more margin just in case?  Or do we normally boot in 2 seconds (guessing) so 4 seconds is plenty?

eric-ch: I picked 4s because it is usually, as you guessed, up around 2s on the machine I tested. The stubdom timeout is defined to 30s and based on an `xs_watch`, we could possibly index on that?

This is a little on the exotic side of things I guess. Maybe it could also be refactored in an AO along the stubdom creation path, but that should be tackled in a different PR.

eric-ch:
@jandryuk `libxl-cleanup-string-to-backend.patch` is only a cleanup. It can go upstream I would think, although we may want to send it with other cleanups we may have (assuming there are others in our patch-queue)?

eric-ch:
v2:
- Fix typo in `strcmp(d_config.disks[i].vdev, "atapi-pt")` return value check.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1870

stacktrust:
If all of these changes will be merged/reverted as a group, can they  have separate JIRA tickets, referenced in the PR?  

Ticket exists for 64-bit dom0.  Patches for upstream can be tracked in JIRA (components Upstream + toolstack).  If blktap requires testing, that should be a separate ticket.  Should the blktap changes go upstream?  Any of these needed in stable 8?  Are they all the same priority?

eric-ch:
v3:
- Add tickets numbers,
- Remove `libxl-cleanup-string-to-backend.patch` for eventual submission upstream.

eric-ch:
v4:
- Rebase on master (Xen 4.11 pq).

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1888

LGTM, merging soon.

PR #1093, opened 2019-02-27
OXT-1509: Xterm resizable TrueType fonts

jandryuk:
Font resizing was broken during the pyro upgrade.  xterm was no longer
built with TrueType font support and there were no X fonts installed.
It seems like only the default X fallback font was available to xterm.
Enable TrueType fonts through xft in PACKAGECONFIG to re-enable terminal
font size changing.

I tried installing xorg-minimal-fonts, but that was actively harmful.
With that package installed, any attempt to change the font through
ctrl-right click led to xterm exiting through an abort().

The default font size is set to 10 point.

### DIFF
--- a/recipes-graphics/xorg-app/xterm_%.bbappend
+++ b/recipes-graphics/xorg-app/xterm_%.bbappend
@@ -1,3 +1,5 @@
+# Enable TrueType fonts
+PACKAGECONFIG += "xft"
 
 do_configure_prepend() {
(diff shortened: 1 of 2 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1875

eric-ch:
Will merge soon.

PR #1095, opened 2019-03-05
Xen 4.11 Read-only IDE fix

jandryuk:
This PR fixes libxl-support-hvm-readonly-disks.patch after the 4.11 uprev for read-only IDE support.  You can see this with HVM NDVM.

It also cleans up the patch to drop the changes to COLO since we don't use or support that functionality.


jean-edouard:
Oops, my bad, thanks for the fix. Building soon.

jandryuk:
No problem.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1885

jean-edouard:
LGTM, merging soon.

PR #1096, opened 2019-03-07
ALSA AVCs and cleanup

jandryuk:
This PR fixes two audio_helper AVCs.

It also removes the ALSA bbappend and converts to using /etc/asound.conf as a volatiles symlink instead of the /etc/asound directory symlink.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1894

jean-edouard:
LGTM, merging soon.

PR #1097, opened 2019-03-07
module-signing: Handle no module case

jandryuk:
With an external signing key, do_sign_modules will fail when there are
no modules.
| DEBUG: Executing shell function do_sign_modules
| Usage: scripts/sign-file [-dp] <hash algo> <key> <x509> <module> [<dest>]
|        scripts/sign-file -s <raw sig> <hash algo> <x509> <module> [<dest>]

If there are no modules, we don't have to run sign-file.  Have xargs
skip running the command in that case.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1894

jean-edouard:
LGTM, merging soon.

PR #1098, opened 2019-03-08
ahci support: provide multiple disk support in case of stubdom

mahantesh-ais:
OXT-1123

By  emulating qemu SATA (hdtype='ahci') disk, it is possible to
add upto 6 disks to a hvm guest (guest should support AHCI).

### DIFF
--- a/recipes-extended/xen/files/libxl-RFC-fix-multiple-disks.patch
+++ b/recipes-extended/xen/files/libxl-RFC-fix-multiple-disks.patch
@@ -41,7 +41,22 @@ PATCHES
                  target_path = "/dev/xvdc";
              } else {
                  if (format == NULL) {
-@@ -1605,8 +1605,10 @@ static int libxl__build_device_model_arg
+@@ -1592,6 +1592,14 @@ static int libxl__build_device_model_arg
+                              "qemu-xen doesn't support read-only AHCI disk drivers");
+                         return ERROR_INVAL;
+                     }
++                    if (b_info->stubdomain_version == LIBXL_STUBDOMAIN_VERSION_LINUX) {
++                        target_path = (char *[]) {"/dev/xvda",
++                                                  "/dev/xvdb",
++                                                  "/dev/xvdc",
++                                                  "/dev/xvdd",
++                                                  "/dev/xvde",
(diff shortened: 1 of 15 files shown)


mahantesh-ais:
To be merged along with:
https://github.com/OpenXT/manager/pull/133
https://github.com/OpenXT/idl/pull/32

mahantesh-ais:
Built here:
http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1896

recipes-extended/xen/files/libxl-RFC-fix-multiple-disks.patch
```diff
@@ -41,7 +41,22 @@ PATCHES
                  target_path = "/dev/xvdc";
              } else {
                  if (format == NULL) {
-@@ -1605,8 +1605,10 @@ static int libxl__build_device_model_arg
+@@ -1592,6 +1592,14 @@ static int libxl__build_device_model_arg
+                              "qemu-xen doesn't support read-only AHCI disk drivers");
+                         return ERROR_INVAL;
+                     }
++                    if (b_info->stubdomain_version == LIBXL_STUBDOMAIN_VERSION_LINUX) {
++                        target_path = (char *[]) {"/dev/xvda",
```
jandryuk: This is a 6 disk duplication of the IDE case below.  Maybe we should use something like the proposed upstream linux-stubdom support:
https://github.com/marmarek/xen/commit/060e89a5cca415871c19d05853a31e7d7bc7f6cc#diff-b43617695b9412434577d745e4f3102dR1820

That's a single setting of target_path for all disks.  That allows 26 disks, but you must need emulated scsi to use more than 6 of them.

Another option would be to move your 6 disk code earlier and let it apply to both AHCI or IDE.

mahantesh-ais: I thought of this, but I didn't wanted to mix IDE and AHCI code path and trigger any unknown issues in IDE path.

jandryuk: Okay.

jean-edouard:
Merging soon.

PR #1100, opened 2019-03-11
OXT-1506: Calculate event 0x40f in pcr-calc taking ACM capabilities into account

eric-ch:
Patches against TBoot are to be folded into `0006-pcr-calc-Add-pcr-calculator-tool.patch`. This is to try and help the review process.

This PR adds a couple of things to handle eventlog event 0x40f emulation to pcr-calc:
- `txt_info` is a simple platform device driver module that exposes TXT configuration registers in the sysfs. This is required to match an ACM with the current platform without accessing `/dev/mem`.
- Add `-F` option to pcr-calc to read the event-log from a file instead of the sysfs. This is intended to ease testing.
- Add `-A` and `-V` to pcr-calc to pass the path to the ACM and the version of Tboot to emulate event 0x40f. While only event 0x40f is currently implemented, the required scafolding should be there to implement others. Since `-e` already allowed to replace an hash in the event-log, handle emulation through the "emulate" keyword.
- Add `acmmatch` utility to load and parse and ACM file and compare it against the current platform. While `acminfo` does this already, it uses `/dev/mem` which is not desired. Instead, rely on `CPUID`, modules `msr` and `txt_info` to perform the match.

### DIFF
--- a/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions
+++ b/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions
@@ -415,13 +415,15 @@ calculate_drtm_pcrs() {
     local conf=${2}
     local hashalg=${3}
     local srtm="${4}"
-    shift 4
+    local tbver="${5}"
+    shift 5
     local evtlog_ovr="${@}"
     local p17=''
     local p18=''
     local p19=''
     local modules=''
     local quirk=$(check_txt_quirk)
+    local acm=''
 
     if [[ $srtm -eq 1 ]]; then
         modules=$(parse_eficfg $conf)
@@ -456,6 +458,17 @@ calculate_drtm_pcrs() {
         args="${args} -p ${policy}"
 
         [[ ${quirk} -eq 1 ]] && args="${args} -q"
+
+        for f in ${root}/boot/*.{bin,BIN}; do
+            if acmmatch "${f}" >/dev/null; then
+                acm="${f}"
+                break;
+            fi
+        done
+        if [ -z "${acm}" ]; then
+            return 1
+        fi
+        args="${args} -A ${acm} -V ${tbver}"
 
         for m in $modhashes; do
             args="${args} -m ${m}"
--- a/recipes-openxt/txt-info-module/files/sources/txt_info.c
+++ b/recipes-openxt/txt-info-module/files/sources/txt_info.c
@@ -1,0 +1,167 @@
+#include <linux/kobject.h>
+#include <linux/module.h>
+#include <linux/device.h>
+#include <linux/platform_device.h>
+#include <linux/io.h>
+
+#define TXT_PUB_CR_BASE	    0xfed30000
+#define TXT_PUB_CR_SIZE	    0x10000
+static const struct resource txt_resources[] = {
+	{
+		.start = TXT_PUB_CR_BASE,
+		.end = TXT_PUB_CR_BASE + TXT_PUB_CR_SIZE - 1,
+		.flags = IORESOURCE_MEM,
+	},
+};
+#define TXT_PUB_CR_INDEX    0
+
+struct platform_device *pdev;
+struct txt_info {
+	void __iomem *cr_pub;
+	void __iomem *cr_priv;
+};
+static struct txt_info txt_info;
+
+static void __iomem *txt_info_map_regs(struct platform_device *pdev,
+	size_t index)
+{
+	struct resource *res;
+	void __iomem *base;
+
+	res = platform_get_resource(pdev, IORESOURCE_MEM, index);
+	if (IS_ERR(res)) {
+		dev_dbg(&pdev->dev,
+			"Failed to access IOMEM resource %zu.\n", index);
+		return res;
+	}
+
+	base = devm_ioremap(&pdev->dev, res->start, resource_size(res));
+	if (IS_ERR(base))
+		dev_dbg(&pdev->dev,
+			"Failed to ioremap configuration registers.\n");
+
+	return base;
+}
+
+/* Registers offset from TXT_PUB_CR_BASE */
+#define TXT_STS_OFFSET		0x000
+#define TXT_ESTS_OFFSET		0x008
+#define TXT_ERRORCODE_OFFSET	0x030
+#define TXT_VER_FSBIF_OFFSET	0x100
+#define TXT_DIDVID_OFFSET	0x110
+#define TXT_VER_QPIIF_OFFSET	0x200
+
+#define DECLARE_PUB_SHOW_U8(name, offset)				\
+static ssize_t name##_show(struct kobject *kobj,			\
+	struct kobj_attribute *attr, char *buf)				\
+{									\
+	uint8_t v = ioread8(txt_info.cr_pub + (offset));		\
+	return sprintf(buf, "%#04x\n", v);				\
+}									\
+static struct kobj_attribute txt_attr_##name = __ATTR_RO(name);
+
+#define DECLARE_PUB_SHOW_U32(name, offset)				\
+static ssize_t name##_show(struct kobject *kobj,			\
+	struct kobj_attribute *attr, char *buf)				\
+{									\
+	uint32_t v = ioread32(txt_info.cr_pub + (offset));		\
+	return sprintf(buf, "%#010x\n", v);				\
+}									\
+static struct kobj_attribute txt_attr_##name = __ATTR_RO(name);
+
+#define DECLARE_PUB_SHOW_U64(name, offset)				\
+static ssize_t name##_show(struct kobject *kobj,			\
+	struct kobj_attribute *attr, char *buf)				\
+{									\
+	uint64_t v = ioread32(txt_info.cr_pub + (offset) + 0x4);	\
+	v <<= 32;							\
+	v |= ioread32(txt_info.cr_pub + (offset));			\
+	return sprintf(buf, "%#018llx\n", v);				\
+}									\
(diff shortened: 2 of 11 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1899

eric-ch:
v2:
- Refactor to call `destroy_tpm()` when necessary added to the first batch of small fixes.
- call `acm_unload()` if the acm was loaded in the first place.

recipes-security/tboot/tboot-1.9.9/0008-pcr-calc-Add-0x40f-event-emulation.patch
```diff
@@ -0,0 +1,930 @@
...
++static acm_info_table_t *get_acm_info_table(const acm_hdr_t *hdr)
++{
++	uint32_t ovf;
++	uint32_t offset;
++	acm_info_table_t *infotable;
++
++	/* Check infotable is in ACM. */
++	if (__builtin_uadd_overflow(hdr->header_len, hdr->scratch_size, &ovf))
++		return NULL;
++	if (__builtin_umul_overflow(hdr->header_len + hdr->scratch_size, 4, &ovf))
```
jandryuk: __builtin_uadd_overflow stores the result into ovf, so you can do `__builtin_umul_overflow(ovf, 4, &ovf)` and then use the new ovf again `ovf > (hdr->size * 4)`

recipes-security/tboot/tboot-1.9.9/0009-acmmatch-Add-utility-to-match-ACM-with-platform.patch
```diff
@@ -0,0 +1,505 @@
...
++	if (rc < 0) {
++		rc = -ENOSYS;
++		goto out;
++	}
++	rc = read_txt_cr_ver_fsbif(&fsbif);
++	if (rc < 0) {
++		rc = -ENOSYS;
++		goto out;
++	}
++	rc = read_txt_cr_ver_qpiif(&qpiif);
```
jandryuk: The read_txt_* functions could be moved outside of this function since it's called in a loop.

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions
```diff
@@ -457,6 +459,17 @@ calculate_drtm_pcrs() {
 
         [[ ${quirk} -eq 1 ]] && args="${args} -q"
 
+        for f in ${root}/boot/*; do
```
jandryuk: `${root}/boot/*.{bin,BIN}` would skip some things the are definitely not ACMs.

recipes-openxt/txt-info-module/files/sources/txt_info.c
```diff
@@ -0,0 +1,174 @@
...
+		.flags = IORESOURCE_MEM,
+	},
+};
+#define TXT_PRIV_CR_INDEX   0
+#define TXT_PUB_CR_INDEX    1
+
+struct platform_device *pdev;
+struct txt_info {
+	void __iomem *cr_pub;
+	void __iomem *cr_priv;
```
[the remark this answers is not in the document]
eric-ch: +1, @dpsmith pointed to me this morning that the private config registers are actually protected by TBoot and not accessible.

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions
eric-ch: Absolutely.
Also, this is based on the assumption that only one ACM will match a given platform. I do not have proof that is the case.
It would require the have the ACM tested in the same order as TBoot to match the same.

dpsmith: Technically that is an invalid assumption as there could be a revocation ACM and/or an older version of the ACM as well as the SINIT ACM. I have never seen a revocation ACM in the wild and the later can be managed by good housekeeping on us maintainers.

jandryuk: tboot may depend on how grub orders the modules in memory :(

This is a link to a revocation ACM:
https://software.intel.com/system/files/article/183305/3rd-gen-i5-i7-racm-sinit-67_1.zip
chipset_acm_type: 0x9 (unknown)

An SINIT ACM has:
chipset_acm_type: 0x1 (SINIT)

I guess we should check that field in the `info_table`

recipes-security/tboot/tboot-1.9.9/0002-pcr-calc-Generic-print_hash-function.patch
```diff
@@ -0,0 +1,66 @@
...
++        default:
++            return;
+     }
+-    else
+-        return;
++
++    for (i = 0; i < hash_size - 1; ++i)
++        printf("%02x ", b[i]);
++
++    printf("%02x\n", b[i]);
```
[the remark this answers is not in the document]
eric-ch: That would leave a trailing white space ;).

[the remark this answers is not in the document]
jandryuk: Oh, yes, good point :)

`printf("%02x%s", b[i], i == hash_size - 1 ? "\n" : " ")`

Or don't print spaces...

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions
eric-ch: > tboot may depend on how grub orders the modules in memory :(

Grub will organize them as they are passed in the config file it seems. TBoot seems to parse them from the last one to the first one, but the different types are not sorted.

So either we could parse the configuration ... which ends up being different between legacy and UEFI, or ... we maintain a separate list and make sure to update it to reflect the ACM loaded in memory.

We can definitely check for revocation ACM though.

recipes-security/tboot/tboot-1.9.9/0002-pcr-calc-Generic-print_hash-function.patch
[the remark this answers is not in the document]
eric-ch: > printf("%02x%s", b[i], i == hash_size - 1 ? "\n" : " ")

That is an unnecessary `if` in the loop, but yes its not like this is any sort of bottleneck that would need optimisation :). I confess the current version assumes `hash_size > 0`.

> Or don't print spaces...

Some scripts may expect the spaces, I didn't want to risk it.

eric-ch:
v3:
- Remove Private configuration registers definitions from `txt_info`
- Use built-in overflow results instead of re-calculating in `acm.c`
- Move read_txt_* functions out of the main loop.

recipes-security/tboot/tboot-1.9.9/0009-acmmatch-Add-utility-to-match-ACM-with-platform.patch
eric-ch: Addressed in v3.

recipes-security/tboot/tboot-1.9.9/0008-pcr-calc-Add-0x40f-event-emulation.patch
eric-ch: Addressed in v3.

recipes-openxt/txt-info-module/files/sources/txt_info.c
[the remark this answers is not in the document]
eric-ch: Addressed in v3.

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions
jandryuk: ChipsetACMType
Module type (00h =BIOS; 01h = SINIT)Bit 3, if set, flags the ACM as a revocation module, hence “8” is a BIOS revocation AC, “9” is an SINIT revocation AC.

From: https://www.intel.com/content/dam/www/public/us/en/documents/guides/intel-txt-software-development-guide.pdf

recipes-security/tboot/tboot-1.9.9/0009-acmmatch-Add-utility-to-match-ACM-with-platform.patch
```diff
@@ -0,0 +1,500 @@
...
++{
++	struct acm *acm;
++	int rc;
++
++	assert(path != NULL);
++
++	acm = acm_load(path);
++	if (acm == NULL)
++		return -EINVAL;
++
```
jandryuk: Add a 
```
if (acm->infotable->chipset_acm_type != ACM_TYPE_SINIT)
     return -EINVAL
```
(`#define ACM_TYPE_SINIT 1` in pcr-calc/acm.h)

eric-ch:
v4:
- Have acmmatch check ACM_TYPE_SINIT. Add some renaming to make the macros more explicit.
- Check `mmap(3p)` return value against `MAP_FAILED`.

eric-ch:
v5:
- Fold patches into 0006 for merge.
- List ACMs in `${root}/boot/*.{bin,BIN}` instead of `${root}/boot/*`.

That last point might not be enough and I would like to address this problem in a futur PR. There is at least 1 other PR (TCG agile event log 2, ticket OXT-1507) that sits on this and could start receiving testing in the mean time.

eric-ch:

jean-edouard:
LGTM, merging soon.

PR #1103, opened 2019-03-14
refpolicy: Turn global_ssp on

jandryuk:
Turn the global_ssp policy boolean on.  SSP - Stack Smashing Protection -
requires reading urandom to get a random cookie value.  Instead of
adding permissions to individual domains, this gives all SELinux domains
permission to read /dev/urandom.

This patch was sitting in-tree but never added to SRC_URI, so never
applied.


crogers1:
Tested this on x64 because we were seeing the random/urandom denials.  Seems to resolve them.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt/builds/1903

eric-ch:
Will merge soon.

PR #1104, opened 2019-03-17
OXT-1476: xen: remove hardcoded target arch in recipes

dozylynx:
xen-libxl and xen-ocaml-libs recipes: Use TARGET_PREFIX to identify the
correct C compiler to supply to the OCaml toolchain.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6461

PR #1106, opened 2019-03-20
Fixup rsyslog.conf packaging

jandryuk:
I had $(sysconfdir} - note the erroneous parenthesis - so it wasn't
expanding the CONFFILES properly.  FILES was also not getting set since
${CONFFILES}_${PN} wasn't getting expanded.  This meant
/etc/rsyslog.conf was still packaged in "rsyslog".

However, /etc/rsyslog.conf was still getting installed from
rsyslog-conf-dom0.  /etc/rsyslog.conf was marked as a CONFFILE even in
the "rsyslog" package.  So when you installed rsyslog-conf-dom0, it
would rename the original to /etc/rsyslog.conf-opkg and then install the
new conf file as /etc/rsyslog.conf.  Dom0 and installer have those extra
-opkg files hanging around.

Adding BAD_RECOMMENDATIONS to both dom0 and installer avoids the *-opkg files hanging around.

NOTE: on a dirty build, I had to do some cleanalls to get rid of linger machine specific rsyslogs that were still hanging around - specifically for the installer.

### DIFF
--- a/conf/distro/openxt-main.conf
+++ b/conf/distro/openxt-main.conf
@@ -86,6 +86,10 @@ PREFERRED_PROVIDER_jpeg-native = "jpeg-native"
 # xen-xenstored-c is a PROVIDE alias to xen.bb.
 PREFERRED_PROVIDER_xen-xenstored = "xen-xenstored-c"
 
+# Default to rsyslog recipe which RPROVIDES rsyslog-conf
+# Dom0 and Installer install rsyslog-conf-dom0 instead.
+PREFERRED_RPROVIDER_rsyslog-conf = "rsyslog"
+
 # rpcgen staging and install paths.
 require xc-rpcgen.inc
 
--- a/recipes-extended/rsyslog/rsyslog-conf-dom0.bb
+++ b/recipes-extended/rsyslog/rsyslog-conf-dom0.bb
@@ -6,7 +6,8 @@ inherit allarch
 
 SRC_URI = "file://rsyslog.conf"
 
-RPROVIDES_$PN = "rsyslog-conf"
+RPROVIDES_${PN} = "rsyslog-conf"
+RCONFLICTS_${PN} = "rsyslog-conf"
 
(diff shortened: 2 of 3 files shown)


eric-ch:
So the documentation of [RRECOMMENDS](https://www.yoctoproject.org/docs/2.7/ref-manual/ref-manual.html#var-RRECOMMENDS) is valid and all is well :).

Back to https://github.com/OpenXT/xenclient-oe/pull/1085, that means any new image specific configuration will have to add `rsyslog-conf` to `BAD_RECOMMENDATION` or `PACKAGE_EXCLUDE` in the image recipe IIUC.

jandryuk:
I can't believe how many ways I've screwed this up.  Sorry about this.

Ok, it seems my '$PN' typo messed up setting RPROVIDES.  Correcting that to ${PN}, the system installs only the expected config file.  BAD_RECOMMENDATIONS is not needed.

It adds a new bitbake note:
'NOTE: consider defining a PREFERRED_RPROVIDER entry to match rsyslog-conf'

Adding `PREFERRED_RPROVIDER_rsyslog-conf = "rsyslog"` to conf/distro/openxt-main.conf silences the NOTE and we still get the config file installed.  PREFERRED_RPROVIDER takes the recipe name and not the package name.

stacktrust:
Please create a JIRA ticket so we can track changes which may affect logging and auditing.

jandryuk:
Force pushed the branch to replace `BAD_RECOMMENDATIONS` with `RPROVIDES` and `PREFERRED_RPROVIDER`.  Only 1 rsyslog.conf is installed.

If I explicitly list rsyslog-conf and rsyslog-conf-dom0 for installation, only the rsyslog-conf-dom0 config file is installed.  I expected an conflict, but maybe the `RPROVIDES` from rsyslog-conf-dom0 means the rsyslog-conf package doesn't get pulled in.

This PR only affects how files are packaged and not the configuration.

eric-ch:

PR #1108, opened 2019-03-21
Quirk to drop PCR1 from sealing set

jandryuk:
Some devices change PCR1 when booted docked versus undocked.  This PR consolidates storage of the pcr sealings set for TPM2 so we can drop out entries if needed - I'm surprised the TPM2 sealing blob doesn't include the sealing set.  The PCRs are stored in /config/config.pcrs and /boot/system/tpm/config.tss.pcrs.  We need to read them out of the unencrypted /boot/system/tpm/config.tss.pcrs so that we can unseal.  If an attacker modified the file, they would break unsealing.  I think the $( < file.pcrs ) construct is safe from command injection - I haven't been able to perform a command injection - the output seems to be interpreted as just a string by the shell.  A ';' is just passed as an argument and does not end the statement.

Next it fixes up the existing pcr1 quirk to work and cover one Dell laptop model.  This should be expanded to cover move devices, but I don't know what to check for.  There is a dmidecode entry for DOCK/JDOCK1, but I don't know what that actually corresponds to.

The PCR1 measurement changes because BoorOrder changes and a new "USB NIC" boot entry is measured into PCR1.

The PR also converts tpm-scripts to allarch since it's shell scripts, and it deletes the unused tpm-setup script and associated files.

FYI, as the installer is currently written, the PCR1 quirk will only be applied on fresh install.

Is tpm-functions tpm_seal() even used?

### DIFF
--- a/recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/pcr1-detect.sh
+++ b/recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/pcr1-detect.sh
@@ -25,5 +25,18 @@ if [ "${VENDOR_ID}" = "8086" ]; then
             MANUFACTURER=$(dmidecode --type 1 | sed -n -re 's/.*Manufacturer: (.*)$/\1/p' | cut -d ' ' -f 1)
             VERSION=$(dmidecode --type 1 | sed -n -re 's/.*Version: (.*)$/\1/p')
             ;;
+        5910)
+            MANUFACTURER=$(dmidecode --type 1 | \
+                           sed -n -re 's/.*Manufacturer: (.*)$/\1/p' | \
+                           cut -d ' ' -f 1)
+            PRODUCT=$(dmidecode --type 1 | \
+                      sed -n -re 's/.*Product Name: (.*)$/\1/p')
+            if [ "$MANUFACTURER" = "Dell" ] && \
+               [ "$PRODUCT" = "Latitude 5580" ] ; then
+                exit 1
+            fi
+            ;;
     esac
 fi
+
+exit 0
--- a/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
+++ b/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
@@ -395,7 +395,7 @@ encrypted_unlock() {
     then
         if pcr_bank_exists "sha256"; then
             local unseal_file=${key_file}
-            tpm2_unsealdata -H 0x81000000 -n "${unseal_file}.sha256" -u "${key_file}.pub.sha256" -g 0xB -r 0 -r 1 -r 2 -r 3 -r 4 -r 5 -r 7 -r 8 -r 15 -r 17 -r 18 -r 19 | cryptsetup -q -d - -S ${ESLOT} luksOpen "${part}" "${name}" >/dev/null 2>&1
+            tpm2_unsealdata -H 0x81000000 -n "${unseal_file}.sha256" -u "${key_file}.pub.sha256" -g 0xB $( < ${unseal_file}.pcrs )| cryptsetup -q -d - -S ${ESLOT} luksOpen "${part}" "${name}" >/dev/null 2>&1
             ret=$?
         fi
     else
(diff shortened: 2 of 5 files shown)


recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/pcr1-detect.sh
```diff
@@ -25,5 +25,18 @@ if [ "${VENDOR_ID}" = "8086" ]; then
...
             VERSION=$(dmidecode --type 1 | sed -n -re 's/.*Version: (.*)$/\1/p')
             ;;
+        5910)
+            MANUFACTURER=$(dmidecode --type 1 | \
+                           sed -n -re 's/.*Manufacturer: (.*)$/\1/p' | \
+                           cut -d ' ' -f 1)
+            PRODUCT=$(dmidecode --type 1 | \
+                      sed -n -re 's/.*Product Name: (.*)$/\1/p')
+            if [ "$MANUFACTURER" = "Dell" ] && \
+               [ "$PRODUCT" = "Latitude 5580" ] ; then
```
pearsonk: I foresee this growing to a large list of quirks over time. Would it make more sense to make an install/answerfile option to include or exclude PCR[1]? This should be accompanied with some sort of disclaimer citing what the security implications of ignoring PCR[1] are to the overall trust of the system.

jandryuk: I was going for "just works" to avoid adding knobs.  But like you say, it does change the security.  If a user doesn't own a dock, then they could leave it enabled.  But then later they purchase one and it throws off the measurement.

Regarding the list size, yes, it could grow quite large.  I would prefer to base detection off some attribute rather than the manufacturer and product.  I just don't know what attribute that would be.  USB-C port?  That could be too broad, and I'm not sure how to check it.

You were right on the telecon.  It's the USB NIC that was changing the measurement and not an Option ROM.  I updated the patch description to reflect reality.

jean-edouard: Yeah, I've been wanting to add that quirk for some time myself, after seeing machines that measure stupid stuff into PCR1.
The reason I've never done it is that if you don't measure PCR1, you can't really guarantee anymore that the hardware/firmware haven't been tampered with, which pretty much means all bets are off... Unless you know exactly what's measured into PCR1, and it happens to not be anything important.
That being said, it's useful to be able to exclude PCR1, but I agree with @pearsonk that the user should be made aware of it, or even have the ability to decide whether or not to include it.

jandryuk:
I force pushed to fixup the last patch.  It was missing removal of the tpmsetup files from refpolicy-msc SRC_URI.

stacktrust:
Please create a JIRA ticket so we can track this quirk for future devices and release notes.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/pcr1-detect.sh
pearsonk: Ah, the USB/PCR[1] issue is fairly ubiquitous in KBL+ systems. This can be seen by adding/removing drives or even changing where a keyboard/mouse are plugged in on a system. It is very annoying and we are pushing for it to get changed.

jandryuk:
I dropped the last two cleanup commits from this PR.  I had not tested them and they break the system since removing tpm-setup drops `xenclient-tpm-scripts openxt-keymanagement openxt-measuredlaunch` from RDEPENDS.  The two commits still in the PR have been tested.

This incorrect push happened because the pcr1-quirk branch on my workstation (where I pushed to github) and in my build system had diverged.  I mistakenly thought they were in sync when I pushed the workstation version.

eric-ch:

stacktrust:
Please add this ticket number to the PR

https://openxt.atlassian.net/browse/OXT-1594

jandryuk:
This PR is broken. seal-system hard-codes its list of PCRs, so it doesn't follow /config/config.pcrs.

PR #1110, opened 2019-03-25
Update console-setup to 1.188 while also updating the SRC_URI

rneilturner:
Signed-off-by: turnerr <turnerr@ainfosec.com>


rneilturner:
The version we use (1.88) is distributed under Debian Wheezy which is no longer supported. In addition, 1.88 version of console-setup was removed from the Debian ftp site. I created this PR to update console-setup to the version supported in Debian Buster.

eric-ch:

PR #1113, opened 2019-03-27
OXT-1507: Support TCG EFI Agile crypto log 2.0.

eric-ch:
- Add the required logic to the `txt` module to export the eventlog. Compared to previous version, the heap header structure is not necessary.
- Pass the format of the event-log from TBoot to Xen and from Xen to module `txt`. That last one will create a `securityfs` node named differently to reflect the version of the ABI of the event-log. The issue here is the legacy TPM 2.0 agile log that passed in the TBoot heap structure the algorithm used for the hashes. While the TCG (EFI?) TPM 1.2 and TCG EFI TPM 2.0 format defined in the `EFI Protocol Specification` give enough information to figure out the version of the log from its content header, it is not the case for the legacy format that was used in between. Since TPM 1.2 was already handled with a different `securityfs` node, follow that path.
- Parse the new format in `pcr-calc`. The version of the event-log is determined by the `securityfs` node existence and name, or if `-F` is used, by the TPM version and the new option `-o` that will inform `pcr-calc` that it should parse the TPM 2.0 legacy format.
- The last commit fixes an annoyance when building `pcr-calc` with `debug=y` in `Config.mk`, this was introduced by https://github.com/OpenXT/xenclient-oe/pull/1100 and only affects manual configuration of the build system.


recipes-kernel/linux/4.19/patches/xen-txt-add-xen-txt-eventlog-module.patch
```diff
@@ -181,13 +181,21 @@ Subject: [PATCH] xen-txt: add xen txt eventlog module
...
++			break;
++		case TXTOP_EVTLOG_FORMAT_LEGACY_20:
++			filename = "tpm20_binary_evtlog_legacy";
++			break;
++		case TXTOP_EVTLOG_FORMAT_TCG_20:
++			filename = "tpm20_binary_evtlog_tcg";
++			break;
++		default:
++			printk(KERN_ERR "Incompatible event-log format: %x\n", txt_log.format);
++                        ret = EINVAL;
```
jean-edouard: Spaces here too.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6474
Merging soon after indent fixes.

eric-ch:
v2:
- Amend invalid indentation.

PR #1117, opened 2019-04-02
module-signing: Prefer the build-tree sign-file

jandryuk:
My first build after the 4.19 uprev, I saw do_sign_modules fail for the
linux-openxt recipe:
/home/build/openxt-master/build/tmp-glibc/work-shared/xenclient-dom0/kernel-build-artifacts/scripts/sign-file: error while loading shared libraries: libcrypto.so.1.0.2: cannot open shared object file: No such file or directory

sign-file's path is
build/tmp-glibc/work-shared/xenclient-dom0/kernel-build-artifacts/scripts/sign-file - note there is no version, so it is from 4.14.  And it's rpath is:
/home/build/openxt-master/build/tmp-glibc/work/xenclient_dom0-oe-linux/v4v-module/gitAUTOINC+3f8a139271-r0/recipe-sysroot-native/usr/lib:/home/build/openxt-master/build/tmp-glibc/work/xenclient_dom0-oe-linux/v4v-module/gitAUTOINC+3f8a139271-r0/recipe-sysroot-native/lib

The v4v-module temp dir has been cleaned, so recipe-sysroot-native no
longer exists.

Order ${B} first so that a kernel build prefers the sign-file in its own
build directory over ${STAGING_KERNEL_BUILDDIR}.  That way, we won't try
to use a stale sign-file that has an rpath into some module's
recipe-sysroot-native.  Modules won't find sign-file in ${B}, so they'll
use ${STAGING_KERNEL_BUILDDIR}.

### DIFF
--- a/classes/module-signing.bbclass
+++ b/classes/module-signing.bbclass
@@ -17,6 +17,9 @@ export KERNEL_MODULE_SIG_KEY
 # build-time keypair will be generated and used for signing and embedding.
 export KERNEL_MODULE_SIG_CERT
 
+# Kernel builds will override this with ${B}/scripts/sign-file
+SIGN_FILE = "${STAGING_KERNEL_BUILDDIR}/scripts/sign-file"
+
 do_sign_modules() {
     if [ -n "${KERNEL_MODULE_SIG_KEY}" ] &&
        grep -q '^CONFIG_MODULE_SIG=y' ${STAGING_KERNEL_BUILDDIR}/.config; then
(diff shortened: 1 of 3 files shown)


classes/module-signing.bbclass
```diff
@@ -26,8 +26,8 @@ do_sign_modules() {
         [ -z "$SIG_HASH" ] && bbfatal CONFIG_MODULE_SIG_HASH is not set in .config
 
         # ${B} for kernel, ${STAGING_KERNEL_BUILDDIR} for modules
-        for SIGN_FILE in ${STAGING_KERNEL_BUILDDIR}/scripts/sign-file \
-                         ${B}/scripts/sign-file ; do
+        for SIGN_FILE in ${B}/scripts/sign-file \
+                         ${STAGING_KERNEL_BUILDDIR}/scripts/sign-file ; do
```
eric-ch: While I'm fine with this, shouldn't this be addressed with a variable to use either the native sysroot for out-of-tree modules and `${B}` for the kernel?
A little like `autotools-brokensep`, except this one would be `module-signing-out-of-tree.bbclass` or something?

jandryuk: Yeah, that's probably better.

I could do:
SIGN_FILE = "${STAGING_KERNEL_BUILDDIR}/scripts/sign-file"
in module-signing.bbclass
and then SIGN_FILE = "${B}/scripts/sign-file" in linux-openxt.inc without introducing a new bbclass.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6483

jandryuk:
Force pushed a new change using a variable as suggested by @eric-ch

eric-ch:

eric-ch:
Will merge soon.

eric-ch:
@jandryuk Can you PR against stable-9 as well please?

PR #1118, opened 2019-04-02
OXT-1541: [xen] Make sure xen Kconfig uses the new TXT_OP

crogers1:
The x64 uprev was missing a key patch for xen.  With the txt_op
  hypercall ifdef'd out to support building xen-firmware such as
  the pv-shim, the xen/common/Kconfig needs an entry so we can
  optinally set CONFIG_TXT_OP to true and enable the txt_op
  hypercall.

  OXT-1541

### DIFF
--- a/recipes-extended/xen/files/tboot-xen-evtlog-support.patch
+++ b/recipes-extended/xen/files/tboot-xen-evtlog-support.patch
@@ -13,6 +13,8 @@ CHANGELOG
 Written for OpenXT. Feature introduced in the stable-7 release.
 Patch updated for Xen 4.8, 4.9.
 
+ 4/2/2019: Added txt_op config option for Kconfig
+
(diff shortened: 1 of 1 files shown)


crogers1:
Verified the txt module will load properly after recompiling xen-hypervisor with this patch included.  Also verified the xen.bb (xen-firmware, etc) will build without complaining about txt_op.

crogers1:
Patch is now a part of tboot-xen-evtlog-support.patch

eric-ch:
Tested on a local tree, the CONFIG is now kept.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6483

eric-ch:
Will merge soon.

PR #1119, opened 2019-04-03
linux: revert 2 commits that force console output to the screen

jean-edouard:

### DIFF
--- a/recipes-kernel/linux/4.19/linux-openxt_4.19.32.bb
+++ b/recipes-kernel/linux/4.19/linux-openxt_4.19.32.bb
@@ -37,6 +37,8 @@ SRC_URI += "${KERNELORG_MIRROR}/linux/kernel/v${PV_MAJOR}.x/linux-${PV}.tar.xz;n
     file://netback-vwif-support.patch \
     file://gem-foreign.patch \
     file://xen-txt-add-xen-txt-eventlog-module.patch \
+    file://revert-x86-Xen-further-refine-add_preferred_console-.patch \
+    file://revert-x86-xen-add-tty0-and-hvc0-as-preferred-consol.patch \
     file://xsa-155-qsb-023-add-RING_COPY_RESPONSE.patch \
     file://xsa-155-qsb-023-xen-blkfront-make-local-copy-of-response-before-usin.patch \
     file://xsa-155-qsb-023-xen-blkfront-prepare-request-locally-only-then-put-i.patch \
--- a/recipes-kernel/linux/4.19/patches/revert-x86-Xen-further-refine-add_preferred_console-.patch
+++ b/recipes-kernel/linux/4.19/patches/revert-x86-Xen-further-refine-add_preferred_console-.patch
@@ -1,0 +1,25 @@
+[PATCH 1/2] Revert "x86/Xen: further refine add_preferred_console()
+ invocations"
+
+This reverts commit 71dc05635983ae711694f748d006e107387f1bab.
+---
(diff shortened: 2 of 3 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6487

jean-edouard:
Note: this a less-than-ideal way of fixing the problem. I'll gladly close this if a better fix gets PR-ed.

jandryuk:
I tried a non-EFI boot adding "quiet loglevel=0" to the linux command line.  That seems to remove all the messages.

eric-ch:
@jandryuk That will silence the output entirely, which is not what this PR means to do. The reverted commits forced the output on `tty0` (current console) which will end up on the installer display (tty1 I presume). It was not the case before when specifying `tty2` on the cmdline.

jandryuk:
I thought this PR was trying to reduce boot output during a normal boot as opposed to the installer since that changed with the 4.19 uprev.  @jean-edouard what are you trying to achieve here?

jean-edouard:
As @eric-ch said, the point is to avoid to logs to be shown on screen at boot time. But we do want them in dmesg/syslog. This is purely cosmetic.

jandryuk:
With "quiet loglevel=0", dmesg still has the messages and syslog picks them up and writes them to /var/log/messages - they just don't appear on the display.

jean-edouard:
@jandryuk oh interesting, I thought that would have limited the actual logs we get from the kernel / modules. Will test.

eric-ch:
@jandryuk (I got confused and thought of the installer spurious logs, this is for initial console output at boot time)

`quiet loglevel=0` will shutdown the serial log output as well as the default console.

jandryuk:
Oh, that's unfortunate that it stops serial console output.  I had not checked that.

eric-ch:
I am surprised these original patches did not use `console_set_on_cmdline` to avoid overriding cmdline configuration. Not that it was checked before either though. Would that make sense?

jandryuk:
I wasn't aware of console_set_on_cmdline, but it looks like it might be usable.  Provided the cmdline is parsed...  That doesn't look promising.

The add_preferred_console calls are in xen_start_kernel
xen_start_kernel
  x86_64_start_reservations
    start_kernel
      setup_arch - this is where it grabs the command line.

I think the above is correct - I haven't found where it is parsed.

Xen ARM has:
```
        if (!console_set_on_cmdline && !xen_initial_domain())
                add_preferred_console("hvc", 0, NULL);
```
But that may come after the command line is parsed.

eric-ch:

eric-ch:
@jandryuk Right, the cmdline is not parsed yet, so `console_set_on_cmdline` is not ready.

This change only affected dom0, do you think it is preferable to add a single patch removing the `add_preferred_console("tty", 0, NULL);`, putting it back under `if (!xen_initial_domain())`, as a temporary solution instead of reverting upstream changes?

jandryuk:
For the time being, the simplest thing may be just to add a patch commenting out the tty lines.  With a nice patch description stating exactly why we are doing this.

The proper fix would probably be to delay `add_preferred_console` calls until after the command line is parsed so the entire code block can be placed under `if (!console_set_on_cmdline)`.  Provided x86 actually parses the command line before enabling the regular consoles...

eric-ch:
Without the reverts: https://github.com/OpenXT/xenclient-oe/pull/1126

PR #1120, opened 2019-04-04
shim: switch to upstream + patchqueue

jean-edouard:
This is a first step, which switches to the upstream head that was used for the fork.
There shouldn't be any code difference.
Next step is to upgrade to v15.

OXT-1413


eric-ch:
That should probably be submitted to stable-9 as well.

jean-edouard:
Good point, done. https://github.com/OpenXT/xenclient-oe/pull/1125

eric-ch:

eric-ch:
Will merge soon.

PR #1121, opened 2019-04-04
OXT-1507, OXT-1540 & more: Changes to forward-sealing binaries usage.

eric-ch:
OXT-1506 and OXT-1507 introduced non-retro compatible changes that have to be mitigated when performing OTA upgrade.
The running kernel:
1. Does not have the new entries in the sysfs to access the eventlog.
2. Does not have access to he TXT configuration registers through sysfs.

The first point is mitigated by using the previous sysfs entry, this is also useful to determine which capabilities we can expect to do the forward sealing.
The second can be worked around using `txt-stat` that will report the required values to find the ACM that will be loaded upon reboot into the upgraded system. `acmmatch` is modified in consequence to accept value overrides.

### DIFF
--- a/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions
+++ b/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions
@@ -381,6 +381,8 @@ hash_modules() {
 }
 
 check_txt_quirk() {
+    local args
+
     modprobe txt > /dev/null 2>&1
     [ -d /sys/kernel/security/txt ] || return 0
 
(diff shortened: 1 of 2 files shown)


eric-ch:
v2:
- Port the non stable-9 dependent changes to the script and measurement tools. These changes could be useful later for debugging/analysis.
- Change `pcr-calc` behavior to consider the former sysfs node `/sys/kernel/security/txt/tpm20_binary_evtlog` a "default" and parse this eventlog without requiring command-line options.

eric-ch:

jean-edouard:
LGTM, merging soon.

PR #1126, opened 2019-04-11
linux: Don't set tty0 as preferred console for pv.

eric-ch:
None of the PV kernels used in OpenXT rely on that, yet silencing the loglevel is not desired as the serial output is often used to debug boot issues. This change does _not_ add tty0 as a default preferred console.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6513

jean-edouard:
LGTM, merging soon.

jean-edouard:
@eric-ch, stable-9?

eric-ch:
@jean-edouard https://github.com/OpenXT/xenclient-oe/pull/1129

PR #1127, opened 2019-04-11
mod-sign: Add dep on later kernel task.

eric-ch:
This is mostly noticeable if `KERNEL_MODULE_SIG_KEY` is not provided. Refreshing a patch, or changing kernel revision, will not always have the out-of-tree modules be signed with the new key.

Add a dependency on the kernel `do_install` task to have the new key files present in the shared workdir when running sign_modules and re-run do_sign_modules if `virtual/kernel:do_install` was done since last signing.

### DIFF
--- a/classes/module-signing.bbclass
+++ b/classes/module-signing.bbclass
@@ -26,7 +26,7 @@ do_sign_modules() {
         SIG_HASH=$( grep CONFIG_MODULE_SIG_HASH= \
                         ${STAGING_KERNEL_BUILDDIR}/.config | \
                       cut -d '"' -f 2 )
-        [ -z "$SIG_HASH" ] && bbfatal CONFIG_MODULE_SIG_HASH is not set in .config
+        [ -z "$SIG_HASH" ] && bbfatal "CONFIG_MODULE_SIG_HASH is not set in .config"
 
         [ -x "${SIGN_FILE}" ] || bbfatal "Cannot find scripts/sign-file"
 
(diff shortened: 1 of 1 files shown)


jandryuk:
If `KERNEL_MODULE_SIG_KEY` is not provided, i.e. we are using an automatically generated key, then the kernel build system signs the modules during `make modules_install`.  `do_sign_modules` is a no-op unless you are using an explicit key pair.

You're seeing a case where you changed linux-openxt, but a module wasn't re-built?  In my experience, whenever I change linux-openxt, it rebuilds all the modules and a bunch of userspace - at least when I re-build the entire image.

If you are seeing a kernel rebuild resulting in stale keys for out-of-tree modules, do_shared_workdir copies the auto-generated cert & key from the kernel build directory into the shared workdir.  Out-of-tree modules use it as their `O=` directory, so the dependency should be on do_shared_workdir?

In pyro, that dependency is already present in module.bbclass:
`do_make_scripts[depends] += "virtual/kernel:do_shared_workdir"`
and
`addtask make_scripts after do_prepare_recipe_sysroot before do_compile

In sumo, do_make_scripts goes away for a separate make-mod-scripts recipe
http://git.openembedded.org/openembedded-core/commit/?id=ea12c46fe8748fb6606c603d463075a8624e6563
... but it still has `do_configure[depends] += "virtual/kernel:do_shared_workdir"`

I'm still a little confused how you experienced the stale keys.  As far as I know, a kernel re-compile should re-compile all out-of-tree modules.

I've wondered if do_sign_modules should be changed to do_install_append.  That would simplify some of the dependencies.  I forget why I made a separate task originally.  As long as we run the action after the modules are in ${D}, I think we should be okay.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6513

eric-ch:
@jandryuk You are correct, I misread the use of `KERNEL_MODULE_SIG_KEY` thinking the `do_sign_modules` was doing the signing.

The out-of-tree modules were signed, after making changes to `linux-openxt` using a stale (or invalid) key. As you point out the dependency is already there. I had merged https://github.com/OpenXT/xenclient-oe/pull/1117 in my tree, but already made a build before merging. Maybe something was confused by that.

Anyway I need to get a more consistent repro, if one occurs, to figure this one out.

Closing.

PR #1128, opened 2019-04-12
STABLE-9: module-signing: Re-work sign-file selection

jandryuk:
On a Linux uprev, module signing can fail on a dirty build when
sign-file is called with a rpath pointing into a modules sysroot.  The
module's sysroot path will be gone, so sign-file fails to run.
/home/build/openxt-master/build/tmp-glibc/work-shared/xenclient-dom0/kernel-build-artifacts/scripts/sign-file: error while loading shared libraries: libcrypto.so.1.0.2: cannot open shared object file: No such file or directory

For a kernel build, we should run ${B}/scripts/sign-file.  Instead of
searching for this at runtime, use a variable to set the the SIGN_FILE
location for modules and override for a kernel build.

While doing this, move 'require classes/module-signing.bbclass' into
kernel-signing.bbclass and rename to kernel-module-signing.bbclass.
That consolidates the functionality required for the kernel in one
place.

(cherry picked from commit 50524913e0b9d8dd5607501cd4a7789780878f25)

Stable-9 version of https://github.com/OpenXT/xenclient-oe/pull/1117#event-2269424019

### DIFF
--- a/classes/kernel-module-signing.bbclass
+++ b/classes/kernel-module-signing.bbclass
@@ -1,3 +1,7 @@
+require classes/module-signing.bbclass
+# override value from module-signing to use the in-tree utility
+SIGN_FILE = "${B}/scripts/sign-file"
+
 # Set KERNEL_MODULE_SIG_CERT in local.conf to the filepath of a public key
 # to embed in the kernel to verify signed modules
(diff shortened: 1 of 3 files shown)


jean-edouard:
Merging soon.

PR #1129, opened 2019-04-15
STABLE-9: linux: Don't set tty0 as preferred console for pv.

eric-ch:
None of the PV kernels used in OpenXT rely on that, yet silencing the loglevel is not desired as the serial output is often used to debug boot issues. This change does _not_ add tty0 as a default preferred console.

(cherry picked from commit ef42a9036283da7bfab5d049df383927c9dd581d)


jean-edouard:
Merging soon.

PR #1130, opened 2019-04-16
Add tag MULTIBOOT2_TAG_TYPE_EFI64 from Xen EFI to TBOOT

rneilturner:
The multiboot header going from Xen EFI to TBOOT was missing the the MULTIBOOT2_TAG_TYPE_EFI64 tag, which is a pointer to the EFI System Table. TBOOT would fail to load the ACPI tables from EFI thereby causing a failure when going to SENTER. With this tag TBOOT now recognizes the preload as EFI and loads the proper tables including the Root System Description Table (RSDT). This resolves the UEFI+TXT boot loop on both HP's and NCS systems when going into SENTER. There are still other issues with HP systems such as incorrectly reporting PCR 1 changes when powering off.  I have not tested Intel systems such as the NUC. 

OXT-1307

### DIFF
--- a/recipes-extended/xen/files/tboot-measure-and-launch-from-xen-efi.patch
+++ b/recipes-extended/xen/files/tboot-measure-and-launch-from-xen-efi.patch
@@ -136,7 +136,7 @@ Signed-off-by: Tamas K Lengyel <lengyelt@ainfosec.com>
  
  static void __init edd_put_string(u8 *dst, size_t n, const char *src)
  {
-@@ -212,6 +239,224 @@ static void __init efi_arch_process_memo
+@@ -212,6 +239,231 @@ static void __init efi_arch_process_memo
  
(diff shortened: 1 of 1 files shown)


rneilturner:
Once reviewed and accepted I will create a stable-9 PR.

stacktrust:
JIRA is not associating this PR with the ticket, the ticket number may be required in the title

dpsmith:
That is because @rneilturner put the word "See" in the front of the ticket line. The hook looks for a sole line starting the "OXT-".

rneilturner:
> That is because @rneilturner put the word "See" in the front of the ticket line. The hook looks for a sole line starting the "OXT-".

All set

jean-edouard:

jean-edouard:
LGTM, merging soon.

PR #1133, opened 2019-04-18
[xcpmd] Wait for db to initialize before reading policy

tsirakisn:
This only occurs on certain platforms, but occasionally
xcpmd will try to load its policy from the db before the
db has fully initialized. Make sure the db is reachable
over dbus before starting up.

OXT-1561

Signed-off-by: Nicholas Tsirakis <tsirakisn@ainfosec.com>

### DIFF
--- a/recipes-openxt/dbd/dbd/dbd.initscript
+++ b/recipes-openxt/dbd/dbd/dbd.initscript
@@ -43,6 +43,7 @@ start() {
         fi
 	echo -n "Starting dbd: "
 	/usr/bin/dbd 2> /dev/null 2>&1
+	wait_for_db
 	echo "OK"
 }
 
@@ -55,6 +56,32 @@ stop() {
 restart() {
 	stop
 	start
+}
+
+# make sure the db is fully initialized so other
+# processes can read the rules they need
+wait_for_db() {
+	local attempts=0
+	local max_attempts=10
+
+	while [ $attempts -lt $max_attempts ]; do
+		check_db
+		if [ $? -eq 0 ]; then
+			return
+		fi
+
+		attempts=$((attempts + 1))
+		sleep .2
+	done
+	echo "WARN: db initialization timed out" | logger -t dbd
+	echo "WARN: db may be unavailable over dbus" | logger -t dbd
+}
+
+check_db() {
+	dbus-send --print-reply --system
+			--type=method_call \
+			--dest=com.citrix.xenclient.db \
+			/ com.citrix.xenclient.db.exists string:"/" >/dev/null 2>&1
 }
 
 case "$1" in


pearsonk:
Quick glance looks like this works (max_tries spinlock) but it is implemented in the init script. Wouldn't it make more sense for it to be implemented in xcpmd itself?

jean-edouard:
If the dbd initscript ensured everything is up before returning, wouldn't dependent initscripts be guaranteed that the db is up? This has been a problem for other programs in the past, it would be nice to fix this once and for all! (or maybe a more generic version could be added to `/etc/init.d/functions`?)
That being said, I'm ok with this fix, even if I also agree with @pearsonk that it would make more sense to do it in the binary.

tsirakisn:
Doing it in the binary does make more sense. The init script was the quick and dirty way for me to test/verify, so I wasn't even thinking of other options. @jean-edouard does bring up a good point about having dbd itself wait until it is initialized though, would that be a more desirable fix? Would the lower-priority init scripts be guaranteed to run after dbd's if there is a delay in there? Genuine question, don't know how reliable the startup order is for the rc*.d scripts.

pearsonk:
+1 to Jed's alternative approach. If DBD's init script ensures that it is ready (not just a started process), that would reduce this problem. The only problematic window I can think of would be a DBD crash and restart in the time window of sysvinit starts a dependency. Very low likelihood of occurrence, so probably an acceptable risk.

jean-edouard:

eric-ch:
> Would the lower-priority init scripts be guaranteed to run after dbd's if there is a delay in there? Genuine question, don't know how reliable the startup order is for the rc*.d scripts.

Not after `dbd`, but after the initscript that started it.
From `/etc/init.d/rc` (skipping most of the details):
```sh
        for i in /etc/rc$runlevel.d/S*
        do
# [...] checks on what the initscript is and if it's a script of binary, transition runlevel, etc...
. $scriptname
```
If `$scriptname` blocks, the bootstrap is held up until it no longer blocks. Globbing expansion is alphabetically sorted[[1]](https://linux.die.net/man/1/sh).

tsirakisn:
This one slipped through the cracks. Updated PR to wait for db initialization in dbd initscript.

tsirakisn:
Ready for review.

recipes-openxt/dbd/dbd/dbd.initscript
```diff
@@ -57,6 +58,32 @@ restart() {
...
+
+		attempts=$((attempts + 1))
+		sleep .2
+	done
+	echo "WARN: db initialization timed out" | logger -t dbd
+	echo "WARN: db may be unavailable over dbus" | logger -t dbd
+}
+
+check_db() {
+	dbus-send --print-reply --system
```
eric-ch: Missing `\` here.

tsirakisn:
This PR needs to be put on hold again. The missing backslash caused me to miss an SELinux denial that was occurring during startup. Trying to figure out if a rule needs to be added to rectify this or if there is another way to do this that doesn't anger SELinux.

For those curious, the denial is dbd_t --> initrc_t, send_msg.

jandryuk:
Closing.  Waiting for service initialization is handled by https://github.com/OpenXT/xenclient-oe/pull/1410

PR #1134, opened 2019-04-18
seal-system: Exit if unable to read rootfs

jandryuk:
Running seal-system as sysadm_t does not have proper permission to read
/dev/mapper/xenclient-oe.  However there is no error checking for this
case, so seal-system will continue on and forward seal to an invalid
rootfs.  If we are unable to read the rootfs, abort instead of
clobbering the current sealed blobs.


jean-edouard:

jean-edouard:
LGTM, merging soon.

jean-edouard:
stable-9?

PR #1136, opened 2019-04-19
STABLE-9: OXT-1547: [key-functions] Maintain backwards compat

crogers1:
  is in uppercase, and in stable-9, it's lowercase. get-config-key.c
  hasn't changed, but puts() must have changed how it dumps to
  stdout.

  OXT-1547

### DIFF
--- a/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
+++ b/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
@@ -137,7 +137,7 @@ seed_entropy() {
 gen_platform_key() {
     local key_file=`mktemp -t`
 
-    get-config-key > ${key_file}
+    get-config-key | tr '[:lower:]' '[:upper:]' > ${key_file}
 
     [ "$(getenforce)" == "Enforcing" ] &&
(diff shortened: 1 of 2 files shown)


crogers1:
This was my knee-jerk reaction for fixing the problem, but I almost prefer this over changing get-config-key.c, which hasn't changed at all in several releases.

jean-edouard:
I think it's just the Linux dmi driver that recently decided to change the case of `/sys/class/dmi/id/product_uuid`, no? That actually rings a bell...

jean-edouard:
Either way, we need to use the uppercase version forever to avoid breaking upgrade. This should probably be PR-ed against master.

jean-edouard:
https://savannah.nongnu.org/bugs/index.php?53569

crogers1:
If you guys are good with this approach I'll PR it against master too

jean-edouard:
I would have patched get-config-key, just because the name of the program implies that it will get you the config key (not the product uuid), but this is ok too.
FYI, we usually PR against master first and cherry-pick into stable-9.

crogers1:
Closing this and opening a master PR with an updated commit message.

PR #1137, opened 2019-04-19
STABLE-9: OXT-1564: Remove /etc/modules from installer image.

eric-ch:
Change how `/etc/modules` is packaged to make it machine agnostic and use a different recipe per image.
`modules-installer` is removed since the installer does not deploy any module.
This silence spurious warning during the installation bootstrap.


jean-edouard:
Merging soon.

PR #1138, opened 2019-04-22
STABLE-9: seal-system: Exit if unable to read rootfs

jandryuk:
Stable-9 version of https://github.com/OpenXT/xenclient-oe/pull/1134/commits/1a80fbebedae58c0111d3b9610850e3976cf5c90

Running seal-system as sysadm_t does not have proper permission to read
/dev/mapper/xenclient-oe.  However there is no error checking for this
case, so seal-system will continue on and forward seal to an invalid
rootfs.  If we are unable to read the rootfs, abort instead of
clobbering the current sealed blobs.

(cherry picked from commit 1a80fbebedae58c0111d3b9610850e3976cf5c90)


jean-edouard:
Merging soon.

PR #1139, opened 2019-04-22
OXT-1547: [key-functions] Maintain backwards compat

crogers1:
  is in uppercase, and in stable-9, it's lowercase. This is a
  result of a change to dmidecode's product uuid output. This
  commit maintains support for upgrading from older versions
  of OpenXT where the UUID is uppercase.

  OXT-1547

### DIFF
--- a/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
+++ b/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
@@ -137,7 +137,7 @@ seed_entropy() {
 gen_platform_key() {
     local key_file=`mktemp -t`
 
-    get-config-key > ${key_file}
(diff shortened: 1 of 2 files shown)


eric-ch:
I do not like that `get-config-key` at all. While I would have preferred to fix the issue in the tool, this is fine and we should add a ticket to remove it altogether.

dpsmith:
@eric-ch I would not disagree but would say that a JIRA ticket should be opened so that a discussion can be captured on how we would like to handle the function that it provided.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6525

jean-edouard:
LGTM, merging soon.

eric-ch:
Ticket for replacement: https://openxt.atlassian.net/browse/OXT-1573

PR #1140, opened 2019-04-23
STABLE-9: OXT-1547: [key-functions] Maintain backwards compat

crogers1:
  is in uppercase, and in stable-9, it's lowercase. This is a
  result of a change to dmidecode's product uuid output. This
  commit maintains support for upgrading from older versions
  of OpenXT where the UUID is uppercase.

  OXT-1547

### DIFF
--- a/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
+++ b/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
@@ -137,7 +137,7 @@ seed_entropy() {
 gen_platform_key() {
     local key_file=`mktemp -t`
 
-    get-config-key > ${key_file}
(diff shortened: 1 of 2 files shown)


crogers1:
Can't upgrade from a release with uppercase product_uuid (8.0 and earlier) without the 32-bit compat stuff, and that's only in stable-9.  Normally master would be tested first, and then cherry-picked into stable-9, but we need to do the reverse here.  @jean-edouard, I think that explains it.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6528

jean-edouard:
LGTM, merging soon.

eric-ch:
Building: [$6531](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6531]

PR #1142, opened 2019-04-25
OXT-1565: base-files: /var/lock -> /run/lock

eric-ch:
In openxt-installer and other MACHINE that do not have a specific /etc/default/volatile:
fstab.early already mounts a tmpfs on /run.
/var/lock is a symlink to /run/lock (created with populate-volatile.sh), so trying to mount a separate tmpfs on /var/lock will fail with:
> Mounting early filesystems...
> mount: mount point /var/lock is a symbolic link to nowhere

Avoid machine specific configuration.

/var/lock used to be a 1M partition, make it a symlink an simplify things.

Keeping /var/lock separated would avoid a rogue software to fill up /run in a DoS attempt, but things that have access to /var/lock probably already have access to /run (or subdirectory) in UGO perms or SELinux policy.

### DIFF
--- a/recipes-core/base-files/files/fstab.early
+++ b/recipes-core/base-files/files/fstab.early
@@ -13,7 +13,6 @@ tmpfs       /tmp                    tmpfs       defaults,rootcontext=system_u:ob
 
 tmpfs       /var/volatile           tmpfs       defaults,rootcontext=system_u:object_r:var_t:s0,size=2M         0 0
 tmpfs       /var/cache              tmpfs       defaults,rootcontext=system_u:object_r:var_t:s0,size=100M       0 0
-tmpfs       /var/lock               tmpfs       defaults,rootcontext=system_u:object_r:var_lock_t:s0,size=1M    0 0
 
 # OpenXT: modutils.sh loads before mountall.sh, and modutils.sh will load the txt module, which needs this:
-securityfs                     /sys/kernel/security    securityfs  defaults                                                                        0 0
+securityfs  /sys/kernel/security    securityfs  defaults                                                        0 0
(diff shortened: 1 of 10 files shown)


jean-edouard:

jean-edouard:
With this change, `/var/lock/asound.state.lock` and `/var/lock/lvm/` are missing.

eric-ch:
@jean-edouard `/run/lock` (target of `/var/lock`) is created by `S37populate-volatile.sh` now, instead of  `/var/lock` being the mount point of the `tmpfs` setup during `S01mountearly`. The lock files are created by `alsactl` and `lvm-utils` commands being used by udev rules (those in `/lib/udev/rules.d`), which are triggered much earlier (`S04udev`).

The path to lock files could be changed per package, but it's less than ideal.
I don't really like `mountearly` so a `populate-volatile-early` doesn't look good either.

Instead, the `udev` initscript is already creating `/var/volatile/tmp` for that reason and run `restorecon` on `/run` as well, it might create `/run/lock` while at it.

Does anyone have a better approach?

jandryuk:
You could just modify S04udev like you say.  Or create S03lock or something to create it without modifying udev's script.

Another option would be to run volatiles before udev.  But, that makes me uneasy since it's deviating from OE's setup.

Maybe modifying udev and upstreaming that is the right thing since others could hit it?

eric-ch:
> run volatiles before udev

Assuming some volatiles are on an encrypted partition, `udev` might be required already and, as you said, deviating from OE means slightly more misery for the future.

> modify S04udev

Actually I now think this is not such a good idea. That file is already replaced by `meta-selinux`, which means even more deviation creating our own.

> S03lock

Kinda sounds like `populate-volatile-early`, but more targeted. It feels like the best compromise. Thanks!

eric-ch:
### v2
- Rebased on master,, this was a bit stale.
- Add the udev-volatiles initscripts to work-around population of early tmpfs.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6594

jean-edouard:
This works fine, I was just worried about the SELinux implications of /var/lock being a symlink, since that's a separate permission target.
However, that's a pretty standard layout, /var/lock is a symlink on my Debian dev machine as well.
Also I didn't run into any AVC while testing this, so I'd say it's good to go.
Merging soon.

PR #1143, opened 2019-04-26
OXT-1571: [libxl] Retry on pci add failure for PT GPU

crogers1:
Recent changes in the platform have exposed a race condition
  between libxl and the stubdomain. When libxl sends the qmp command
  for "device_add" to QEMU, pcifront has not yet finished setting
  up the sysfs entries for the passthrough GPU. This causes QEMU
  to fail to find these entries, one example is

  "received an error message from QMP server: Could not open
  '/sys/bus/pci/devices/0000:01:00.0/config': No such file or directory"

  This patch introduces a retry loop with a short timeout to give
  pcifront some time to initialize the sysfs nodes before libxl
  decides to error and completely fail domain creation.

  OXT-1571

### DIFF
--- a/recipes-extended/xen/files/libxl-retry-qmp-pci-add.patch
+++ b/recipes-extended/xen/files/libxl-retry-qmp-pci-add.patch
@@ -1,0 +1,26 @@
+--- a/tools/libxl/libxl_qmp.c
++++ b/tools/libxl/libxl_qmp.c
+@@ -876,6 +876,7 @@ int libxl__qmp_pci_add(libxl__gc *gc, in
+     libxl__json_object *args = NULL;
+     char *hostaddr = NULL;
+     int rc = 0;
++    int i = 0;
+ 
+     qmp = libxl__qmp_initialize(gc, domid);
+     if (!qmp)
+@@ -907,8 +908,13 @@ int libxl__qmp_pci_add(libxl__gc *gc, in
+     if (pcidev->permissive)
+         qmp_parameters_add_bool(gc, &args, "permissive", true);
+ 
+-    rc = qmp_synchronous_send(qmp, "device_add", args,
+-                              NULL, NULL, qmp->timeout);
++    do {
++        rc = qmp_synchronous_send(qmp, "device_add", args,
++                                  NULL, NULL, qmp->timeout);
++        if (rc)
++            sleep(1);
++    } while (rc && (++i < 15));
++
+     if (rc == 0) {
+         rc = qmp_synchronous_send(qmp, "query-pci", NULL,
+                                   pci_add_callback, pcidev, qmp->timeout);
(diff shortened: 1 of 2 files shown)


crogers1:
This was tested successfully with windows 10 tools, Nvidia Quadro K2000 GPU passthrough.

jandryuk:
Did you consider a watch on the sub-device state in XenStore?  It may be a pain since PCI has those weird sub-devices.  You'd have to match the BDF to determine the sub-device number and then check the state-N key.  That's assuming that state is updated in a way that indicates the sysfs nodes are ready.

crogers1:
@jandryuk it's tough to tell just by watching and logs, would need to test it out.

jean-edouard:

recipes-extended/xen/files/libxl-retry-qmp-pci-add.patch
```diff
@@ -0,0 +1,28 @@
...
++    int i = 0;
+ 
+     qmp = libxl__qmp_initialize(gc, domid);
+     if (!qmp)
+@@ -880,8 +881,17 @@ int libxl__qmp_pci_add(libxl__gc *gc, in
+     if (pcidev->permissive)
+         qmp_parameters_add_bool(gc, &args, "permissive", true);
+ 
++    
+     rc = qmp_synchronous_send(qmp, "device_add", args,
```
jean-edouard: IMO this call should be replaced by the loop, instead of being duplicated there.

crogers1: I'd have to initialize rc to something non-zero or else I have no way to get in the loop

jean-edouard: Remove this line and the `if` below, swap the sleep and the qmp call, and done :)

crogers1:
Broke something with that push, hold on

crogers1:
Updated and tested with Jed's suggestion to not duplicate the qmp_synchronous_send call

recipes-extended/xen/files/libxl-retry-qmp-pci-add.patch
```diff
@@ -0,0 +1,25 @@
...
+@@ -907,8 +908,12 @@ int libxl__qmp_pci_add(libxl__gc *gc, in
+     if (pcidev->permissive)
+         qmp_parameters_add_bool(gc, &args, "permissive", true);
+ 
+-    rc = qmp_synchronous_send(qmp, "device_add", args,
+-                              NULL, NULL, qmp->timeout);
++    do {
++        rc = qmp_synchronous_send(qmp, "device_add", args,
++                                  NULL, NULL, qmp->timeout);
++        sleep(1);
```
jandryuk: This will always sleep.
```
if (rc)
    sleep(1);
```
To avoid it on success.

Have you measured the maximum number of tries you've needed for the problematic GPU?  I wonder about shortening the sleep below 1 second and potentially increasing the maximum iterations.

crogers1: On my box it takes 1 try before it succeeds.  Will update with the if (rc) to prevent sleep on success for first iteration

jean-edouard:
Merging soon.

PR #1145, opened 2019-04-26
STABLE-9: Fix forward seal: Revert "lvm udev rules: avoid trying to create bad …

jandryuk:
…symlinks"

This reverts commit aa1aea57a722c6e5740922e5d0be770b3605609d.

"lvm udev rules: avoid trying to create bad symlinks" broke forward
sealing during upgrade.  The forward seal was performed against root.old
since the symlinks weren't updated properly.

The upgrade LV operations are roughly:
lvremove root.old
lvcreate root.new
lvrename root root.old
lvrename root.new root

At the point of forward seal, we mount /dev/xenclient/root and chroot
in.  Here is the ls output at that time:

brw-------. 1 root root 253, 7 Apr 26 11:45 /dev/mapper/xenclient-root
brw-------. 1 root root 253, 6 Apr 26 11:45 /dev/mapper/xenclient-root.old
lrwxrwxrwx. 1 root root      7 Apr 26 12:27 /dev/xenclient/root -> ../dm-6
lrwxrwxrwx. 1 root root      7 Apr 26 12:27 /dev/xenclient/root.old -> ../dm-7

/dev/xenclient/root still points to dm-6 which is actually root.old.

Reverting this patch lets udev overwrite the device nodes so that
/dev/xenclient/root properly symlinks to dm-7.

They start out as:
lrwxrwxrwx. 1 root root      7 Apr 26 12:13 /dev/xenclient/root -> ../dm-6
lrwxrwxrwx. 1 root root      7 Apr 26 12:13 /dev/xenclient/root.old -> ../dm-7
and are then updated appropriately:
lrwxrwxrwx. 1 root root      7 Apr 26 12:33 /dev/xenclient/root -> ../dm-7
lrwxrwxrwx. 1 root root      7 Apr 26 12:33 /dev/xenclient/root.old -> ../dm-6

(cherry picked from commit 7b18e8f619eb3f717f5d1457e5ec1b6a1de169ba)


jean-edouard:
Merging soon.

PR #1146, opened 2019-04-26
[seal-system] Use dd and iflag=direct

crogers1:
Calling sha256sum on the root_dev for computing pcr15 hash
  can fail to compute the correct hash if there are cached writes
  hanging about. The 'direct' flag bypasses the buffer cache
  to read the root dev accurately.

  OXT-1583

### DIFF
--- a/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
+++ b/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
@@ -251,7 +251,12 @@ forward)
     # During early init, rootfs is hashed and is given to
     # tpm_extend which for tpm1.2 hashes the rootfs hash again and
     # hands that value to the TPM to be extended into PCR 15
-    root_hash=$(${hashalg}sum ${root_dev}|cut -f1 -d' ')
+
+    # Use dd here to perform direct I/O. On some machines in some cases, stale data in
+    # the storage layer, or perhaps the disk cache, causes sha256sum {root_dev} to compute
+    # an incorrect hash, leading to a bad PCR15 value. Direct I/O always computes an accurate hash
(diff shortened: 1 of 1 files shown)


jandryuk:
Is this PR addressing something seen in a normal OTA or under testing when manually changing the rootfs - i.e. `rw ; touch /foo ; ro`?

The commit messages says you want to avoid the buffer cache, but don't you need to take all pending writes into account?  Does `iflag=direct` bypassing the cache or force coherency by forcing the writeback of outstanding data?  Is `sync` sufficient to force consistency?

crogers1:
This is something found while manually changing the rootfs, as in your example.

I'm double checking ```sync```, but I think it's not enough in this case. I'm also trying to find more documentation about direct I/O as it relates to reads with dd.  "bypassing" I suppose isn't quite right, if we're getting an accurate measure of the drive by using it.  Might take a little time.

crogers1:
@jandryuk it still repros with ```sync```

crogers1:
There must be something stale in the read cache causing sha256sum to produce a bad checksum, yet direct IO with ```dd``` to succeed.

jandryuk:
Thanks for checking `sync`.

I think it would be the other way - not something stale in the read cache, but something stale (un-written) on the block device.  The storage stack is something like VFS, ext4, LVM, sda.  Seems to me the VFS layer could have new data, but hashing the block device gives the old data.

How `iflags=direct` makes it consistent - I don't know.  It does seem to slow the hashing down a little - 5s vs. 6.5s in a quick test - which gives more time for a pending write to be flushed.  Why did you think to try it?

Anyway, a code comment explaining the non-obvious use of dd would be good.

jean-edouard:

jean-edouard:

jean-edouard:
Merging soon.

PR #1147, opened 2019-04-29
OXT-1574: [xen] Drop hide-cores-from-cpuid.patch

crogers1:
This patch was preventing Windows 10 guests from seeing an accurate
  count of vcpus assigned. It was originally intended for XP
  support, but now that XP is officially end of life and breaking
  functionality in a modern OS that is supported, this patch
  should be dropped.

  OXT-1574


jean-edouard:
+1 for dropping hide-cores-from-cpuid.patch
However, I like having the other 2 debug patches included in every build. Obviously if there's a good reason to drop them then so be it.

jandryuk:
Like @jean-edouard, +1 for dropping hide-cores-from-cpuid.patch

For the other patches, I'd prefer them not to be left dangling in the repo.  Either delete them, or leave them in SRC_URI.  A third option would be to conditionally include them in debug builds, but I'm not sure if that's worth it.

crogers1:
Ok, I didn't have a strong preference either way, I'll re-enable them in the SRC_URI and do a build quick.

jean-edouard:

jean-edouard:
Merging soon.

PR #1148, opened 2019-04-29
STABLE-9: OXT-1583: [seal-system] Use dd and iflag=direct

crogers1:
Calling sha256sum on the root_dev for computing pcr15 hash
  can fail to compute the correct hash if there are cached writes
  hanging about. The 'direct' flag bypasses the buffer cache
  to read the root dev accurately.

  OXT-1583

### DIFF
--- a/recipes-core/images/xenclient-upgrade-compat-image.bb
+++ b/recipes-core/images/xenclient-upgrade-compat-image.bb
@@ -11,6 +11,7 @@ IMAGE_FSTYPES = "ext3.gz"
 export IMAGE_BASENAME = "xenclient-upgrade-compat-image"
 IMAGE_INSTALL = " \
     busybox \
+    coreutils \
     lvm2 \
     grub \
     policycoreutils \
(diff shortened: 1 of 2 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6534

jean-edouard:
That won't work in the case of upgrade-compat, since it uses dd from busybox (instead of coreutils) which doesn't support these options.

crogers1:
Included an additional commit that adds coreutils to the upgrade-compat image so seal-system can use a version of dd that supports iflag options

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6540

jean-edouard:
LGTM, merging soon.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6540

PR #1150, opened 2019-05-01
STABLE-9: OXT-1538: refpolicy: Silence mount AVC

jandryuk:
Stable-9 version of https://github.com/OpenXT/xenclient-oe/pull/1149

The changes for xenmgr to use losetup has triggered an AVC when calling
mount on /dev/loop0p1:

audit: type=1400 audit(1556562031.429:4): avc: denied { read } for pid=1200 comm="mount" name="loop0p1" dev="devtmpfs" ino=14478 scontext=system_u:system_r:mount_t:s0-s0:c0.c1023 tcontext=system_u:object_r:device_t:s0 tclass=blk_file permissive=0

This seems to be a race with udev labeling loop0p1 to
fixed_disk_device_t, but mount succeeds anyway.

Add a dontaudit to eliminate the noise. mount already has a few rules
to deal with device_t devtmpfs files. There isn't an existing interface to
dontaudit device_t:blk_file, so I just open-coded it.

To minimize churn, only the mount.te portions of the refpolicy-mcs patchqueue were refreshed. This means the patches have a combination of default Quilt and OpenXT Quilt formats. I can refresh the entire patches if desired.

OXT-1538

Fixes AVC from: https://github.com/OpenXT/manager/pull/140

### DIFF
--- a/recipes-security/refpolicy/refpolicy-mcs-2.%/patches/blktap-interfaces.diff
+++ b/recipes-security/refpolicy/refpolicy-mcs-2.%/patches/blktap-interfaces.diff
@@ -64,11 +64,9 @@ Index: refpolicy/policy/modules/system/lvm.te
  	bootloader_rw_tmp_files(lvm_t)
  ')
  
-Index: refpolicy/policy/modules/system/mount.te
-===================================================================
---- refpolicy.orig/policy/modules/system/mount.te
-+++ refpolicy/policy/modules/system/mount.te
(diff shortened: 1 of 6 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6540

jean-edouard:
LGTM, merging soon.

PR #1151, opened 2019-05-02
OXT-1594: tpm2: Store pcr configuration in ${tss}.pcrs

jandryuk:
This is just the PCR consolidation commit from: https://github.com/OpenXT/xenclient-oe/pull/1108
Including this will not change the PCR selection, but it allows changing the selection.

Instead of hardcoding the pcr configuration into scripts, store the
configuration into the ${tss}.pcrs file.  This way we remove hardcoding
in our scripts.  Also, it will enable the tpm-quirks functionality to
tweak the pcr config.

TPM 1.2 sealing structure incorporates the PCR configuration.  Not sure
why TPM 2.0 does not.  Anyway, it's not sensitive, so it's okay to
store.  If an attacker changes it, they still cannot unseal.

### DIFF
--- a/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
+++ b/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
@@ -395,7 +395,7 @@ encrypted_unlock() {
     then
         if pcr_bank_exists "sha256"; then
             local unseal_file=${key_file}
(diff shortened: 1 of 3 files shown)


stacktrust:
Please add this ticket number to the PR:

https://openxt.atlassian.net/browse/OXT-1594

jean-edouard:

jandryuk:
force-pushed to add ticket.

jandryuk:
This PR is broken.  seal-system hard-codes its list of PCRs, so it doesn't follow /config/config.pcrs.

PR #1152, opened 2019-05-03
openxt_image_types: Support ext4 disk images

jandryuk:
IMAGE_CMD_ext3 calls oe_mkext234fs and then our _append commands disable
future file system checks.  There is nothing ext3 specific here.

Change the _append from IMAGE_CMD to oe_mkext234fs so any of ext2, 3 & 4
are usable for VM images.

NOTE: this does not change any disk formats.  I've only tested (successfully) with some custom images and not tried converting any default openxt images.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6541

jean-edouard:
LGTM, merging soon.

PR #1153, opened 2019-05-06
OXT-1593: [init] Load i915 driver to accommodate multiple

crogers1:
efifb devices.  This successfully shows the ncurses measurement
  failure screens on boot.

  OXT-1593


eric-ch:
On the side of this PR: Could the `init.root-ro` be moved to early initscripts that would be on the bootsrap list if/when SELinux/Sealing/Read-only are enabled? This would simplify everything, but I have a feeling I am missing a point in there...

jean-edouard:
Merging soon.

PR #1155, opened 2019-05-06
STABLE-9: OXT-1574: [xen] Drop hide-cores-from-cpuid.patch

crogers1:
This patch was preventing Windows 10 guests from seeing an accurate
  count of vcpus assigned. It was originally intended for XP
  support, but now that XP is officially end of life and breaking
  functionality in a modern OS that is supported, this patch
  should be dropped.

  OXT-1574


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6546

jean-edouard:
Merging soon.

PR #1156, opened 2019-05-06
passwd: set all non-root shells to false

jean-edouard:
OXT-1585
OXT-1586


eric-ch:
@jean-edouard @crogers1 stable-9?

jean-edouard:
https://github.com/OpenXT/xenclient-oe/pull/1174

PR #1157, opened 2019-05-07
STABLE-9: OXT-1571: [libxl] Retry on pci add failure for PT GPU

crogers1:
Recent changes in the platform have exposed a race condition
  between libxl and the stubdomain. When libxl sends the qmp command
  for "device_add" to QEMU, pcifront has not yet finished setting
  up the sysfs entries for the passthrough GPU. This causes QEMU
  to fail to find these entries, one example is

  "received an error message from QMP server: Could not open
  '/sys/bus/pci/devices/0000:01:00.0/config': No such file or directory"

  This patch introduces a retry loop with a short timeout to give
  pcifront some time to initialize the sysfs nodes.

  OXT-1571


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6550

jean-edouard:
LGTM, merging soon.

PR #1159, opened 2019-05-09
kernel: enable cgroups for dom0

jean-edouard:
OXT-1379


jandryuk:
What about dropping the revert patch and just mounting the memory cgroup at '/sys/fs/cgroup/memory'?

blktap will use the cgroup in dom0.  Is there a use for uivm or ndvm?

jean-edouard:
Good questions!

About the first one, I would have to first mount a tmpfs in /sys/fs/cgroup, then mkdir /sys/fs/cgroup/memory, then mount the cgroup there, which can't all be done in fstab and would have to be hardcoded somewhere like mountall.sh. Also that's just not what's usually done out there. You either have systemd and the separate dirs, or sysvinit and the one directory.

About the second one, I just followed what the ticket said, but you're right, this is probably not really needed in service VMs. I'll ask the author @eric-ch about it :)

jandryuk:
Makes sense on the systemd vs. sysvinit cgroups.  If more software starts requiring the systemd layout, then we may want to match that layout in the future.

I think you want to split your commit into 2 or 3.  One enables cgroups in the kernel, and the second (or rolled into the previous one) adds the fstab entry.  That would be OXT-1379.  The other commit would change blktap under OXT-1312.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6555

eric-ch:
@jean-edouard The ticket was a blanket statement. I do not think there is anything using cgroups in UIVM or NDVM currently.

jean-edouard:
Fixed, dom0 only now.

eric-ch:
Build [$6596](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6596)

jean-edouard:
Do not merge, AVCs spotted...

jean-edouard:
Tapdisk AVCs fixed.
For the mount AVCs, see https://github.com/OpenXT/xenclient-oe/pull/1149#issuecomment-501381058

eric-ch:
Build [$6607](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6607)

eric-ch:
Will merge soon.

@crogers1 Is this something you want in stable-9?

crogers1:
@eric-ch sure, let's bring it into stable-9

PR #1160, opened 2019-05-13
OXT-1374: [xen] Remove Kconfig option pv linear-pagetables

tijko:
Recent versions of Xen allow for the removal of this feature which
defaults to being turned on.  Since this feature isn't used by Linux
removing it reduces the security footprint by removing unnecessary code.

Frome [xen/arch/x86/Kconfig](https://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=xen/arch/x86/Kconfig;h=5c2d1070b657635b9fb959fb0c69e8cf4a685307;hb=997d6248a9ae932d0dbaac8d8755c2b15fec25dc):

```
config PV
	def_bool y
	prompt "PV support"
	---help---
	  Interfaces to support PV domains. These require guest kernel support
	  to run as a PV guest, but don't require any specific hardware support.

	  This option is needed if you want to run PV domains.

	  If unsure, say Y.

config PV_LINEAR_PT
       bool "Support for PV linear pagetables"
       depends on PV
       default y
       ---help---
         Linear pagetables (also called "recursive pagetables") refers
         to the practice of a guest operating system having pagetable
         entries pointing to other pagetables of the same level (i.e.,
         allowing L2 PTEs to point to other L2 pages).  Some operating
         systems use it as a simple way to consistently map the current
         process's pagetables into its own virtual address space.

         Linux and MiniOS don't use this technique.  NetBSD and Novell
         Netware do; there may be other custom operating systems which
         do.  If you're certain you don't plan on having PV guests
         which use this feature, turning it off can reduce the attack
         surface.

         If unsure, say Y.

config HVM
	def_bool !PV_SHIM_EXCLUSIVE
	prompt "HVM support"
	---help---
	  Interfaces to support HVM domains.  HVM domains require hardware
	  virtualisation extensions (e.g. Intel VT-x, AMD SVM), but can boot
	  guests which have no specific Xen knowledge.

	  This option is needed if you want to run HVM or PVH domains.

	  If unsure, say Y.
```
Signed-off-by: Tim Konick <konickt@ainfosec.com>


eric-ch:
Build: [$6556](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6556)

eric-ch:
@tijko Can you submit a stable-9 PR please?

tijko:
@eric-ch opened: https://github.com/OpenXT/xenclient-oe/pull/1172

PR #1173, opened 2019-05-24
S9: OXT-1594: Forward Seal according to /config/config.pcrs

jandryuk:
This is the stable-9 version of https://github.com/OpenXT/xenclient-oe/pull/1158

This PR is groundwork for OXT-1594. It makes seal-system forward seal according to /config/config.pcrs. It does not change the pcr selection.

It is now flexible and will seal against any selection of PCRs. If the selection set is empty, then seal-system fails (seal-system: forward seal of key failed) since tpm2_sealdata fails (ERROR: Unable to run tpm2_sealdata). We can also drop out PCR 15 which has the rootfs measurement. Is that too flexible?

There are also some other cleanups I noticed along the way.


eric-ch:
Will merge soon.

PR #1178, opened 2019-06-11
S9: OXT-1565: base-files: /var/lock -> /run/lock

eric-ch:
In openxt-installer and other MACHINE that do not have a specific /etc/default/volatile:
fstab.early already mounts a tmpfs on /run.
/var/lock is a symlink to /run/lock (created with populate-volatile.sh), so trying to mount a separate tmpfs on /var/lock will fail with:
> Mounting early filesystems...
> mount: mount point /var/lock is a symbolic link to nowhere

Avoid machine specific configuration.

/var/lock used to be a 1M partition, make it a symlink an simplify things.

Keeping /var/lock separated would avoid a rogue software to fill up /run in a DoS attempt, but things that have access to /var/lock probably already have access to /run (or subdirectory) in UGO perms or SELinux policy.

Master PR: https://github.com/OpenXT/xenclient-oe/pull/1142


jean-edouard:
Merging soon.

PR #1179, opened 2019-06-11
S9 stubdom tap backend

tijko:


tijko:
> Is there a benefit for using blktap for partitions? It still seems to me that making blkback work in openxt would be better since that doesn't use the tapdisk processes.

I'm unsure of which would be a better fit but I'm testing this now and will provide any additional insight on how each method operates.

eric-ch:
This is duplicated by https://github.com/OpenXT/xenclient-oe/pull/1255. Closing.

PR #1180, opened 2019-06-12
OXT-1494: [xen] Force synchronous page scrub

crogers1:
Recent versions of Xen have reworked page scrubbing so that it
  is now asynchronous, called from the idle loop. For client
  virtualization, this is a concern since it is possible for Xen
  to never schedule the loop to run under the correct conditions,
  leaving unscrubbed pages from old domains hanging about. While
  Xen will scrub pages before giving them to a new domain, this
  in-between state where scrubbing is not guaranteed is a concern.

  This commit reintroduces page scrubbing synchronously when a domain
  is shutdown, and uses the bootscrub=1 flag to scrub pages on
  platform boot.

  OXT-1494

### DIFF
--- a/recipes-extended/xen/xen-common.inc
+++ b/recipes-extended/xen/xen-common.inc
@@ -99,6 +99,7 @@ SRC_URI_append = " \
     file://argo-fix-full-ring-write-bug.patch \
     file://argo-fix-requeue-msg-len-bug.patch \
     file://libxl-seabios-ipxe.patch \
+    file://memory-scrub-on-domain-shutdown.patch \
 "
 
 COMPATIBLE_HOST = 'i686-oe-linux|(x86_64.*).*-linux|aarch64.*-linux'
--- a/recipes-extended/xen/files/memory-scrub-on-domain-shutdown.patch
+++ b/recipes-extended/xen/files/memory-scrub-on-domain-shutdown.patch
@@ -1,0 +1,83 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Force page scrub after domain shutdown, not during idle loop.
+
+################################################################################
+LONG DESCRIPTION:
+################################################################################
+In recent Xen versions, page scrubbing has been reworked to occur asynchronously
+during an idle loop. Therefore, it is possible that pages once owned by a guest
+will never have their pages scrubbed if Xen never schedules the loop to run. Xen
+WILL scrub pages before given to a newly booted guest, but during the time the 
+guest is off, it's possible to have unscrubbed pages from previous guests
+hanging about.
+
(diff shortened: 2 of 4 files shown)


crogers1:
@dozylynx @eric-ch @jean-edouard

jean-edouard:
Looks fine, except for the indentation.

recipes-extended/xen/files/memory-scrub-on-shutdown.patch
```diff
@@ -0,0 +1,82 @@
...
+################################################################################
+REMOVAL
+################################################################################
+Never
+
+################################################################################
+UPSTREAM PLAN
+################################################################################
+None, synchronous page scrubbing seems to be a client-side concern, Xen seems
+more concerned with the performance benefit of asynchronous scrubbing.
```
dozylynx: This statement needs amending. The concern about scrubbing was raised and discussed on the Xen Community Call; there is awareness of the problem upstream and the committers stated that code to enable selection of the scrubbing behaviour would be welcome.

ie. An option to enable the defensive synchronous scrubbing that OpenXT requires, whilst allowing other Xen users -- particularly those on systems with large amounts of memory that are performing operations such as tests that frequently tear down VMs with lots of memory (think: a database company) -- to retain the background scrub-when-idle option.

recipes-extended/xen/xen-common.inc
```diff
@@ -94,6 +94,7 @@ SRC_URI_append = " \
     file://v4v.patch \
     file://v4v-viptables.patch \
     file://openxt-xen-xsmv4vuse.patch \
+    file://memory-scrub-on-shutdown.patch \
```
dozylynx: suggest naming: `memory-scrub-on-domain-shutdown.patch`
just to make it clearer that this isn't only about host shutdown.

recipes-extended/xen/files/memory-scrub-on-shutdown.patch
crogers1: Ah, I was unaware this was discussed upstream.  Will amend.

crogers1:
Amended Upstream Plan section, renamed patch to indicate it's domain shutdown we're concerned with, fixed indentation.

eric-ch:
Build [$6610](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6610)

crogers1:
Rebased on argo commits

eric-ch:
Build [$6618](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6618)

jean-edouard:
LGTM, merging soon.

crogers1:
Resolved patch conflict

jean-edouard:
Merging soon

PR #1181, opened 2019-06-12
STABLE-9: OXT-1494: [xen] Force synchronous page scrub

crogers1:
Recent versions of Xen have reworked page scrubbing so that it
  is now asynchronous, called from the idle loop. For client
  virtualization, this is a concern since it is possible for Xen
  to never schedule the loop to run under the correct conditions,
  leaving unscrubbed pages from old domains hanging about. While
  Xen will scrub pages before giving them to a new domain, this
  in-between state where scrubbing is not guaranteed is a concern.

  This commit reintroduces page scrubbing synchronously when a domain
  is shutdown, and uses the bootscrub=1 flag to scrub pages on
  platform boot.

  OXT-1494

(cherry picked from commit 58c4ab989f6d457c6a195276d5621840a07a15c0)

### DIFF
--- a/recipes-extended/xen/files/memory-scrub-on-domain-shutdown.patch
+++ b/recipes-extended/xen/files/memory-scrub-on-domain-shutdown.patch
@@ -1,0 +1,83 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Force page scrub after domain shutdown, not during idle loop.
+
+################################################################################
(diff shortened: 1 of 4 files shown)


crogers1:
Updated based on master PR comments

jean-edouard:
Merging soon.

jean-edouard:
Merging soon.

PR #1182, opened 2019-06-12
OXT-839: KCONFIG grant table patches and a new defconfig for Xen 4.12

dozylynx:
The three new patches enable fine-grained compile-time control over the grant table code, in order to reduce its attack surface.

1)  kconfig-grant-table.patch

The first patch is a backport from upstream Xen, to introduce control over inclusion of any of the grant table code at all. OpenXT does require the grant table code to be included so this option must be enabled. This is a prerequisite patch for the following two patches.

2) kconfig-grant-table-v2-interface.patch

This patch introduces control over the v2 grant table interface.
OpenXT does not require this interface and the option should be disabled.

3) kconfig-grant-table-exotic.patch

This patch introduces control over several legacy grant table operations in the v1 interface that are not in use by the modern Linux PV drivers.
OpenXT does not require this interface and the option should be disabled.

Notes on the updated defconfig for Xen 4.12:

Credit2 is now the default scheduler for Xen 4.12, so enable that and for minimization of attack surface remove all the other schedulers.

Argo is enabled.

The supported number of physical CPUs has been decreased from 256 to 32.
This reduces memory consumption and aligns with the default value used by the PV shim.

The virtual machine introspection interface (which is not security supported by upstream Xen) is disabled. So are MEM_PAGING and MEM_SHARING.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6612

jean-edouard:
LGTM, merging soon.

PR #1186, opened 2019-06-17
OXT-1626: Update Xen with current Linux's list macros

dozylynx:
Add patch to switch Xen to use modern Linux's list macros and simplify
tracking upstream Linux's header files involved.

Xen's list headers were forked from Linux a long time ago and at this point
are missing significant updates. Bring in the current Linux files and apply
only the minimum necessary changes to them to enable use within Xen.

This change separates out the RCU list functions into a separate header,
rculist.h, as per Linux. Likewise the poison values are moved into a
separate header, as per Linux, so that they are available to both list.h
and rculist.h.

Xen's unique additions to its version of list.h are moved into a
new separate header, list_xen.h, in order to simplify keeping the contents
of Xen's list.h exactly in sync with Linux's list.h.

Necessary macros are imported from Linux's compiler.h into Xen's, with
minor changes to avoid ratholing into causing further imports.

Linux removed the list_for_each_rcu macro, replacing all uses of it with
list_for_each_entry_rcu, so do the same to the one call site Xen has
within arch/x86/mm/mem_sharing.c.

The hlist macros also changed to drop an unnecessary argument, so
apply that change too -- only affects xsm/flask/avc.c.

### DIFF
--- a/recipes-extended/xen/files/unfork-linux-lists.patch
+++ b/recipes-extended/xen/files/unfork-linux-lists.patch
@@ -1,0 +1,2738 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Replace Xen's list macros that were forked from Linux a long time ago with same
+from current Linux where they have been actively maintained.
+
+################################################################################
+LONG DESCRIPTION:
+################################################################################
+
+################################################################################
+CHANGELOG
+################################################################################
+Patch written by Christopher Clark for OpenXT 9.0.
+Developed against upstream staging (pre-Xen 4.13) and cherry-picked to Xen 4.12.
+
+################################################################################
+REMOVAL
+################################################################################
+Only if upstreamed. Do not remove.
+
+################################################################################
+UPSTREAM PLAN
+################################################################################
+This change belongs in upstream Xen.
+Applied to OpenXT to enable inclusion in the 9.0 release.
+
+If the list macro files change in Linux, this patch should track the changes:
+Linux is the correct upstream for these list macros.
+
+################################################################################
+INTERNAL DEPENDENCIES
+################################################################################
+None.
+
+################################################################################
+PATCHES
+################################################################################
+From 82d068fbe05af0ef76f720db312bdf61b36bb80b Mon Sep 17 00:00:00 2001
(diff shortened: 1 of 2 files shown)


dpsmith:
Haven't reviewed yet, but a small nit/personal opinion from the description. You aligned with Linux naming convention for Linux derived files but then deviated for the Xen specific. Would it not make sense to name the Xen list header, xenlist.h?

jandryuk:
What is the tangible benefit of this change for OpenXT?

Aligning the macros is a worth goal, but it seems to me this patch should go into upstream Xen.  OpenXT will pick it up with a Xen uprev.  As it is, I don't see a rationale for a backport.

eric-ch:
Build [$6619](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6619)

dozylynx:
> Haven't reviewed yet, but a small nit/personal opinion from the description. You aligned with Linux naming convention for Linux derived files but then deviated for the Xen specific. Would it not make sense to name the Xen list header, xenlist.h?

@dpsmith : I have no strong attachment to the `list_xen.h` header filename, am happy to rename it, so I've just refreshed it to use `xenlist.h` instead.

stacktrust:
> What is the tangible benefit of this change for OpenXT?
> 
> Aligning the macros is a worth goal, but it seems to me this patch should go into upstream Xen. OpenXT will pick it up with a Xen uprev. As it is, I don't see a rationale for a backport.

@jandryuk Upstream discussions are linked on https://openxt.atlassian.net/browse/OXT-1625.  In summary, a new process is being developed for review of upstream files.  Due to resource constraints, this will be applied during the next (Xen 4.13) release cycle.  Now that OpenXT has moved to the latest Xen (4.12) release, OpenXT test results (e.g. by moving to actively maintained macros) are potentially valuable to the upstream community.

As discussed on the June 20th OpenXT community call, I recommend we move to the macros maintained by upstream Linux and include them in the next stable-9 test cycle.   Once we have confidence in the changes, we can share our approach and results with upstream Xen.

jandryuk:
Superseeded by https://github.com/OpenXT/xenclient-oe/pull/1461 as the Xen uprev updated the macros.

PR #1188, opened 2019-06-19
OXT-1627: [hvmloader] Pass ipxe ROM to hvmloader

crogers1:
With Xen 4.12, etherboot is no longer built with hvmloader.
  The expectation is to use a separate ipxe ROM and pass it along
  to hvmloader via libxl. This patch completes the codepath for
  seabios so that we can pass along a ROM from ipxe.bb to use
  for guest pxe booting.

  OXT-1627


crogers1:
Stable-9 PR to follow once argo is merged into that baseline.

eric-ch:
Build [$6618](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6618)

jean-edouard:
`82540em.rom` also seems to work for 82544gc, 82545em and 82574l(e1000e).
However, it (obviously) won't work for rtl8139. We used to use that NIC in some situations, but it doesn't seem to be the case anymore.
Anyway, LGTM, merging soon.

PR #1190, opened 2019-06-20
OXT-1623: libxl: add back stubdom check for helper kill

mahantesh-ais:
OXT-1623

This check was missed when upgrading from xen-4.11

### DIFF
--- a/recipes-extended/xen/files/libxl-disable-json-updates.patch
+++ b/recipes-extended/xen/files/libxl-disable-json-updates.patch
@@ -49,7 +49,7 @@ PATCHES
 ################################################################################
 --- a/tools/libxl/libxl_internal.h
 +++ b/tools/libxl/libxl_internal.h
-@@ -3604,7 +3604,7 @@ _hidden void libxl__bootloader_run(libxl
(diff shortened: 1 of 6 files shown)


eric-ch:
Build [$6625](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6625)

eric-ch:
So the warning is no longer there for dom0, but instead I have that same warning back at shutdown for every domain with stub-domain.

eric-ch:
Last warning is triggered by platforms that don't have an ATAPI device to pass-through as the only criteria for this log is to try to read a Xenstore node . Not a regression.

jean-edouard:
@mahantesh-ais, is the new change another thing that was lost in upgrade? Does that change apply to the stable-9 PR too?

mahantesh-ais:
@jean-edouard, new commit is suggested by @eric-ch, which is to log when `kill()` fails rather than logging when read on Xenstore node fails. If all agree with this change, I can PR the change to stable-9 as well.

eric-ch:
Build [$6631](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6631)

eric-ch:
No more warnings, thanks!

PR #1191, opened 2019-06-20
S9: OXT-1623: libxl: add back stubdom check for helper kill

mahantesh-ais:
OXT-1623

This check was missed when upgrading from xen-4.11


mahantesh-ais:
Adding a new commit to reflect changes done in master branch.
new commit: log when kill() fails rather than logging when read on Xenstore node fails.

eric-ch:
Will merge soon.

PR #1192, opened 2019-06-21
STABLE-9: OXT-1627: [hvmloader] Pass ipxe ROM to hvmloader

crogers1:
With Xen 4.12, etherboot is no longer built with hvmloader.
  The expectation is to use a separate ipxe ROM and pass it along
  to hvmloader via libxl. This patch completes the codepath for
  seabios so that we can pass along a ROM from ipxe.bb to use
  for guest pxe booting.

  OXT-1627


jean-edouard:
Merging soon.

PR #1196, opened 2019-06-27
Cleanup real-time clock support

jandryuk:
Only the Dom0 and Installer machines have access to a hardware real-time clock.  By removing 'rtc' from MACHINE_FEATURES (through MACHINE_FEATURES_BACKFILL_CONSIDERED), we also remove the hwclock.sh initscript from the virtual machines.

We can then clean up some image recipes.

save-rtc.sh is redundant, so we can drop that as well.


eric-ch:
Build [$6634](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6634)

eric-ch:
`MACHINE_FEATURES_BACKFILL_CONSIDERED` is pretty useful.
I seem to recall we were looking for something like that with regard to other `FEATURES`, during or after the Pyro upgrade, I cannot remember which though, does that sound familiar?

jandryuk:
I was surprised I needed `MACHINE_FEATURES_BACKFILL_CONSIDERED` and `MACHINE_FEATURES_remove = "rtc"` didn't work.  I previously fought to remove rtc. and found `MACHINE_FEATURES_remove += "rtc"` with a `+=` worked (I think).

I didn't know about the `BACKFILL` stuff until I stumbled upon it in `bitbake -e` output.

Not sure about the other `FEATURES`.  Maybe it was a general through to review them?  I removed nfs from `DISTRO_FEATURES` to get rid of rpcbind.  Maybe you are thinking of that?

eric-ch:
Oh that must be `nfs` you are correct.

eric-ch:
Will merge soon.

@crogers1 This looks very low risk, do you want it in stable-9?

crogers1:
Yes, would be great to have in stable-9

PR #1197, opened 2019-06-29
OXT-1628: Uprev libtirpc to 1.1.4

rneilturner:
Signed-off-by: turnerr <turnerr@ainfosec.com>

### DIFF
--- a/recipes-extended/libtirpc/libtirpc_1.1.4.bb
+++ b/recipes-extended/libtirpc/libtirpc_1.1.4.bb
@@ -1,0 +1,25 @@
+SUMMARY = "Transport-Independent RPC library"
+DESCRIPTION = "Libtirpc is a port of Suns Transport-Independent RPC library to Linux"
+SECTION = "libs/network"
+HOMEPAGE = "http://sourceforge.net/projects/libtirpc/"
+BUGTRACKER = "http://sourceforge.net/tracker/?group_id=183075&atid=903784"
+LICENSE = "BSD"
+LIC_FILES_CHKSUM = "file://COPYING;md5=f835cce8852481e4b2bbbdd23b5e47f3 \
+                    file://src/netname.c;beginline=1;endline=27;md5=f8a8cd2cb25ac5aa16767364fb0e3c24"
+
+PROVIDES = "virtual/librpc"
+
+SRC_URI = "${SOURCEFORGE_MIRROR}/${BPN}/${BP}.tar.bz2"
+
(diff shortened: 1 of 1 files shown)


stacktrust:
If this is part of https://openxt.atlassian.net/browse/OXT-1628, please include the ticket number in the PR so that this PR is tracked for backports, future fixes, etc

dozylynx:
As a legal declaration, the `Signed-off-by` should specify your real name, not a handle, please.
Agree with request for a ticket number in the commit message too.

recipes-extended/libtirpc/libtirpc_%.bbappend
```diff
@@ -1 +0,0 @@
-CFLAGS += "-fPIC"
```
jandryuk: @rneilturner I have only compile tested, but it looks like leaving `CFLAGS += "-fPIC"` in this .bbappend enables PIC without needing your `compile-static-with-fPIC.patch`.  The option moves on the command line, but it's still there.

I prefer a separate .bbappend since it makes a clear distinction between our customization and the upstream recipe.  It also retains the history of why @Tmcg2 introduced the change.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6640

rneilturner:
I have been on vacation but will test out the suggested changes next week. Thanks everyone for the feedback.

eric-ch:
Amend change-set per comments:
- Re-introduce `CFLAGS += "-fPIC"` in `.bbappend` to keep local changes out of the upstream recipe.
- Amend commit-message.

jandryuk:
If you wouldn't mind expanding the commit message, please add:
> With pyicbinn linked with -Bsymbolic, sync-client segfaulted with:
> VM syncvm (5): [ 18.669233] python[555]: segfault at 0 ip 0000000000000000 sp 00007ffe5a6c1328 error 14 in python2.7[400000+1000]
> VM syncvm (5): [ 18.669247] Code: Bad RIP value.
> 
> Upreving libtirpc to 1.1.4 avoids the segfault.

I added similar to OXT-1628 if you just want to commit.

Locally, I grabbed the entire upstream libtirpc 1.1.4 recipe which includes a musl patch we don't need and two extra variables, UPSTREAM_CHECK_URI & UPSTREAM_CHECK_REGEX which I don't think we use.  I'm fine with this PR.

jean-edouard:
Merging soon.

PR #1202, opened 2019-07-05
OXT-1635: layer: Remove deprecated and/or unused recipes.

eric-ch:
- `dialog_1.2-20150225.bb+patches`: meta-oe has `dialog_1.3-20160828.bb`.
- `libxklavier_5.0.bb`: Unused and meta-oe provides `libxklavier_5.4.bb`.
- `python-epydoc_3.0.1.bb` & `python-epydoc-native_3.0.1.bb`: meta-oe provides `python-epydoc_3.0.1.bb`, local changes are no longer necessary (carried over from previous version?).
- `python-dbus_0.84.0.bb`: Stale, meta-oe provides `python-dbus_1.2.4.bb`.
- `flex_2.6.0.bb`: Duplicated from upstream.
- `oe-addons/*.bb`: Unused recipes that are not parsed by bitbake.


dpsmith:
:clap:

eric-ch:
Build [$6644](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6644)

jean-edouard:
With this PR, dialogs display in "pure ascii", i.e. the dialog borders are dashes, pipes and pluses, as opposed to special characters.
Not a problem for me, it actually looks kinda cool, but probably worth mentioning.

eric-ch:
@jean-edouard Indeed! As you already noticed, the former recipe added `--with-ncurses` to `EXTRA_OECONF`. If that settings requires a patch (use `pkg-config` to find `libncurse`), I agree the aesthetic change is not worth the patch. Instead it should be sent upstream and picked up when the new version comes out, should it be accepted of course.

jean-edouard:
Sounds good, merging soon.

PR #1203, opened 2019-07-08
S9: Cleanup real-time clock support

jandryuk:
This is the stable-9 version of https://github.com/OpenXT/xenclient-oe/pull/1196

Only the Dom0 and Installer machines have access to a hardware real-time clock. By removing 'rtc' from MACHINE_FEATURES (through MACHINE_FEATURES_BACKFILL_CONSIDERED), we also remove the hwclock.sh initscript from the virtual machines.

We can then clean up some image recipes.

save-rtc.sh is redundant, so we can drop that as well.


jean-edouard:
Merging soon.

PR #1206, opened 2019-07-10
STABLE-9: OXT-1629: Backport critical CVEs from master OE layers

crogers1:
This PR includes several critical CVE patches and upgrades (which include CVEs) to a number of packages built and/or deployed in OpenXT. They are all backported and adapted from upstream commits in meta-openembedded and openembedded-core. Each package has its own commit and is marked with the OXT-1629 ticket so the backport can be tracked and removed when we upgrade OE to Zeus or newer.

`Signed-off-by: Chris Rogers <rogersc@ainfosec.com>`


jean-edouard:
Merging soon.

PR #1207, opened 2019-07-15
S9 : OXT-1636 : libnl : fix disabling of network

mahantesh-ais:
OXT-1636

Unchecking the Enable Networking button under Network icon is broken.
With the network disabled, guests (with wired networking) are still able to access the internet.
This is essentially because NetworkManager fails to bring the brbridged interface down.
NetworkManager uses libnl to bring an interface up or down, in case of bringing brbridged down,
libnl gives an error (-NLE_OPNOTSUPP) of operation not supported.

In stable-7 branch this feature is working. libnl version in stable-7 is 3.2.25
The current version is 3.2.29. The change that is causing this issue is the first argument
that is passed to the function build_link_msg(), it used to be RTM_NEWLINK, now for brbridged
it is essentially RTM_SETLINK. This PR reverts this change to fix our issue.

Looking at the description for rtnl_link_change()/rtnl_link_build_change_request() functions,
it is clear that if  msg fails with -NLE_OPNOTSUPP when msg type is RTM_NEWLINK,
the operation is retried with msg type altered to RTM_SETLINK. This is because RTM_NEWLINK is supposed to work for newer linux kernels and RTM_SETLINK for older ones.

### DIFF
--- a/recipes-support/libnl/libnl/revert-rtnl-link-change.patch
+++ b/recipes-support/libnl/libnl/revert-rtnl-link-change.patch
@@ -1,0 +1,13 @@
+--- a/lib/route/link.c
++++ b/lib/route/link.c
+@@ -1729,9 +1729,7 @@ int rtnl_link_build_change_request(struc
+ 	    !strcmp(orig->l_name, changes->l_name))
+ 		changes->ce_mask &= ~LINK_ATTR_IFNAME;
+ 
+-	rt = af_request_type(orig->l_family);
+-
+-	if ((err = build_link_msg(rt, &ifi, changes, flags, result)) < 0)
++	if ((err = build_link_msg(RTM_NEWLINK, &ifi, changes, flags, result)) < 0)
+ 		goto errout;
+ 
+ 	return 0;
--- a/recipes-support/libnl/libnl_3.2.29.bbappend
+++ b/recipes-support/libnl/libnl_3.2.29.bbappend
@@ -1,0 +1,5 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${BPN}:"
+
+SRC_URI += " \
+    file://revert-rtnl-link-change.patch \
+"


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6655

eric-ch:
Builds [$1207](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6655)

jean-edouard:
I tested before and after, this PR indeed fixes the issue.
Nice find, that was pretty critical!
Merging soon.

jandryuk:
Is there a master PR for this change?

Looks like https://github.com/thom311/libnl/commit/3c427d0f4fcca91a661d85d21cdea89d41271c6c upstream will make this patch unnecessary.

mahantesh-ais:
master branch has Network Manager version updated to 0.9.8.10, where bringing up/down brbridged interface is not an issue since its type is **bridge** as it should be. In stable-9 (or older versions), the Network Manager version is 0.9.2 where we kinda hack an **802-3-ethernet** (eth) device and change its name to brbridged ( this can be seen by running `nmcli d` in ndvm, which shows the device type).

If the underlying device type is **bridge** then link msg type being RTM_SETLINK is not a problem, but since it is **802-3-ethernet**, the msg fails.

Essentially, the above commit at [thom311/link ](https://github.com/thom311/libnl/tree/master)in master branch sets the msg type to RTM_NEWLINK if the BRIDGE_ATTR_FLAGS are "not" set in **ce_mask** attribute of the bridge device. In our case BRIDGE_ATTR_FLAGS are never set (at least to my knowledge) so msg type will always be RTM_NEWLINK. 

But, yes the above change in master branch of libnl can also be applied as a patch. I only checked their latest release (3.4.0) where this is change not present.

jandryuk:
Oh, okay.  Thanks @mahantesh-ais .  I didn't realize your NM 0.9.8.10 update only went into master and not stable-9.

PR #1195, opened 2019-06-27
SyncVM fixes

jandryuk:
The syncvm image is a little broken.  This PR provides some fixes.

syncvm was using the default SELinux fstab.early, but syncvm doesn't run SELinux.  Additionally, the regular fstab was incorrect.  Fix up both of those.  One symptom was sshd failed to start, so you couldn't sshv4v in.

We also have some fixes for setting up volatiles.

sync-client wasn't running since pyicbinn couldn't import importlib.  Fix that dependency.

PR #1208, opened 2019-07-19
S9: SyncVM fixes

jandryuk:
This is the stable-9 version of https://github.com/OpenXT/xenclient-oe/pull/1195

The syncvm image is a little broken. This PR provides some fixes.

syncvm was using the default SELinux fstab.early, but syncvm doesn't run SELinux. Additionally, the regular fstab was incorrect. Fix up both of those. One symptom was sshd failed to start, so you couldn't sshv4v in.

We also have some fixes for setting up volatiles.

sync-client wasn't running since pyicbinn couldn't import importlib. Fix that dependency.


jean-edouard:
Merging soon.

PR #1218, opened 2019-07-23
S9: OXT-1577: udev: fix cdrom eject by disabling locking of cdrom devices

mahantesh-ais:
OXT-1577

Currently udev locks the cdrom device and on pressing the physical eject button the cd-tray doesn't open. Typically udev should recieve an event (DISK_EJECT_REQUEST) on pressing the physical eject button, but in OpenXT it doesn't. The only way now to fix this is to not lock the cdrom device.

cdrom_id is used by udev to get device information and capabilities, this info is used to create symlinks like /dev/cdrom or /dev/dvd. Also, cdrom_id is used to lock the device. This PR removes the argument to lock the device. Locking was basically done to recieve DISK_EJECT_REQUEST event, as this event is not getting triggered there is no use of locking.

Note: If mounted, care should be taken to unmount the device before pressing eject button.


jean-edouard:
Merging soon.

PR #1220, opened 2019-07-25
S9: OXT-1649: linux: Add PVH support config for UIVM.

eric-ch:
UIVM does not have hard limitation to run as a PVH guest.
Add the relevant kernel configuration to handle that use-case.


jean-edouard:
Merging soon.

PR #1221, opened 2019-07-26
cross recipe: renaming initramfs.gz to initrd

dpsmith:
OpenEmbedded classes and recipes follow the convention that an image's
init image is called initrd when built into the image. In order to
better use OpenEmbedded's existing tooling it is helpful to follow then
naming convention of the init image.

This is a cross cutting change that must be made in a single commit. As
a result this make the appropriate changes across the supporting files
for both openxt-measuredlaunch and xenclient-dom0-tweaks.

This is tied to OpenXT/installer#114


eric-ch:
Will merge soon.

PR #1223, opened 2019-07-30
STABLE-9: OXT-1659: [upgrade] Add case for upgrading from SRTM-only

crogers1:
When upgrading from an SRTM-only install, it is impossible to
  forward seal for a DRTM+SRTM configuration since DRTM PCRs
  must exist at forward seal time to accurately predict them
  on the subseqent boot. Instead of displaying a generic
  measurement failure message, provide more context about
  why the next boot will fail measurement and prompt for a
  reseal to use a DRTM+SRTM configuration.

  OXT-1659

### DIFF
--- a/recipes-openxt/xenclient-root-ro/xenclient-root-ro/init.root-ro
+++ b/recipes-openxt/xenclient-root-ro/xenclient-root-ro/init.root-ro
@@ -161,6 +161,30 @@ txt_failure()
     exit 1
 }
 
+srtm_upgrade()
+{
+    dialog --colors --backtitle "${BACK_TITLE}" --defaultno \
+        --yes-label "Continue" --no-label "Shutdown" --yesno "
+    \ZbSRTM Upgrade Detected\ZB
+
+    The system has detected a Static Root of Trust for Measurement(SRTM) 
+    only configuration on the previous install. This is usually due to a
+    recent upgrade from a version of OpenXT that only supported SRTM.
+
+    This version of OpenXT uses Dynamic Root of Trust for Measurement (DRTM)
+    in addition to SRTM. Because of this, measurement will fail to unseal the
+    platform. To continue utilizing the highest level of measurement for OpenXT, 
+    please reseal now to use both DRTM and SRTM." 0 0
+
+    [ $? -ne 0 ] && poweroff
+
+    recovery
+    [ $? -eq 0 ] && return 0
+
+    poweroff
+    exit 1
+}
+
 unseal_failure()
 {
     dialog --colors --backtitle "${BACK_TITLE}" --defaultno \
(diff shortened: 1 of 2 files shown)


crogers1:
Would like some opinions on this.  I didn't want to just use a flag, but what stops me from verifying that SRTM passed on the subsequent boot is that we still seal against 17, 18, and 19 while they're all f's.  On DRTM+SRTM boot, they won't be f's so unsealing will fail no matter what.  At least we can try to differentiate the case and explain the reason needed for resealing.  If this sort of thing is unnecessary, I have no problem declining it.

jandryuk:
One thing I've thought about is our initial installation sealing gap where we reboot to perform the initial sealing.  That is so we seal against the correct lower PCRs (boot config).  I've wondered about having the installer forward seal PCRs 15, 17-19 on a fresh install.  Those should all be known at install time.  Then on first boot, we can re-seal against the full PCR set.

For SRTM -> SRTM+DRTM, the same sort of thing could be done - seal against a subset and then reboot sealing against the full set.  I guess if the re-seal flag is inside /config, we have some protection about ensuring we only re-seal when we unlocked successfully.  But having an automated re-seal process also makes me a little wary that it could be exploited to re-seal against a bad config.

crogers1:
We could also forward seal a dummy value against just the SRTM PCRs. This way we can verify that the platform passes SRTM checks without risking sealing the actual /config key against a subset of PCRs.  If the dummy blob unseals, then display the SRTM -> SRTM+DRTM message, prompt to unlock via recovery, and reseal.

recipes-openxt/xenclient-root-ro/xenclient-root-ro/init.root-ro
```diff
@@ -337,6 +361,16 @@ unlock_config()
                 touch ${measured_flag}
                 return 0
             }
+
+            if [ -e ${tss_path}.srtm ]; then
+                maybe_break "srtm-upgrade"
+                rm ${tss_path}.srtm
```
jean-edouard: Shouldn't the flag be removed later? If for example the user is instructed to shutdown and call an admin when dialog messages happen at boot time, which would make sense, the admin will not see the message again next boot. But maybe that's the desired behavior here...

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6677

recipes-openxt/xenclient-root-ro/xenclient-root-ro/init.root-ro
crogers1: Honestly the flag approach seems so flimsy I'd rather just release note the upgrade behavior for this case and decline this PR.

PR #1227, opened 2019-08-08
OXT-1664: Refactor code duplication between packages.

eric-ch:
Remove code duplication between the different scripts used to perform sealing/unsealing operations by implementing common behaviors in `xenclient-tpm-scripts` and using the existing `tpm-functions` files in dependent scripts. Duplicate implementations were starting to drift apart (OXT-1665, OXT-1666).

Helper introduced in `tpm-functions`:
- tpm_get_syspath: print sysfs path to TPM device.
- tpm_get_ev_separator: print EV_SEPARATOR value.
- tpm_list_pcrs: list PCR values (1.2 && 2.0)
- tpm_pcrlist_save: Sanity check and save PCR id to file.
- tpm_pcrs_to_opts: Convert list of PCR ids to list of options suitable to use with TSS tools.
- tpm_pcr_value_normalize: Normalize PCR values across TSS version.
- tpm_get_hash_algorithm: Print default hash algorithm for platform.

This also changes the content of some files, mostly used for diagnostic:
- `/boot/system/tpm/config.tss.pcrs` no longer has the TSS options depending on the TSS tools used for the relevant TPM version. It is now a list of PCR ids separated by spaces. This makes it easier to be used generically in `tpm-functions`.
- `/boot/system/tpm/forward_pcr.lst` now contains only the PCR values predicted during forward sealing, one `PCR-ID:HASH` pair per line.

### DIFF
--- a/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
+++ b/recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
@@ -388,24 +388,11 @@ encrypted_unlock() {
     local part="${1}"
     local name="${2}"
     local key_file="${3}"
-    local ret=0
 
-    is_tpm_2_0
-    local tpm2=$?
-    if [ "${tpm2}" -eq 0 ];
-    then
-        if pcr_bank_exists "sha256"; then
-            local unseal_file=${key_file}
-            tpm2_unsealdata -H 0x81000000 -n "${unseal_file}.sha256" -u "${key_file}.pub.sha256" -g 0xB $( < ${unseal_file}.pcrs )| cryptsetup -q -d - -S ${ESLOT} luksOpen "${part}" "${name}" >/dev/null 2>&1
-            ret=$?
-        fi
-    else
-        tpm_unsealdata_sa -z -i $key_file | \
-            cryptsetup -q -d - -S ${ESLOT} luksOpen "${part}" "${name}" >/dev/null 2>&1
-        ret=$?
-    fi
-
-    return $ret
+    # Since the key is read from stdin, `-d -` is absolutely necessary, or
+    # cryptsetup will stop reading it at the first LF character.
+    tpm_unseal "${key_file}" | \
+        cryptsetup -d - -S "${ESLOT}" luksOpen "${part}" "${name}"
 }
 
 remove_own_key() {
--- a/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
+++ b/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
@@ -68,14 +68,6 @@ pcr_in_selection() {
     return 1
 }
 
-# Usage: read_pcr_selection
-# Sanitize and print the content of file ${config_pcrs} on stdout.
-# ${config_pcrs} is defined in the 'Configurables' section.
-read_pcr_selection() {
-    # strip out -r/-p to leave just the digits
-    sed -e 's/-[rp]//g' "${config_pcrs}"
-}
-
 # Usage rotate_sealed_files
 # Rename the existing sealed key files to *.old.
 rotate_sealed_keyfiles() {
@@ -146,7 +138,7 @@ while getopts ":sufc:t:p:g:b:r:4:8:h" opt; do
 done
 
 # read_pcr_selection will read ${config_pcrs} which can be changed on cmdline.
-pcr_selection="$( read_pcr_selection )"
+pcr_selection="$(cat "${config_pcrs}")"
 
 case "${operation}" in
 seal)
--- a/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
+++ b/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
@@ -57,10 +57,10 @@ EOF
 # ${pcr_selection} is defined in the 'Configurables' section, it contains the
 # list of PCRs defined in ${config_pcrs}.
 pcr_in_selection() {
-    pcr="$1"
+    local pcr="$1"
 
-    for p in $pcr_selection ; do
-        if [ $p -eq $pcr ] ; then
+    for p in ${pcr_selection}; do
+        if [ "${p}" -eq "${pcr}" ]; then
             return 0
         fi
     done
@@ -73,22 +73,34 @@ pcr_in_selection() {
 # ${config_pcrs} is defined in the 'Configurables' section.
 read_pcr_selection() {
     # strip out -r/-p to leave just the digits
-    sed -e 's/-[rp]//g' $config_pcrs
+    sed -e 's/-[rp]//g' "${config_pcrs}"
+}
+
+# Usage rotate_sealed_files
+# Rename the existing sealed key files to *.old.
+rotate_sealed_keyfiles() {
+    local ext
+
+    for ext in {,.pub}.sha1 {,.pub}.sha256; do
+        if [ -e "${sealed_key}${ext}" ]; then
+            mv "${sealed_key}${ext}" "${sealed_key}${ext}.old"
+        fi
+    done
 }
 
 # Globals
-TPM_DEV="$(find /sys/class -name tpm0)/device"
-
-is_tpm_2_0
-tpm2=$?
+hashalg="$(tpm_get_hash_algorithm)"
+if is_tpm_2_0; then
+    tpm2="0"
+else
+    tpm2="1"
+fi
 
 # Configurables
 operation="seal"
 config_key="/config/keys/config.key"
 sealed_key="/boot/system/tpm/config.tss"
 config_pcrs="/config/config.pcrs"
-pcr_selection="$( read_pcr_selection )"
-hashalg="sha1"
 good_pcrs="/config/good.pcrs"
 bad_pcrs="/boot/system/tpm/bad.pcrs"
 root_dev="/dev/mapper/xenclient-root"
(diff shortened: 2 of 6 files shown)


recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
```diff
@@ -619,26 +619,128 @@ tpm2_write_tboot_policy() {
...
-        else
-            echo "$(tpm2_pcrlist -L sha1:${pcr} | tail -1 | cut -f2 -d: | tr 'A-Z' 'a-z' | tr -d \ )"
-        fi
+    if is_tpm_2_0; then
+        tpm_pcr_value_normalize "$(tpm2_pcrlist -L ${hashalg}:${pcr})"
     else
-        local tpm="$(find /sys/class -name tpm0 2>/dev/null)/device"
-
-        echo "$(grep PCR-${pcr} ${tpm}/pcrs|cut -f2 -d:|tr 'A-Z' 'a-z'|tr -d \ )"
+        tpm_pcr_value_normalize "$(grep PCR-${pcr} "$(tpm_get_syspath)/device/pcrs")"
```
jandryuk: Are we losing the PCR-${pcr}:/${pcr}: prefix?  Seems like tpm_pcr_value_normalize only prints the value.

Or is the documentation of this function wrong?

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
```diff
@@ -619,26 +619,128 @@ tpm2_write_tboot_policy() {
...
+}
+
+# Usage: tpm_pcr_value_normalize PCR-VALUE
+# Print the PCR-VALUE, from TPM-TSS output, normalized on stdout.
+# Normalized here means no space, all lowercase, no PCR identifier, and only
+# the last line.
+# This is relevant for to be consistent across TSS versions and keeps format
+# sanitizing in one centralized place, should it require to be amended.
+tpm_pcr_value_normalize() {
+    echo "$1" | awk -F ':' 'END{ gsub(" ", "", $2); print tolower($2) }'
```
jandryuk: Aside: I've noticed our awk comes from busybox.  I think we also don't have explicit dependencies on busybox or awk.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
```diff
@@ -619,26 +619,128 @@ tpm2_write_tboot_policy() {
...
+# Write the list of PCR-IDS to the PCRLIST-FILE, overwriting the existing one
+# if necessary.
+tpm_pcrlist_save() {
+    local file="$1"
+    shift 1
+    local pcr
+    local pcrlist=""
+
+    # Sanity check list of number in PCR id range ([0-23])
+    for p in $@; do
```
jandryuk: Usually you want "$@" to ensure parameters aren't space split again. `foo "1 2" "3 4"` has `1="1 2"` and `2="3 4"`.  It would be wrong to call this function that way, but it stood out to me.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
```diff
@@ -643,7 +642,7 @@ tpm_get_ev_separator() {
     if is_tpm_2_0; then
         # TPM 2.0: `printf "\xff\xff\xff\xff" | sha256sum`, best guess...
         evs="df3f619804a92fdb4057192dc43dd748ea778adc52bc498ce80524c014b81119"
-     else
+    else
```
jandryuk: Should be squashed in the commit introducing this.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
```diff
@@ -631,7 +631,6 @@ tpm_get_syspath() {
         return 1
     fi
     echo "${syspath}"
-    return 0
```
jandryuk: Should be squashed in the commit introducing this.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
```diff
@@ -748,200 +747,155 @@ tpm_get_pcr() {
...
-# -p has no effect.
 tpm_unseal() {
     local root=""
-    local hashalg="sha1"
+    local hashalg="$(tpm_get_hash_algorithm)"
     local pcrs=""
-    local pcr_params=""
     local OPTIND
     local opt
+    # We capture stderr but let stdout untouched using that variable.
```
jandryuk: "Capture stderr into this variable while stdout is output normally" ?

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
```diff
@@ -73,22 +80,43 @@ pcr_in_selection() {
...
+        mv ${sealed_key}.pub.sha256 ${sealed_key}.pub.sha256.old
+    fi
+}
 
-is_tpm_2_0
-tpm2=$?
+# Globals
+if is_tpm_2_0; then
+    tpm2="0"
+    hashalg="sha256"
```
jandryuk: One call of hashalg=$( tpm_get_hash_algorithm ) ?

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch_1.0.bb
```diff
@@ -35,4 +35,5 @@ RDEPENDS_${PN} = " \
     tboot-utils \
     tboot-pcr-calc \
     openxt-keymanagement \
+    xenclient-tpm-scripts \
```
jandryuk: Make openxt-keymanagement RDEPEND on xenclient-tpm-scripts instead of adding it explicitly.

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
```diff
@@ -301,39 +271,42 @@ forward)
...
         pcr_params="${pcr_params} -p ${p}"
     done
 
-    [ -e "${sealed_key}.${hashalg}" ] || err "sealed key file missing"
-    [ "${tpm2}" -eq 0 ] && [ -e "${sealed_key}.pub.${hashalg}" ] ||
+    if [ ! -e "${sealed_key}.${hashalg}" ]; then
+        err "sealed key file missing"
+    fi
+
+    if [ "${tpm2}" -eq "0" -a ! -e "${sealed_key}.pub.${hashalg}" ]; then
```
jandryuk: You mostly eliminated tpm2, but I would switch the tpm2 variable to tpm_ver - set it to "1" or "2" (or "1.2" or "2.0") and then check against that.
`if [ "${tpm_ver}" -eq "2" -a ! -e "${sealed_key}.pub.${hashalg}" ];` reads more intuitively to me.

recipes-openxt/xenclient-root-ro/xenclient-root-ro/init.root-ro
```diff
@@ -223,64 +223,37 @@ fail_first_seal()
...
-        sealerr=$?
-    fi
-
-    if [ -e ${tss_file} ]; then
-        file_exists=0
-    elif [ -e ${tss_file}.sha256 ]; then
-        file_exists=0
-    else
-        file_exists=1
+    for p in $(sed -e 's/-[rp]//g' "/config/config.pcrs"); do
```
jandryuk: I thought you dropped the -r/-p from config.pcrs?  Use `read_pcr_selection`?

dpsmith:
> You mention in a comment how tpm 1.2 doesn't need the unseal PCRs specified. I would be nicer if the tpm2-tools generated standalone sealed data blobs in the same way. It would make some nice simplication to the unseal side. I haven't looked at the TPM 2 stuff enough to know what would be involved in doing that.

The last time this came up, I pulled open the spec and confirmed that it is not in the seal data. To support this, a new customer record wrapper format would have to be written either directly in tpm2-tools or wrapping shell script that would persist the PCRs used in the seal.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
dpsmith: Good point, the recipe should add RDEPENDS for any shell utilities it expects to be there.

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
```diff
@@ -1,20 +1,33 @@
...
+        Is openxt-measuredlaunch installed correctly?"
+fi
+. /usr/lib/openxt/ml-functions
+
+# Load xenclient-tpm-scripts helpers.
+if [ ! -f /usr/lib/tpm-scripts/tpm-functions ]; then
+    err "Failed to load tpm-functions. \
+        Is xenclient-tpm-scripts installed correctly?"
+fi
+. /usr/lib/tpm-scripts/tpm-functions
```
jandryuk: ml-functions already sources `/usr/lib/tpm-scripts/tpm-functions` & errors with a plain echo & exit 1: https://github.com/OpenXT/xenclient-oe/blob/master/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions#L26

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
eric-ch: Absolutely, these quotes are missing.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
eric-ch: @jandryuk Correct, the `RDEPENDS` has been missing for a while.

In our case this is provided by the `busybox` package. Do we want a `PREFERRED_PROVIDER` in the `DISTRO`, should we ever want to switch to a different implementation?

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
eric-ch: The documentation is confusing and merely documented why it was necessary: The raw data format changes based on the TPM version.

Only the PCR value is intended to be returned.

eric-ch: Yes, the documentation is incorrect. Only the normalized value of the PCR is meant to be printed.

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch_1.0.bb
eric-ch: `seal-system` does not use `key-functions`.
So it seems preferable to leave `xenclient-tpm-scripts` in here (`tpm-functions` is sourced in `seal-system`) and add `xenclient-tpm-scripts` as `RDEPENDS` to `openxt-keymanagement` as well.

recipes-openxt/xenclient-root-ro/xenclient-root-ro/init.root-ro
eric-ch: `/boot/system/tpm/config.tss.pcrs` not longer has the options.
`/config/config.pcrs` still does.

Maybe we should remove that across the board then.

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
eric-ch: Indeed, which drops the `RDEPENDS` on `xenclient-tpm-scripts` in the mean time.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
jandryuk: I think the trickiness is that RDEPENDS/PREFERRED_PROVIDER work on packages and not executables.  Adding awk to openxt-measuredlaunch.bb fails `ERROR: Nothing RPROVIDES 'awk'`

Looking in oe-core, I think awk handling is a little broken there too.  The initscripts recipe packages scripts (populate-volatiles.sh, read-only-rootfs-hook.sh & functions) using `awk` without RDEPENDS.

I don't know how to handle this.  That said, I don't think this series is responsible for fixing it.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
```diff
@@ -619,213 +621,295 @@ tpm2_write_tboot_policy() {
...
+                -g "${TPM_ALG_SHA256}" ${pcr_opts} 2>&1 1>&3 3>&-)"; } 3>&1
+            if [ "$?" -ne 0 ]; then
+                echo "tpm_unseal: Failed to unseal data (TPM 2.0) "\
+                    "using sha256 algorithm:" >&2
+                echo "${stderr}" >&2
+                return 1
+            fi
         ;;
         *)
+            echo "tpm_unseal: Algorithm \"${hashalg}\" is not supported." >&2
```
jandryuk: Since the calls are basically identical, you could refactor to use variables in a single tpm_unseal call.  Something like:
```
case ${hashalg} in
    sha1)
        handle=$OXT_HANDLE_SHA1
        alg=$TPM_ALG_SHA1
        ;;
    sha256)
        handle=${OXT_HANDLE_SHA256}
        alg=${TPM_ALG_SHA256}
        ;;
    *) error "tpm_unseal Alg $hashalg unsupported"
        ;;
esac
{ stderr="$(tpm2_unsealdata -H "${handle}" \
    -n "${tss}.${hashalg}" -u "${tss}.pub.${hashalg}" \
    -g "${alg}" ${pcr_opts} 2>&1 1>&3 3>&-)"; } 3>&1
```

recipes-openxt/xenclient-root-ro/xenclient-root-ro/init.root-ro
jandryuk: Yeah, I would favor just dropping them.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
eric-ch: ... and the same for `tpm_seal` of course.

eric-ch:
### v2:
- Amend documentation of `tpm_get_pcr`
- Quote `$@` in `tpm_pcrlist_save` & `tpm_pcrs_to_opts`
- Re-arrange history changes in their respective commits.
- Factorise `tpm_seal`/`tpm_unseal` logic by introducing new helpers (`hashalg_to_alg` & `alg_to_handle`).
- Remove redundant sourcing of `tpm-functions` in `seal-system`
- Use string for TPM version in `seal-system` and factorize the algorithm selection (`tpm_get_version` & `tpm_get_hash_algorithm` helpers).
- Amend redirection comment in `tpm_unseal`

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6686

recipes-openxt/openxt-keymanagement/openxt-keymanagement/key-functions
```diff
@@ -390,22 +390,10 @@ encrypted_unlock() {
     local key_file="${3}"
     local ret=0
```
jean-edouard: `ret` unused.

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
```diff
@@ -1,20 +1,26 @@
...
     . /etc/openxt/seal-system.conf
 fi
 
-err() {
-    echo "seal-system: $1" >&2
-    exit 1
-}
-
+# Usage: constains_only STRING CHARACTER
+# Succeeds when STRING contains only CHARACTER.
```
jean-edouard: `Succeeds when all the characters in STRING are CHARACTER.`?

recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
```diff
@@ -46,36 +52,59 @@ Usage: seal-system [-h] [-s|-u|-f] [-ctpgbr path] -[48 object[,objects]]"
...
-    sed -e 's/-[rp]//g' $config_pcrs
+    sed -e 's/-[rp]//g' "${config_pcrs}"
 }
 
-# Globals
-TPM_DEV="$(find /sys/class -name tpm0)/device"
+# Usage rotate_sealed_files
+# Rename the existing sealed key files to *.old.
+rotate_sealed_keyfiles() {
+    if [ -e "${sealed_key}" ]; then
```
jean-edouard: How about a loop like `for extension in "" .sha1 .pub.sha1 .sha256 .pub.sha256; do [...]`?

recipes-openxt/xenclient-root-ro/xenclient-root-ro/init.root-ro
```diff
@@ -305,11 +278,10 @@ unlock_config()
     local lv_name="config"
     local tss_path="${SYS_TPM_DIR}/config.tss"
     local measured_flag="/config/tpm/measured-boot"
+    local allout
```
jean-edouard: `allout` doesn't seem to be used in this function.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
```diff
@@ -111,33 +119,52 @@ handle_alg() {
     esac
 }
 
+# Usage: alg_to_handle TPM-ALGORITHM-CODE
+# Print the matching handle code for the given TPM algorithm code, or nothing
+# if not match is found.
```
jean-edouard: Typo, s/not/no/.

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
```diff
@@ -619,213 +646,259 @@ tpm2_write_tboot_policy() {
...
-                pcrs="${pcrs} ${OPTARG}"
-            ;;
-        esac
+# tpm_pcrlist_save PCRLIST-FILE PCR-IDS
+# Write the list of PCR-IDS to the PCRLIST-FILE, overwriting the existing one
+# if necessary.
+tpm_pcrlist_save() {
+    local file="$1"
+    shift 1
+    local pcr
```
jean-edouard: `pcr` unused?

recipes-openxt/xenclient-tpm-scripts/xenclient-tpm-scripts/tpm-functions
```diff
@@ -619,213 +646,259 @@ tpm2_write_tboot_policy() {
...
     done
-    shift $((OPTIND-1))
 
-    local secret="${root}/${1}"
-    local tss="${root}/${2}"
+    echo "${pcrlist}" > "${file}"
+}
+
+# tpm_pcrs_to_opts PCR-IDS
+# Print a list of optionts from PCR-IDS suitable to use as argument with TSS
```
jean-edouard: Typo ("optionts")

eric-ch:
### v3:

- Rebase on master.
- Attend typos.
- Remove lingering unused variables.
- Remove TSS options from `/config/config.pcrs`, both files containing the PCR IDs share the same format now.
- Fix `pcr_bank_exists` leaking variables and missing quoting.
- Move `rotate_sealed_keyfiles` logic in `tpm-functions` and add two helpers `tpm_save_keyfiles` & `tpm_restore_keyfiles`. These are now used in `tpm_seal` for error handling/recovery. This centralize the key-files management for `tpm_seal`.

PR #1233, opened 2019-08-14
libargo: Don't call restorecon from udev

jandryuk:
Stop calling restorecon on the /dev/argo_{dgram,stream} device nodes.

udev already labels the devices properly when they are created, so
restorecon is unnecessary.

This means uivm won't call have
udevd[99]: failed to execute '/sbin/restorecon' '/sbin/restorecon /dev/v4v_stream': No such file or directory
udevd[99]: failed to execute '/sbin/restorecon' '/sbin/restorecon /dev/v4v_dgram': No such file or directory

OXT-1535


jean-edouard:
Build here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6690

PR #1234, opened 2019-08-14
udev-volatiles: Conditionally call restorecon

jandryuk:
I just saw "SELinux `/sbin/restorecon' tool is missing." in the uivm
logs.  It is indeed missing since uivm doesn't run SELinux.  This means
udev-volatiles bailed early without creating /run/lock.

/etc/init.d/functions already has a restore() function to only call
restorecon when it is present.  Switch to using that instead of bailing
early from the script.


jean-edouard:
Build here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6690

PR #1235, opened 2019-08-15
OXT-1503: Convey XSM/Flask SID sender context in Argo messages

dozylynx:
Patches for both Xen and the Linux driver, either side of the updated ABI.

Redefines the message header format to accommodate the new 32-bit XSM/Flask SID value field that provides context from the hypervisor about the sender domain to the receiver domain, and adjusting the size of the message_type field from 32-bit to 8-bit, which also affects the range of values accepted for the message_type argument of the sendv hypercall operation. The fixed protocol indicator values for DGRAM and STREAM on the Linux side are revised to fit within the narrower 8-bit range.

### DIFF
--- a/recipes-openxt/argo-module/argo-module/linux-argo-xmit-xsm-context.patch
+++ b/recipes-openxt/argo-module/argo-module/linux-argo-xmit-xsm-context.patch
@@ -1,0 +1,206 @@
+From 9b019db98c2c52d9c34b1734e82b878417a3c89c Mon Sep 17 00:00:00 2001
+From: Christopher Clark <christopher.w.clark@gmail.com>
+Date: Wed, 14 Aug 2019 14:56:35 -0700
+Subject: [PATCH] OXT-1503: Update the argo-linux driver to new Argo ABI with
+ XSM sid in message header
+
+Handles the changes to the Argo message header format and sendv operation
+arguments for the ABI change to indicate XSM/Flask SID in Argo messages:
+the message_type field has been reduced in size to 8 bits from 32 bits,
+and the message source domain and port fields are now standalone rather
+than within a xen_argo_addr struct.
+
+Since the message_type field, which is used to indicate the protocol that
+is in use between driver endpoints -- ie. either datagram or stream -- has
+decreased in size, the protocol indicator values for dgram and stream are
+changed to new values within the 8-bit range.
+
+Signed-off-by: Christopher Clark <christopher.clark6@baesystems.com>
+
+ argo-module.c      | 43 ++++++++++++++++++++-----------------------
+ include/xen/argo.h | 11 +++++++----
+ 2 files changed, 27 insertions(+), 27 deletions(-)
+
+--- a/argo-module.c
++++ b/argo-module.c
+@@ -236,8 +236,8 @@ static struct list_head ring_list;
+ #define ARGO_SHF_PING        (1 << 8)
+ #define ARGO_SHF_PONG        (1 << 9)
+ 
+-#define ARGO_PROTO_DGRAM     0x6447724d
+-#define ARGO_PROTO_STREAM    0x3574526d
++#define ARGO_PROTO_DGRAM     0x01
++#define ARGO_PROTO_STREAM    0x02
+ 
+ struct argo_stream_header
+ {
+@@ -263,7 +263,7 @@ argo_ring_bytes_to_read(volatile struct xen_argo_ring *r, uint32_t ring_size)
+  */
+ static ssize_t
+ argo_copy_out(struct xen_argo_ring *r, uint32_t ring_size,
+-              struct xen_argo_addr *from, uint32_t * protocol,
++              struct xen_argo_addr *from, uint8_t *protocol,
+               void *_buf, size_t t, int consume)
+ {
+     volatile struct xen_argo_ring_message_header *mh;
+@@ -288,15 +288,12 @@ argo_copy_out(struct xen_argo_ring *r, uint32_t ring_size,
+     if ( btr < len )
+         return -1;
+ 
+-#if defined(__GNUC__)
+     if ( from )
+-        *from = mh->source;
+-#else
+-    /* MSVC can't do the above */
+-    if ( from )
+-        memcpy((void *) from, (void *) &(mh->source),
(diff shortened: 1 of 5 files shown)


dpsmith:
Overall looks good to me. One question that does not have to be resolved for this to go in, how long should the driver patch remain a patch vs being PR'ed to the linux-xen-argo project?

recipes-openxt/argo-module/argo-module/linux-argo-xmit-xsm-context.patch
```diff
@@ -0,0 +1,206 @@
...
+--- a/argo-module.c
++++ b/argo-module.c
+@@ -236,8 +236,8 @@ static struct list_head ring_list;
+ #define ARGO_SHF_PING        (1 << 8)
+ #define ARGO_SHF_PONG        (1 << 9)
+ 
+-#define ARGO_PROTO_DGRAM     0x6447724d
+-#define ARGO_PROTO_STREAM    0x3574526d
++#define ARGO_PROTO_DGRAM     0x01
++#define ARGO_PROTO_STREAM    0x02
```
jandryuk: This doesn't have to be addressed in this patch, but you now have ARGO_PROTO_DGRAM=1 & ARGO_PROTO_STREAM=2 here and ARGO_PTYPE_DGRAM=1 & ARGO_PTYPE_STREAM=2, in argo-linux/include/linux/argo_dev.h.  Can ARGO_PROTO_DGRAM/STREAM just be converted to ARGO_PTYPE_*.  It's not clear why argo_dev.h exposes ptype.

Also, looking in /usr/include/bits/socket_type.h the defines are flipped compared to argo:
SOCK_STREAM = 1
SOCK_DGRAM = 2

jandryuk (approved):
LGTM

jandryuk:
It'd be nice to tag or branch linux-xen-argo with a xen-4.12 version in case someone on that version tries to use the driver.

recipes-openxt/argo-module/argo-module/linux-argo-xmit-xsm-context.patch
dpsmith: To pile on, would there be a reason we just drop "ARGO_PROTO_*" and use the "standard" [socket types](https://elixir.bootlin.com/linux/latest/source/include/linux/net.h#L59)?

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6696

recipes-openxt/argo-module/argo-module/linux-argo-xmit-xsm-context.patch
dozylynx: Looking at it... so prior to this change there was a clear distinction between the magic-valued protocol indicator that went between domains (ARGO_PROTO_DGRAM/STREAM) and the protocol indicator that goes between userspace and the kernel (ARGO_PTYPE_DGRAM/STREAM),
eg. supplied in the argument to bind or returned by the ioctl.

After this PR which effectively removes the magic values, yep, it looks like using standard `sock_type` values for both would be fine. There's some conversion logic in `libargo` which converts between ARGO_PTYPE and SOCK values which could go too.

It gets slightly more painful to do the above in a patch carried within `xenclient-oe` because we have several argo recipes as part of resolving our exotic circular dependencies in that layer, and it would touch multiple of them. That points towards running a branch on the `linux-xen-argo` repository for OpenXT to do this work.

dozylynx:
> It'd be nice to tag or branch linux-xen-argo with a xen-4.12 version in case someone on that version tries to use the driver.

agreed; done: https://github.com/OpenXT/linux-xen-argo/releases/tag/xen-4.12

jandryuk:
Thanks for tagging, @dozylynx

dpsmith:
@dozylynx would you mind resolving the conflict and confirm if you want to reject or resolve the last comments by @jandryuk and myself?

dozylynx:
@dpsmith : yes, I will respin this to resolve the conflict and address the comments by @jandryuk and yourself. I agree with the suggestion to switch to using the standard `sock_type` and will look into how to go about that.

dozylynx:
ok, I didn't get an immediate test pass when testing the respin of this with an incremental build, so I'll pick this up tomorrow when I have a new full build completed.

dozylynx:
(build is still under way to enable testing of the respin to resolve the conflict)
I've just filed this ticket: https://openxt.atlassian.net/browse/OXT-1691
"Argo: Use Linux sock_type values for communicating protocol"
for tracking work on addressing the comments about the protocol value with the `sock_type` enum.

dozylynx:
I'm going to respin this PR on top of the code for this one:

https://openxt.atlassian.net/browse/OXT-1692
https://github.com/OpenXT/xenclient-oe/pull/1260

dozylynx:
@dpsmith : have pushed an updated version of the commits to resolve the conflict to the branch for this PR. Observing the encounter with OXT-1692 during this work suggests that test coverage is currently too sparse, which supports looking at OXT-1683 ("Implement BATS test case to provide test coverage of Argo functions") and keeping this PR separate from the `sock_type` work -- does that sound ok?

dpsmith:
> @dpsmith : have pushed an updated version of the commits to resolve the conflict to the branch for this PR. Observing the encounter with OXT-1692 during this work suggests that test coverage is currently too sparse, which supports looking at OXT-1683 ("Implement BATS test case to provide test coverage of Argo functions") and keeping this PR separate from the `sock_type` work -- does that sound ok?

That is fine.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7012

dozylynx:
In testing this PR, please note that it involves an interface change between the guest and the hypervisor, so this is relevant for upgrade: VMs with access to Argo will need to be upgraded at the same time as the hypervisor is changed.

dozylynx:
Am looking into what is involved to enable BATS to test this functionality, which means making the peer SID context reachable from userspace so that a BATS test can read it in order to verify it.

dozylynx:
Testing is blocked on a design question -- comment at: October 11, 2019, 2:03 PM Pacific
https://openxt.atlassian.net/projects/OXT/issues/OXT-1473

PR #1238, opened 2019-08-19
Introduce IMAGE_FEATURE wildcard-sshd-argo

jandryuk:
This is part of work to reduce the number of wilcard argo rings
https://github.com/OpenXT/linux-xen-argo/pull/3

By specifying ARGO_ACCEPT_DOM0_ONLY=1, the argo interposer will create a
single source argo ring to dom0.  We want this in all the servicevms
since they should only accept sshargo from dom0.  Dom0 can't use this
since it needs to accept sshargo from uivm.

Therefore make the default to add ARGO_ACCEPT_DOM0_ONLY=1 and omit it
when IMAGE_FEATURES contains wildcard-sshd-argo.

### DIFF
--- a/classes/openxt-image.bbclass
+++ b/classes/openxt-image.bbclass
@@ -39,6 +39,17 @@ read_only_rootfs_hook_extend() {
 }
 ROOTFS_POSTPROCESS_COMMAND += '${@bb.utils.contains("IMAGE_FEATURES", "read-only-rootfs", "read_only_rootfs_hook_extend; ", "",d)}'
 
+# A hook function to only accept sshargo from dom0.  sshd-argo will bind a
+# single-source ring to dom0 instead of a wildcard ring.
+read_only_rootfs_no_sshd_argo_wildcard() {
+    # Aside from dom0, we only want to allow sshargo from dom0.
+    if [ -f "${IMAGE_ROOTFS}/etc/ssh/sshd_config_argo" ]; then
+        echo "export ARGO_ACCEPT_DOM0_ONLY=1" >> "${IMAGE_ROOTFS}/etc/default/ssh-argo"
+    fi
+}
+ROOTFS_POSTPROCESS_COMMAND += '${@bb.utils.contains("IMAGE_FEATURES", "wildcard-sshd-argo", "", "read_only_rootfs_no_sshd_argo_wildcard; ",d)}'
+IMAGE_FEATURES[validitems] += "wildcard-sshd-argo"
+
 # Add and entry to /etc/inittab to start a tty on hvc0 (debug).
 start_tty_on_hvc0() {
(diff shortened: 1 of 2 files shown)


jandryuk:
This needs one more patch to quiet an AVC when a domain doesn't have the argo enable XSM permission.  Hold on.

jandryuk:
Addition Xen patch added.

@dozylynx what do you think about this argo/xsm patch: https://github.com/OpenXT/xenclient-oe/pull/1238/commits/e558af2cd9a65b3ec350c45101d63026e0d21b60 ?

eric-ch:
Build [$6702](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6702)

dozylynx:
> Additional Xen patch added.
> 
> @dozylynx what do you think about this argo/xsm patch: [e558af2](https://github.com/OpenXT/xenclient-oe/commit/e558af2cd9a65b3ec350c45101d63026e0d21b60) ?

This is fine. I think it's expected for guests to be able to probe to see if Argo is available and not trigger an audit event. The implementation looks good (I've not tested it though) - thanks.

jean-edouard:
LGTM, just one concern about debugging. It is really not obvious why the connection gets refused, and the Xen logs don't really help. It could take a long time to realize that your target actually has ARGO_ACCEPT_DOM0_ONLY set.
That being said, adding verbosity to the refusal in Xen dmesg (like "target only accepts connections from dom0") might be a security concern...
Either way, this PR doesn't introduce the feature, it just uses it, so +1.

jandryuk:
@dozylynx only argo_init() was converted to the `noaudit` variant.  `do_argo_op()` & `compat_argo_op()` both still generate AVCs on guest action.  I think that is fine, but I just want to be clear.

@jean-edouard repeating the steps from @eric-ch, `/var/log/messages` shows:
> VM hypervisor: (XEN) d1v0 argo: vm1 connection refused, src (vm1:2fbe) dst (dm3:8ae).
The hex port numbers are a little confusing, but converted they do show ports 2222/12222.  There is no AVC since Xen couldn't find a UIVM ring for NDVM to send to.  But like you say, getting from there to looking in /etc/default/ssh-argo for `ARGO_ACCEPT_DOM0_ONLY` is not obvious.

stacktrust:
Please create a ticket for this change, include in commit message for all related PRs.

dozylynx:
> @dozylynx only argo_init() was converted to the `noaudit` variant. `do_argo_op()` & `compat_argo_op()` both still generate AVCs on guest action. I think that is fine, but I just want to be clear.

Ack, and that makes sense, yes: the two versions of the op, applied as you've implemented is good. I did have to think about whether adding a separate hook for `argo_init()` (as you've done) but setting "noaudit" for it in the _policy_, rather than "always noaudit" in the code would be the right way to go, and that seems more of a XSM decision than an Argo one, really -- so it might be worth looking at at some point, but I'm fine with how this is and it helps keeps it simple.

PR #1239, opened 2019-08-19
S9: OXT-1670: refpolicy: Let vhd-util create UNIX/DGRAM sockets.

eric-ch:
vhd-util, provided by blktap3, logs to syslog using UNIX/DGRAM socket.

The log now reported is:
```
vhd-util: vhd_util_read_key: using keyfile /config/platform-crypto-keys/ea4f1606-cb49-4788-861f-a68a1168e1ff,aes-xts-plain,512.key, Size (bytes) 64
```
Additional logs is added if `vhd-util` fails.


jean-edouard:
Merging soon.

PR #1240, opened 2019-08-20
Reduce size of stubdom

jandryuk:
This PR reduces the size of the stubdom initramfs image by ~8MB and drops ~180KB from the kernel size.

The biggest gain is dropping rsyslog and just letting qemu log to /dev/hvc0.  On the non-stubdom side, qemu-dm-wrapper now pipes to logger -t qemu-$domid.

A few qemu utilities snuck in and can be deleted.

simple-poweroff is replaced by busybox's poweroff and deleted.


recipes-extended/qemu-dm/qemu-dm/use-relative-xenstore-nodes.patch
```diff
@@ -1,104 +0,0 @@
-################################################################################
-SHORT DESCRIPTION:
-################################################################################
-Use relative paths instead of absolute ones for xenstore device model nodes.
```
eric-ch: That seems necessary for stub-domains to handle these xenstore nodes?

jandryuk: The commit message for the patch removing it explains that it's not used with our version of xen.

jean-edouard: Seems kind of out of the scope of the PR to me, since removing this doesn't help reduce the size of the stubdom. Also the patch remains valid, even if we don't currently use that code path... But I guess if we're absolutely sure we'll never run into this code ever again, then why not...

eric-ch: @jandryuk Indeed, my apologies.

eric-ch:
Build [$6705](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6705)

recipes-extended/qemu-dm/qemu-dm/use-relative-xenstore-nodes.patch
jandryuk: @jean-edouard, yes, it's out of scope for the PR. I can drop it here and PR it separately tomorrow.

eric-ch: @jandryuk @jean-edouard I don't think there is much to gain by having this PR'ed separately.

recipes-kernel/linux/4.19/defconfigs/xenclient-stubdomain/defconfig
```diff
@@ -475,9 +481,7 @@ CONFIG_DMI_SCAN_MACHINE_NON_EFI_FALLBACK=y
 # Tegra firmware driver
 #
 CONFIG_HAVE_KVM=y
```
crogers1: Is it possible to drop this one since we have no need for KVM in the stubdom?

jandryuk: Hi, @crogers1 .  `CONFIG_HAVE_*` variables indicate the availability of features but not selection.  X86 has KVM support, so it selects `CONFIG_HAVE_KVM`.  CONFIG_KVM controls building KVM.

CONFIG_KVM isn't in the defconfig since it's under CONFIG_VIRTUALIZATION which is disabled:
```
# CONFIG_VIRTUALIZATION is not set
```
```
grep KVM recipes-kernel/linux/4.19/defconfigs/xenclient-stubdomain/
# CONFIG_KVM_GUEST is not set
CONFIG_HAVE_KVM=y
```

recipes-extended/qemu-dm/qemu-dm/use-relative-xenstore-nodes.patch
jandryuk: @eric-ch @jean-edouard just let me know what you want me to do regarding the use-relative-xenstore-nodes.patch.

Sorry for putting the out-of-scope patch into this PR.

jean-edouard: It's fine doing it here, but again, please ensure that code path is dead forever.

jandryuk: In `include/hw/xen/xen_common.h`
>#if CONFIG_XEN_CTRL_INTERFACE_VERSION < 41000
>#define XEN_COMPAT_PHYSMAP
 
In `hw/i386/xen/xen-hvm.c` all the `/local/domain/0` code is wrapped in
> #ifdef XEN_COMPAT_PHYSMAP
> static int xen_save_physmap(XenIOState *state, XenPhysmap *physmap)
> ... /local/domain/0 paths modified by patch
> #else 
> ... empty stub
> #endif 

Since we're past 4.10, we're not longer using those functions.  I think we are good.

jandryuk:
Note: stubdom's qemu messages are no longer prefixed with qemu-system-i386.  I think that is okay since qemu is pretty much the only thing running in the stubdom.

recipes-kernel/linux/4.19/defconfigs/xenclient-stubdomain/defconfig
```diff
@@ -138,10 +139,11 @@ CONFIG_ANON_INODES=y
 CONFIG_SYSCTL_EXCEPTION_TRACE=y
 CONFIG_HAVE_PCSPKR_PLATFORM=y
 CONFIG_BPF=y
-# CONFIG_EXPERT is not set
+CONFIG_EXPERT=y
```
jean-edouard: Why is this needed? Just curious.

jandryuk: VT & VGA_CONSOLE are hidden unless we enable EXPERT.  With EXPERT enabled, we can disable them.

recipes-kernel/linux/4.19/defconfigs/xenclient-stubdomain/defconfig
crogers1: Never noticed that distinction.  Thanks for the info!

jean-edouard:
Merging soon.

PR #1241, opened 2019-08-21
qemu chardev-argo fixes

jandryuk:
qemu can go into an infinite loop if it tries to send a message bigger than the destination's ring size or available space.  argo_send() returns EMSGSIZE & argo_chr_write() returns 0.  The chardev code thinks nothing was sent, so it calls back into argo_chr_write() and the process repeats with no progress made.  Change argo_chr_write() to scale back the write size when receiving EMSGSIZE.

Also use the Xen provided define to check for invalid domids.

### DIFF
--- a/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
+++ b/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
@@ -7,7 +7,7 @@
 +chardev-obj-$(CONFIG_XEN) += char-argo.o
 --- /dev/null
 +++ b/chardev/char-argo.c
-@@ -0,0 +1,331 @@
+@@ -0,0 +1,345 @@
 +/*
 + * QEMU System Emulator
 + *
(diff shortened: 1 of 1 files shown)


eric-ch:
Build [$6714](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6714)

crogers1:
I'm interested in taking this for stable-9, if you could open a PR once this master one is approved, that would be very appreciated.

stacktrust:
Please create a JIRA ticket for this fix and associated test case.

Do we need to document how non-QEMU users of chardev-argo can avoid a similar loop?

recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
```diff
@@ -85,20 +85,37 @@
...
++            ret = argo_sendto(s->fd, buf, write_len, 0, &s->remote_addr);
++        }
++
++        if ( ret == -1 && errno == EMSGSIZE) {
++            /* If the write was too big, try a smaller piece */
++            write_len /= 2;
++            continue;
++        } else if (ret == -1) {
++            fprintf(stderr, "%s error: argo_sendto() failed (%s) - %d %d.\n",
++                    ARGO_CHARDRV_NAME, strerror(errno), ret, len);
```
dozylynx: This PR has really good improvements to the code - I especially like the use of clear variable names and the logic to step down the size of the workload to the point that the receiver has available capacity for. Thanks for doing this.
The only (super minor) point I can see in this section is about the error message in the above line -- which was inherited from the prior code: it's not always the case that `argo_sendto` is the function that was invoked -- ie. `argo_send` is called instead for streams. Is the existing error message close enough? Up to you - the improvements here are worthwhile anyway.

jandryuk:
Pushed new version based off the earlier comments with an updated commit message.

eric-ch:
Build [$6731](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6731)

jandryuk:
I force pushed with updated commit messages to capture a little more detail about the changes.

eric-ch:
Build [$6742](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6742)

eric-ch:
Will merge soon.

PR #1242, opened 2019-08-21
linux: Increase dom0 dmesg buffer to 256K

jandryuk:
Dom0 dmesg output is stored in memory before being read by rsyslog.  The
current buffer size is too small, so early dom0 kernel messages don't
make it to the syslog output.  Increase the buffer size to 256K (the
default size for x86) to ensure we don't lose messages.

Other VMs don't need a change since they log to the console which is
captured by dom0's rsyslog.


eric-ch:
Build [$6714](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6714)

stacktrust:
> Increase the buffer size to 256K (the default size for x86)

Do we have a process to audit for differences between OpenXT dom0 defconfig and the defaults for upstream Xen/Linux, e.g. when we upgrade Xen or Linux?

stacktrust:
Should this go into 9.0.1 for supportability?  If so, please add a ticket number to the commit message, so it can be tracked in the release notes.

crogers1:
I wouldn't mind having it in stable-9, but it won't prompt another micro release on its own.

PR #1243, opened 2019-08-22
OXT-1672: refpolicy: Let sysadm_r use xl.

eric-ch:
Let `sysadm_r` use `xl`, even after acceding `sysadm_r` through `newrole` command.
`xl` uses `xm_t`, so add `xen_run_xm` interface to let a given role use `xm_t`.

`setools` recipe is also missing an `RDEPENDS` on `python-importlib`. This will go with upgrading to a newer version of OE that packages `importlib` in `python-core`.


eric-ch:
@jandryuk My bad, I forgot to remove the commit from this one. I intended to send both in one and changed my mind along the way... I'll remove the commit from this one.

PR #1248, opened 2019-08-28
S9: ovmf: update PREBOOT.EXE shasum to match latest version

jean-edouard:
OXT-1678


stacktrust:
Which version is used by upstream Xen and OE meta-virt?

jean-edouard:
@stacktrust none that I could find.

PR #1249, opened 2019-08-28
lvm2: Configuration packaging.

eric-ch:
Use different recipe and packages to provide custom configuration files
in specific images.

The sleight of hands goes:
- Append upstream recipe, add a lvm2-conf package for the default
  configuration. lvm2-conf is added to RRECOMMEND_lvm2.
- PREFERRED_RPROVIDER_lvm2-conf is set to lvm2 in DISTRO configuration.
- Allarch recipes are added to provide the customized configurations.
- Image recipes can RDEPEND on packages providing custom configuration
  or will fallback to lvm2-conf by default.

(Initial implementation for rsyslog: https://github.com/OpenXT/xenclient-oe/pull/1106)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6727

eric-ch:
Silly me... `s/lvm-conf/lvm2-conf`, sorry I did not fold it in the commit and left it out. But that's the the issue here. `lvm2` assigns `RRECOMMENDS_class-target`, so it needs an `_amend`.

Regarding the `-dbg`/`-dev` packages, and I thought empty packages were ignored in the default configuration, it seems [`bitbake.conf`](https://github.com/openembedded/openembedded-core/blob/master/meta/conf/bitbake.conf#L320) actually allow these two specifically.
This [thread](https://github.com/openembedded/openembedded-core/blob/master/meta/conf/bitbake.conf#L320) mentions there may be some default dependencies on `-dbg` and `-dev`...
I do not quite know how that comes into play for packages that simple, or if some check might trip on it down the road.

dpsmith:
A couple of comments you are free to completely ignore. The first for which I am back and forth on myself is that all the change happens under a single commit, would it make sense to incrementally introduce the change? Second. the conf recipe naming seems backwards to me, specifically lvm2-initramfs-conf_1.0.bb is more logical to me than lvm2-conf-initramfs_1.0.bb.

Otherwise :+1:

eric-ch:
@dpsmith For the recipe naming, I followed the naming scheme used for `rsyslog`. If we want to change I can switch both components. The only commit split I could think of would be to remove the per-image files/logic in one and add the configuration recipes in another. That might create more diff noise due to the files not being renamed, but deleted and added.

dpsmith:
> @dpsmith For the recipe naming, I followed the naming scheme used for `rsyslog`. If we want to change I can switch both components. The only commit split I could think of would be to remove the per-image files/logic in one and add the configuration recipes in another. That might create more diff noise due to the files not being renamed, but deleted and added.

The  changes I see are the introduction of lvm2 conf package and making it the preferred package, then the split off to dedicated recipes for the custom confs. And yes that is why I was back and forth on it, breaking it up like that cause some diff noise of delete and add of files instead of a move.

As for the renaming, you are correct and I believe I also objected to the naming scheme with the rsyslog PR.

jandryuk:
You didn't object in the comments on the rsyslog-conf PR 🙂
The benefit of ${PN}-conf-${extra} is that a single grep on ${PN}-conf finds all the results.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6728

PR #1250, opened 2019-08-30
OXT-1682: selinux: labels inconsistency with file_context on clean installation.

eric-ch:
This changeset attends the labeling issues that were revealed by the bats-suite. This can be checked by running `restorecon -Rnv /` in dom0.

I am left with two:
```
Would relabel /run/ifstate from system_u:object_r:initrc_var_run_t:s0 to system_u:object_r:var_run_t:s0
Would relabel /sys/kernel/debug from system_u:object_r:sysfs_t:s0 to system_u:object_r:debugfs_t:s0
```

The first is the result of `ifup` being called by `/etc/init.d/networking` and hence creating `/var/run/ifstate` from the `initrc_t` context, with the default transition from `var_run_t` to `initrc_var_run_t`. I am not quite sure this is wrong, or if the initscript needs its own domain with its own transition. Or if we could just relabel the file in the initscript after it is created.

The second, I think, comes from the fact the `debugfs` is not mounted, the path being labelled during file-system creation[[1](https://patchwork.kernel.org/patch/9799959/)]. I am not sure what to do in this situation, if anything should be done (e.g, remove the expected label from the policy).

Another weirdness I saw was in https://github.com/OpenXT/xenclient-oe/commit/6c045cbd04534d22ae4e332b193c887590f8f913. The fcontext equivalences don't seem to work in this specific case.


dpsmith:
@eric-ch, FYI because the lvm2 change was committed, this PR now has conflicts.

eric-ch:
> So all file contexts should be /run and we shouldn't be using /var/run.

Hopefully that does not conflict with some `refpolicy-mcs` definitions. I'll try to replace the remaining ones and test for `/run/ifstate`.

> Mention the different ways of starting up blktap in the commit message?

+1 will do.

> I wonder if /mnt/upgrade should be something different than tmpfs_t in order to better restrict the location from other processes?

It feels like it should. I would think this needs deeper modifications/additions to the policy on the other hand.

> We want to allow a distinction between the pts

I think it is the original intent to let xl/xenconsoled handle the pts they allocated to guest serial.

> meta-virt xen recipe dynamically creates a volatiles file. Should we include more of that list

Yes, we should. This is not currently used in our recipes though and the volatiles are not deployed. Should this be tackled on step further? Since the "split" between hypervisor and toolstack in the xenclient-oe layer was to accomodate a 32bit dom0 under a 64bit hypervisor, I would think we should go on and remove the relevant xenclient-oe modifications and use the meta-virtualization xen.inc in one single recipe for both hypervisor and toolstack components.

jandryuk:
> > I wonder if /mnt/upgrade should be something different than tmpfs_t in order to better restrict the location from other processes?
> 
> It feels like it should. I would think this needs deeper modifications/additions to the policy on the other hand.

Yeah, that is probably more than you want for this PR.  It's just a thought I had.

> > meta-virt xen recipe dynamically creates a volatiles file. Should we include more of that list
> 
> Yes, we should. This is not currently used in our recipes though and the volatiles are not deployed. Should this be tackled on step further? Since the "split" between hypervisor and toolstack in the xenclient-oe layer was to accomodate a 32bit dom0 under a 64bit hypervisor, I would think we should go on and remove the relevant xenclient-oe modifications and use the meta-virtualization xen.inc in one single recipe for both hypervisor and toolstack components.

I thought the hang-up is the split of xen-libxl to avoid a circular dependency between blktap and xen?  I guess the xen-hypervisor piece is the 32/64 split you mention.  It'd be great to better align with upstream meta-virt.  If we want blktap our of the hotplug scripts, we could avoid that dependency.

Note that the upstream xen hotplug scripts need some of those volatiles directories even if our scripts do not.  Those scripts also depend on perl for a piece of locking which is unfortunate.

eric-ch:
### v2:

- Add details to commit `refpolicy-mcs: blktap devnodes labeling.`.
- Replace `/var/run` file contexts from contributions of the xenclient-oe layer.
- Add search permission on `xen_device_t` to `firstboot_t`
- Add transition for `initrc_t` to `etc_runtime_t` for file `ifstate`. This is to let the `networking` initscript create its runtime configuration file with the correct label.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6749

jean-edouard:
LGTM, merging soon.

PR #1251, opened 2019-09-03
Introduce initramfs-framework functions module

dpsmith:
This PR introduces a functions module which provides a `break` function to interrupt module processing as well a `log` function to send log messages to the kernel log buffer so they can be written to disk stored logs.

### DIFF
--- a/recipes-core/initrdscripts/initramfs-framework/functions
+++ b/recipes-core/initrdscripts/initramfs-framework/functions
@@ -1,0 +1,48 @@
+#!/bin/sh
+# Copyright (C) 2019 Apertus Solutions, LLC
+# Licensed on MIT
+
+# Process any break kernel parameters
+check_break() {
+    local h="${1}"
+    local b="${2}"
+
+    if [ -n "${bootparam_break}" ]; then
+        for i in $(echo "${bootparam_break}" | sed "s/,/ /g"); do
+            if [ "${i}" = "${b}" ]; then
+                msg "Launching ${b} break ${h} hook"
+                sh
+            fi
+        done
+    fi
+}
+
+# Log message using User-level message facility to kernel log buffer
+log() {
+    local level="${1}"
+    local pri=$(( 8 + 6 )) # Default to info
+
+    case "${level}" in
+        "emerg") pri=$(( 8 + 0 )); shift ;;
+        "alert") pri=$(( 8 + 1 )); shift ;;
+        "crit") pri=$(( 8 + 2 )); shift ;;
+        "err") pri=$(( 8 + 3 )); shift ;;
+        "warning") pri=$(( 8 + 4 )); shift ;;
+        "notice") pri=$(( 8 + 5 )); shift ;;
+        "info") pri=$(( 8 + 6 )); shift ;;
+        "debug") pri=$(( 8 + 7 )); shift ;;
+    esac
+    echo "<${pri}> $@" > /dev/kmsg
+}
+
+functions_enabled() {
+    msg "Loading utility functions"
+
+    add_module_pre_hook check_break
+    add_module_post_hook check_break
+    return 1
+}
+
+functions_run() {
+    :
+}
--- a/recipes-core/images/xenclient-initramfs-image.bb
+++ b/recipes-core/images/xenclient-initramfs-image.bb
@@ -15,6 +15,7 @@ IMAGE_INSTALL = " \
     tpm-tools-sa \
     tpm2-tss \
     tpm2-tools \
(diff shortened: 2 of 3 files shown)


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6732

recipes-core/initrdscripts/initramfs-framework/functions
```diff
@@ -0,0 +1,48 @@
...
+
+# Process any break kernel parameters
+check_break() {
+    local h="${1}"
+    local b="${2}"
+
+    if [ -n "${bootparam_break}" ]; then
+        for i in $(echo "${bootparam_break}" | sed "s/,/ /g"); do
+            if [ "${i}" = "${b}" ]; then
+                msg "Launching ${h} break hook"
```
eric-ch: Shouldn't the message be: `msg "Launching ${h} ${b} hook"`? Currently it will only display: `Launching pre break hook` for any hook.

recipes-core/initrdscripts/initramfs-framework/functions
```diff
@@ -0,0 +1,48 @@
...
+# Copyright (C) 2019 Apertus Solutions, LLC
+# Licensed on MIT
+
+# Process any break kernel parameters
+check_break() {
+    local h="${1}"
+    local b="${2}"
+
+    if [ -n "${bootparam_break}" ]; then
+        for i in $(echo "${bootparam_break}" | sed "s/,/ /g"); do
```
eric-ch: `sh` (as provided by busybox), supports parameter expansion. The for iteration list can be replaced with `${bootparam_break//,/ }`.

On a sidenote, `bash` is the default shell in the initramfs.

recipes-core/initrdscripts/initramfs-framework/functions
dpsmith: Yes, i didn't notice that I missed it in testing before PR but as I was testing work that uses it yesterday I noticed that same point as well. So, yes will be pushing a new version today that fixes this.

recipes-core/initrdscripts/initramfs-framework/functions
dpsmith: I didn't want to use sed but trying to stay POSIX shell compliant as it is the least common denominator. In the end I accepted its usage since it is already employed by the core of the initramfs-framework's init script.

Now I would not be opposed to making the shell spawning into a function that attempts to launch more capable shells first then reverting to less capable shells if they are not present.

eric-ch: Not sure I understand your answer, `/bin/sh` is `bash` in the initramfs, probably because `bash` is pulled as an `RDEPENDS` somewhere.

Starting the `sh` binary from `busybox`, the parameter expansion `//` is still supported. I cannot find definite confirmation that it is POSIX compliant though.

It's a nitpick anyway, leaving the `sed` is fine.

dpsmith: Here are the defined [POSIX parameter expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02), very limited compared to `bash`.

Per this [SO answer](https://stackoverflow.com/a/20195842), when you invoke `bash` as `sh` it restricts itself to POSIX mode with the exception of parameter expansion. What I was suggesting is that when break is invoking a shell, it would check if `/bin/bash` existed and invoke it, if not then try `/bin/csh`, etc finally falling back to `/bin/sh`. Maybe that is just being too fancy?

eric-ch: POSIX parameter expansion doesn't define the `//` template, so lets keep `sed` with that in mind.
I did not know about bash [acting like sh](https://github.com/bminor/bash/blob/master/shell.c#L1751) when started as `sh` :), good to know.

If the goal is to stick with POSIX compliance, I do not think trying to determine which shell is present is desirable. Lets keep the assumption to `sh` being present in `PATH`.

> I would not want to rely that `/bin/bash` will always be `/bin/sh`

I did not mean to imply that, or that we should make this assumption. I was merely pointing that it was the setup in the initramfs, so if someone was to test for features being supported by the shell started by a `break` command-line option, said tester would not report a false positive.

dpsmith:
Pushed correction to the info message printed by the break hook.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6749

jean-edouard:
Merging soon.

PR #1252, opened 2019-09-03
Move writing hotplug-status from libxl to xen hotplug scripts

jandryuk:
Upstream xen lets the xen hotplug scripts write the xenstore hotplug-status node.  I believe the idea is that the node is only written connected once the device is actually hooked up in the backend.  For instance, a vif is only connected once attached to a bridge.

Moving the write of hotplug-status out of libxl removes a little bit of our libxl patch queue.

Goes along with https://github.com/OpenXT/toolstack/pull/37


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6733

PR #1253, opened 2019-09-04
OXT-1641: acpica: Switch to up-to-date iasl provider.

eric-ch:
`Iasl` compiler is provided by `acpica`.
The recipe is maintained upstream and its version is far more recent than the `iasl` alternative in OpenXT.


jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6737

PR #1255, opened 2019-09-04
libxl: Use upstream hotplug block scripts

tijko:
When creating a stubdom that wants access to use a raw partition,
the custom xenclient-toolstack block scripts fail.  Using the
scripts from upstream fixes the issue.

Package the legacy udev rules separately for backwards compatibility.

Also, the patch for `xen-hotplug-common.sh` was applied to silence
selinux warnings from the custom logger and now we use syslog

OXT-1356

Signed-off-by: Tim Konick <konickt@ainfosec.com>

### DIFF
--- a/recipes-extended/xen/xen.bb
+++ b/recipes-extended/xen/xen.bb
@@ -8,6 +8,8 @@ XEN_TARGET_ARCH = "x86_64"
 SRC_URI_append = "\
     file://xenconsoled.initscript \
     file://xenstored.initscript \
+    file://xen-vbd3-backend.rules \
+    file://block-vbd3 \
 "
 
 PACKAGES += " \
@@ -98,6 +100,14 @@ INITSCRIPT_PARAMS_${PN}-xenstored-c = "defaults 05"
 FILES_${PN}-misc += " \
     ${sbindir}/xen-diag \
     "
+
+FILES_${PN}-scripts-block += " \
+    ${sysconfdir}/udev/rules.d/xen-vbd3-backend.rules \
+    ${sysconfdir}/xen/scripts/block-vbd3 \
+"
+RDEPENDS_${PN}-scripts-block += " \
+    udev \
+"
 
 EXTRA_OEMAKE += "ETHERBOOT_ROMS=${STAGING_DIR_HOST}/usr/share/firmware/intel.rom"
 
@@ -197,6 +207,12 @@ do_install() {
             echo "d $i 0755 root root - -"  >> ${D}${sysconfdir}/tmpfiles.d/xen.conf
         done
     fi
+
+    # install vbd3 work-around block script and udev rule.
+    install -d ${D}${sysconfdir}/udev/rules.d
+    install -m 0755 ${WORKDIR}/xen-vbd3-backend.rules ${D}${sysconfdir}/udev/rules.d/xen-vbd3-backend.rules
+    install -m 0755 ${WORKDIR}/block-vbd3 ${D}${sysconfdir}/xen/scripts/block-vbd3
+
     rm -rf ${D}/${sbindir}/xen-livepatch
     rm -rf ${D}/${bindir}/xen-cpuid
     rm -rf ${D}/${sysconfdir}/init.d/xencommons
--- a/recipes-extended/xen/xen.bb
+++ b/recipes-extended/xen/xen.bb
@@ -85,6 +85,10 @@ FILES_${PN}-xen-shim = "\
     ${libdir}/xen/boot/xen-shim \
     "
 
+RDEPENDS_${PN}-scripts-common += " \
+    perl \
+"
+
 INITSCRIPT_PACKAGES =+ "${PN}-console ${PN}-xenstored-c"
 INITSCRIPT_NAME_${PN}-console = "xenconsoled"
 INITSCRIPT_PARAMS_${PN}-console = "defaults 20"
@@ -166,7 +170,33 @@ do_install() {
         -i ${D}${sysconfdir}/init.d/xenstored.${PN}-xenstored-c
 
     # These files are not packaged, removing them to silence QA warnings
-    #   sbindir == /usr/sbin, bindir == /usr/bin, sysconfig == /etc
+    VOLATILE_DIRS=" \
+        ${localstatedir}/run/xenstored \
+        ${localstatedir}/run/xend \
+        ${localstatedir}/run/xend/boot \
+        ${localstatedir}/run/xen \
+        ${localstatedir}/log/xen \
+        ${localstatedir}/lock/xen \
+        ${localstatedir}/lock/subsys \
+        ${localstatedir}/lib/xen \
+        "
+
+    # install volatiles using populate_volatiles mechanism
+    install -d ${D}${sysconfdir}/default/volatiles
+    for i in $VOLATILE_DIRS; do
+        echo "d root root 0755 $i none"  >> ${D}${sysconfdir}/default/volatiles/99_xen
+    done
+
+    # workaround for xendomains script which searchs sysconfig if directory exists
+    install -d ${D}${sysconfdir}/sysconfig
+    ln -sf ${sysconfdir}/default/xendomains ${D}${sysconfdir}/sysconfig/xendomains
+
+    if ${@bb.utils.contains(DISTRO_FEATURES, 'systemd', 'true', 'false', d)}; then
+        install -d ${D}${sysconfdir}/tmpfiles.d
+        for i in $VOLATILE_DIRS; do
+            echo "d $i 0755 root root - -"  >> ${D}${sysconfdir}/tmpfiles.d/xen.conf
+        done
+    fi
     rm -rf ${D}/${sbindir}/xen-livepatch
     rm -rf ${D}/${bindir}/xen-cpuid
     rm -rf ${D}/${sysconfdir}/init.d/xencommons
--- a/recipes-extended/xen/files/block-vbd3
+++ b/recipes-extended/xen/files/block-vbd3
@@ -1,0 +1,19 @@
+#! /bin/bash
+
+dir="$(dirname "$0")"
+name="$(basename "$0")"
+if [ ! -e "${dir}/block-common.sh" ]; then
+    logger -p "daemon.err" -t "${name}" -- \
+        "Failed to find xen hotplug script framework in \"${dir}\"."
+    exit 1
+fi
+
+. "${dir}/block-common.sh"
+
+case "${command}" in
+    "add")
+        success
+        ;;
+    *)
+        ;;
+esac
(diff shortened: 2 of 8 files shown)


recipes-extended/xen/xen.bb
```diff
@@ -85,6 +85,10 @@ FILES_${PN}-xen-shim = "\
     ${libdir}/xen/boot/xen-shim \
     "
 
+RDEPENDS_${PN}-scripts-block += " \
```
jean-edouard: I have a couple concerns with this...
First, the scripts included in xen-scripts-block don't appear to use perl. Did you mean `${PN}-scripts-common`? That one includes `locking.sh`, which uses perl once.
The second thing is, do we really want to add perl to dom0 for just one call in one script? I would be against that, unless it's absolutely irreplaceable for some reason... How about a patch to replace that one call with something else? That patch could be submitted upstream, and ultimately removed from OpenXT, without ever having to include perl in dom0.

tijko: I was aware of minimal use of `perl` but that type of decision is out of my hands.  @eric-ch what do you think?

jandryuk: I may have a replacement for Perl using a new little program. Needs some testing, but hopefully I can post it in the next day or two.

jandryuk: This is what I'm thinking:  https://github.com/jandryuk/xenclient-oe/commit/b9f4294cb4b213d8391f69b738e9283bda14d825

I've only tested cmp-fd-file-inode outside of Xen at this time.

eric-ch: @jandryuk I think that makes sense. The `perl` runtime dependency adds around 2.2M (per `packages-split` on `perl` and `perl-lib` its runtime dependency).

If you plan to PR your new tool, do you mind this PR being merged first then yours will remove the `perl` `RDEPENDS`?

jandryuk: cmp-fd-file-inode isn't ready to go yet (looks like I should put it into xen-libxl).  So you can go ahead with this PR and mine will come along later.

jandryuk:
https://github.com/OpenXT/xenclient-oe/pull/1250/commits/6afe48d7e4222c1184944c900efa11f4fda2e894 introduced volatiles to xen-libxl while this PR adds them to xen-volatiles.  The xen & xen-libxl recipes build fine, but we have a failure when installing packages into xenclient-dom0-image:
```
check_data_file_clashes: Package xen-xl wants to install file /home/build/ope
nxt-master/build/tmp-glibc/work/xenclient_dom0-oe-linux/xenclient-dom0-image/1.0
-r0/rootfs/etc/default/volatiles/99_xen
        But that file is already provided by package  * xen-volatiles
```

I think the fix should be to remove the xen-libxl volatiles and let this PR add the larger set of volatiles to xen-volatiles.

eric-ch:
@jandryuk Yes, at least that is how I merged this PR in my development tree. There is no reason to use separate volatile definitions now that we ship them through the recipe.

eric-ch:
In its current state, this PR will no longer let `vbd3` backend be created (since https://github.com/OpenXT/xenclient-oe/pull/1252 was merged):
`tapback` will wait for `hotplug-status` to switch to `connected` before initializing the connection between backend and frontend, in turn changing the `state` of the backend to `connected` which unblocks `xl` before a timeout when adding a new disk device.

This used to be set in `xl` and was removed for disks in https://github.com/OpenXT/xenclient-oe/pull/1252. The `block` hotplug scripts is not called by `xl` for `vbd3` devices either (https://github.com/OpenXT/xenclient-oe/blob/master/recipes-extended/xen/files/libxl-blktap3-support.patch#L220) and this PR removes the `udev` rule that was working around this issue by writting `hotplug-status` when the vbd3 backend appears.
Even with the udev rule in place, the upstream block-script will need a small patch to not fail trying to read the non-existing `params` node (or the `libxl-blktap3-support.patch` needs to be edited to create the `params` node).

All this seems out scope for this PR imho, as its intent is to fix guest receiving an physical partition as backend instead of a file.

We can deal with this in two ways:
- Re-introduce one of the work-around (either really) in this PR. Then in later work, using the now available upstream hotplug scripts, re-work a bit how vbd3 are handled. After talking a bit with @jandryuk , it looks like blktap3 could be supported with small changes to the upstream hotplug block scripts.
- Leave things as they are, so no "raw" partition support until the vbd3 refactoring is done, starting from this PR introducing the required upstream block scripts to get the "raw" partition support in the end.

jandryuk:
Since our patched libxl won't call scripts for blktap3, I think leaving just a udev script to write hotplug-status for vbd3 would be fine to merge this PR.

I wrote some notes on removing the libxl blktap patches and switching to block-tap here: https://openxt.atlassian.net/browse/OXT-1690

eric-ch:
@jandryuk I added a simple `udev` scripts to switch the `hotplug-status` node for `vbd3` devices.

Also I rebased the branch on master and removed the volatile file added during the SELinux work to fallback on the volatiles definition generated by the `meta-virt` recipes.

Testing with Debian, the guest sees both the disk backed by a Logical-Volume, the ISO and the VHD backed by files in dom0.

jandryuk:
For the commit message of https://github.com/OpenXT/xenclient-oe/commit/59c1e7519395e3e561baeab1b2d18e69c74aac7b
The libxl write of hotplug-status was removed in https://github.com/OpenXT/xenclient-oe/commit/31f6540c28aec2d18a5f76e8eb93a6ff40a53a2b, but it was moved into the scripts in https://github.com/OpenXT/toolstack/pull/37.
Those scripts are just no longer installed with this PR.

Overall it looks good to me.

eric-ch:
@jandryuk Message amended to point where the node was set after the patch to libxl was removed and added `udev` runtime dependency.

eric-ch:
Build [$6755](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6755)

recipes-extended/xen/xen.bb
jean-edouard: @jandryuk do you still plan on PR-ing the perl removal? Knowing how many "temporary" fixes became very much permanent in OpenXT, I'm uneasy about merging this without being sure that the proper fix is on its way! Thank you

jandryuk: I have a branch.  I'll give it a quick spin against the latest version of this PR.

jandryuk: Perl removal posted here: https://github.com/OpenXT/xenclient-oe/pull/1268

jean-edouard:
Built here with the perl removal: http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7010

jean-edouard:
Found a couple issues while testing this PR (along with the perl removal).

On a machine with 2 hard drives, I created a VM, disabled stubdoms, added the second drive as a disk for that VM using:
```
xec-vm -n vm add-disk
xec-vm -n vm --disk 2 set phys-type phy
xec-vm -n vm --disk 2 set phys-path /dev/xenclient/vm-disk
```

The VM failed to start due to an AVC. After running `setenforce 0`, the VM ran and I collected these AVCs:
```
Oct  2 15:19:43.296235 kernel:[  730.655008] audit: type=1400 audit(1570029583.279:30): avc:  denied  { search } for  pid=6818 comm="qemu-system-i38" name="blktap-2" dev="devtmpfs" ino=13071 scontext=system_u:system_r:qemu_t:s0:c272 tcontext=system_u:object_r:xen_device_t:s0 tclass=dir permissive=1
Oct  2 15:19:43.296245 kernel:[  730.655423] audit: type=1400 audit(1570029583.279:31): avc:  denied  { read } for  pid=6818 comm="qemu-system-i38" name="sdb" dev="devtmpfs" ino=1545 scontext=system_u:system_r:qemu_t:s0:c272 tcontext=system_u:object_r:fixed_disk_device_t:s0:c272 tclass=blk_file permissive=1
Oct  2 15:19:43.296246 kernel:[  730.655464] audit: type=1400 audit(1570029583.279:32): avc:  denied  { open } for  pid=6818 comm="qemu-system-i38" path="/dev/sdb" dev="devtmpfs" ino=1545 scontext=system_u:system_r:qemu_t:s0:c272 tcontext=system_u:object_r:fixed_disk_device_t:s0:c272 tclass=blk_file permissive=1
Oct  2 15:19:43.296260 kernel:[  730.655489] audit: type=1400 audit(1570029583.279:33): avc:  denied  { getattr } for  pid=6818 comm="qemu-system-i38" path="/dev/sdb" dev="devtmpfs" ino=1545 scontext=system_u:system_r:qemu_t:s0:c272 tcontext=system_u:object_r:fixed_disk_device_t:s0:c272 tclass=blk_file permissive=1
Oct  2 15:19:43.296261 kernel:[  730.655514] audit: type=1400 audit(1570029583.279:34): avc:  denied  { ioctl } for  pid=6818 comm="qemu-system-i38" path="/dev/sdb" dev="devtmpfs" ino=1545 ioctlcmd=0x5326 scontext=system_u:system_r:qemu_t:s0:c272 tcontext=system_u:object_r:fixed_disk_device_t:s0:c272 tclass=blk_file permissive=1
Oct  2 15:19:43.296261 kernel:[  730.655540] audit: type=1400 audit(1570029583.279:35): avc:  denied  { write } for  pid=6818 comm="qemu-system-i38" name="sdb" dev="devtmpfs" ino=1545 scontext=system_u:system_r:qemu_t:s0:c272 tcontext=system_u:object_r:fixed_disk_device_t:s0:c272 tclass=blk_file permissive=1
Oct  2 15:19:43.296262 kernel:[  730.655655] audit: type=1400 audit(1570029583.279:36): avc:  denied  { lock } for  pid=6818 comm="qemu-system-i38" path="/dev/sdb" dev="devtmpfs" ino=1545 scontext=system_u:system_r:qemu_t:s0:c272 tcontext=system_u:object_r:fixed_disk_device_t:s0:c272 tclass=blk_file permissive=1
```

After shutting down the VM, it failed to start again, seemingly because of a stuck tapdisk. Found this in the logs:
```
Oct  2 15:19:43.520405 xl: [6648] libxl_blktap3.c:176:libxl__device_destroy_tapdisk:Failed to destroy tap device id 5796 minor 5
```
The tapdisk points to the vhd of the VM, and trying to destroy it manually (`tap-ctl destroy -p 5796 -m 5`) just hangs forever.

jandryuk:
For the hanging `tap-ctl destroy`, if you do `tap-ctl list`, does it show an entry with '-' where others have numbers?

https://github.com/OpenXT/xenclient-oe/pull/1267 will help with the first AVC for non-stubdom.  For the others, we probably don't want to just give qemu_t permission to write to fixed_disk_device_t.  We'll need some way to label disks & partitions for vm access.  Have xenmgr re-label or make udev assign a label.  For LVM, it may make sense to have a separate volume group besides "xenclient" so that there is better distinction between the system partitions and vm partitions.  For full disks...  I don't know.

jean-edouard:
@jandryuk no dashes in `tap-ctl list`, the line looked like:
`    5796    5    0        vhd /storage/disks/61f0b1f7-0788-40d2-8d5e-6b118adec506.vhd`

Sorry I forgot to include #1267... I'll include it in the next build.

We should already have a way to label disks, that's what svirt-interposer is supposed to do IIRC. Maybe it should be modified to handle raw disks...

Thank you!

jean-edouard:
Merging soon.
Disk access won't work without a stubdom, but that's a separate problem, which could be fixed by also using tapdisk in the non-stubdom case.

PR #1258, opened 2019-09-06
S9: qemu chardev-argo fixes

jandryuk:
argo can return EMSGSIZE when a messages is too big to fit in the
destination ring.  In current code, QEMU will enter an infinite loop and
spam the log.  argo_send returns -1, and then argo_chr_write writes an
error message and returns 0.  The 0 return makes QEMU think there is
still len bytes to send.  Some code paths may just drop the message, but
for the QMP monitor, it will try to resend in a loop without making
progress.

The new code will try backing off the write size until a single
argo_send() succeeds.  Calling argo_send with write_len == 0 is avoided
since that will return zero.  Instead -1 is returned with -EMSGSIZE
which will be returned to the called.  In that case, the monitor code
will drop the message.

While space could become available, we don't know when.  Returning an
error avoids locking up QEMU if the destination ring fills and never
empties.

This issue can be triggered by a QMP command that generates a large
response, such as:
xl qemu-monitor-command $VM 'info mtree'

In the second commit, use the Xen provided define to check for invalid domids.

This is the stable-9 version of https://github.com/OpenXT/xenclient-oe/pull/1241

### DIFF
--- a/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
+++ b/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
@@ -7,7 +7,7 @@
 +chardev-obj-$(CONFIG_XEN) += char-argo.o
 --- /dev/null
 +++ b/chardev/char-argo.c
-@@ -0,0 +1,331 @@
+@@ -0,0 +1,345 @@
 +/*
 + * QEMU System Emulator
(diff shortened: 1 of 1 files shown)


eric-ch:
Build [$6745](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6745)

PR #1259, opened 2019-09-18
OXT-1695: bats: Add an openxt-test image to run the bats-suite at boot.

eric-ch:
Add `bats-auto` to the list of supported `DISTRO_FEATURE` for OpenXT images.
This adds the `bats-suite` to the list of installed packages to dom0 (like `bats` `DISTRO_FEATURE` used to) and install the initscript to run the suite last in the bootstrap.

Since the `bats-suite` tests check for SELinux denials, add a module to the policy to support the tests scripts and not trigger their own AVCs.


recipes-devtools/bats-suite/files/openxt-bats-suite.initscript
```diff
@@ -0,0 +1,22 @@
...
+
+# Source sysvinit rcS.d defaults.
+. /etc/default/rcS
+
+# Source sysvinit functions library.
+. /etc/init.d/functions
+
+# /etc/init.d/openxt-bats-suite: run the OpenXT bats test-suite.
+
+LOG_FILE="/storage/openxt-bats-suite.log"
```
dpsmith: Why not in `/var/log/`?

recipes-devtools/bats-suite/files/openxt-bats-suite.initscript
```diff
@@ -0,0 +1,22 @@
...
+
+begin "Starting OpenXT bats test-suite..."
+
+for dir in /usr/libexec/{dom0,ndvm,uivm}; do
+    if [ -d "${dir}" ]; then
+        bats "${dir}" | tee "${LOG_FILE}" | logger -t 'bats-suite'
```
dpsmith: Why a separate log file if you are already sending everything to syslog? IMHO I would say either manage the logging in the script or just let syslog do it, not both

dpsmith:
Since you are introducing a testing distro, why not just have the distro put SELinux into permissive instead of writing a policy? To some degree it doesn't matter since you already wrote the policy except that we will have to continue to maintain the, which is likely to be little or none.

recipes-devtools/bats-suite/files/openxt-bats-suite.initscript
jean-edouard: IMO it's good to avoid `/var/log`, to be able to test log rotation and stuff.

conf/distro/openxt-test.conf
```diff
@@ -0,0 +1,21 @@
...
+# include conf/distro/${DISTRO}.conf
+# include conf/distro/defaultsetup.conf
+# include conf/documentation.conf
+# include conf/licenses.conf
+# require conf/sanity.conf
+#
+
+DISTRO_FEATURES_append += "bats-auto"
+
+require openxt-common.inc
```
jandryuk: Could you leave openxt-main as-is and just do:
```
require conf/distro/openxt-main.conf

DISTRO_FEATURES += "bats-auto"
```
?

recipes-devtools/bats-suite/files/openxt-bats-suite.initscript
eric-ch: The log file is additional, everything is already logged to `/var/log` through `syslog`. @jean-edouard has a point though, this hits a component we may want to test.

recipes-devtools/bats-suite/files/openxt-bats-suite.initscript
eric-ch: Because I found it easier to read and, should the `syslog` fail, this still let use get the logs. I am not particularly attached to it though, I can remove the `LOG_FILE` entirely.

eric-ch:
@dpsmith Permissive or not, we need a policy for the test-suite if we want to test for AVCs as mentioned in the commit message. It was, IMHO, easier to write a short policy module than to try and filter the AVC triggered by the test-suite. We could make that module a `tunable` though.

conf/distro/openxt-test.conf
eric-ch: That's possible, I am not entirely sure what logic made me split it in a `-common`. Maybe it should be renamed `openxt-main-test`, since it will make the naming consistent should another `DISTRO` emerge.

dozylynx:
@eric-ch : excited to see this -- thanks -- and will be testing with this today, with adding some BATS coverage for Argo in mind (OXT-1683).

conf/distro/openxt-test.conf
dozylynx: Either way you go with populating this file, it looks like ordering the `require` line before doing the append to `DISTRO_FEATURES` (as @jandryuk has it) is the right way around to have the lines.

recipes-devtools/bats-suite/files/openxt-bats-suite.initscript
dpsmith: I am not sure how putting the file into /var/log conflicts with log rotation. Unless it has changed we have a hook in rsyslog that rotates only the files it is writing to based on size, so the only issue I could see is that the log volume filled up though that might be another good test in and of itself.

recipes-devtools/bats-suite/files/openxt-bats-suite.initscript
dpsmith: This gets to a sticking point for me, I personal would like to see us split up syslog logging instead of putting everything in one file. The counter argument that is always given to me is, that is what grep is for. In that same vein, since I will likely never win the syslog split argument, IMHO I think it is probably better to just write to the file from the script and drop the logger part.

jean-edouard: The original idea was that by using one (logrotated) file, we know exactly how much room is available for it, and it's easy to setup things like shutting down the platform when /var/log is about to be full.
If more files are sharing the space, there might be some concurrent access issues to the available disk space...

dozylynx:
ok, I've just tested a build with this PR applied and the test suite does appear to have run correctly and be working in dom0 -- I obtained the following file from `/storage/openxt-bats-suite.log` with the results.
Please note that the non-enforcing test failures are expected, as I'd booted into a configuration where they were intentionally disabled -- this is useful for verifying that the test coverage is actually working. The avc message is not expected as far as I am aware - have copied that from the log below too.

```
1..23
not ok 1 check SELinux enforcing
# (in test file /usr/libexec/dom0/access_control.bats, line 4)
#   `[ "$(getenforce)" = "Enforcing" ]' failed
not ok 2 check XSM enforcing
# (in test file /usr/libexec/dom0/access_control.bats, line 8)
#   `[ "$(xenops getenforce)" = "Enforcing" ]' failed
# /tmp/bats.1415.src: line 8: xenops: command not found
not ok 3 no kernel classes/permissions left undefined in the SELinux policy
# (in test file /usr/libexec/dom0/access_control.bats, line 14)
#   `[ "${count}" -eq 0 ]' failed
# grep: defined: No such file or directory
not ok 4 no avc messages in /var/log/messages
# (in test file /usr/libexec/dom0/access_control.bats, line 20)
#   `[ "${count}" -eq 0 ]' failed
ok 5 check NDVM flask label
ok 6 check UIVM flask label
not ok 7 check SyncVM flask label
# (in test file /usr/libexec/dom0/access_control.bats, line 40)
#   `[ "$status" -eq 0 ]' failed
ok 8 check xenmgr SELinux label
ok 9 check dbd SELinux label
ok 10 check xenstored SELinux label
ok 11 check uid SELinux label
ok 12 check xcpmd SELinux label
ok 13 check input_server SELinux label
ok 14 check rpc-proxy SELinux label
ok 15 check vusb-daemon SELinux label
ok 16 check network-daemon SELinux label
ok 17 read-only root file system
ok 18 check boot SELinux labels
ok 19 check config SELinux labels
ok 20 check storage SELinux labels
ok 21 check core processes are running
ok 22 test VM creation/destroy
ok 23 test VHD creation
```
The avc log message was:
```
Sep 19 01:11:17.125179 kernel:[    4.673334] audit: type=1400 audit(1568855469.208:4): avc:  denied  { module_load } for  pid=396 comm="modprobe" scontext=system_u:system_r:kmod_t:s0 tcontext=system_u:system_r:kmod_t:s0 tclass=system permissive=0
```

dozylynx:
Just filed this ticket to track the bats-suite "verify that XSM is enforcing" test case being broken since it's attempting to use `xenops`, which isn't available. https://openxt.atlassian.net/browse/OXT-1693

conf/distro/openxt-test.conf
dpsmith: I would disagree on keeping openxt-main as I am looking at the possibility of having to introducing a new distro type that would inherit most of the `-common` and overriding what is needed to meet the need for the new distro type.

recipes-devtools/bats-suite/files/openxt-bats-suite.initscript
dpsmith: There is no reason doing multiple files would not work with duplicating the current logrotate wrapper pipe that is in use today. There is nothing in place today that I am aware of that shuts down the platform when `/var/log` is full. If that is desired, an easy approach is to use `auditd` with shutdown policy.

eric-ch:
@dpsmith I agree. I can look into making this a tunable policy feature and have the bats policy module depending on the `DISTRO` setting (through `DISTRO_FEATURES`). I still think it is interesting to have a test-suite that can run actions in an enforced (or at least permissive) environment and check for AVCs.

jandryuk:
When in permissive mode, code can execute codepaths that are not reached under enforcing.  Those code path can then generated AVCs that would not be seen when enforcing.

eric-ch:
I'm not so sure that using the `DISTRO` for this is good idea. Changing that variable is costly in an existing build tree. I would think this could have people either only build with `openxt-main` and not run the suite.

Another alternative would be a different set of images, in the current setup only an alternate dom0, that will ship the `bats-suite` and initscript and a `refpolicy-mcs` with the `bats` module available. This should re-use most of the packages built during a normal build and only add one new rootfs. It also gives the possibility to have a separate installer or a choice at install time to deploy the test image.

dpsmith:
> I'm not so sure that using the `DISTRO` for this is a good idea. Changing that variable is expensive in an existing build tree. I would think this could have people either only build with `openxt-main` and not run the suite.
> 
> Another alternative would be a different set of images, in the current setup only an alternate dom0, that will ship the `bats-suite` and initscript and a `refpolicy-mcs` with the `bats` module available. This should re-use most of the packages built during a normal build and only add one new rootfs. It also gives the possibility to have a separate installer or a choice at install time to deploy the test image.

Back when I first did the BATS work, I suggested a separate image because it would ensure any testing related tweaks and adjustments would have a less probability of slipping into the "production" instance. The pain this brings is in the ISO packaging where we have to handle two different ISO image generation.

stacktrust:
Could we create a ticket for this important change?  Alternately, use OXT-788, where BATS was introduced.

eric-ch:
@stacktrust [OXT-1695](https://openxt.atlassian.net/browse/OXT-1695)

recipes-devtools/bats-suite/files/openxt-bats-suite.initscript
eric-ch: Using the `syslog` for now is trivial. Since this change-set is about bringing an automated test suite capability, I would prefer tackling the `syslog` in a separate PR if that's ok.

eric-ch:
#### v2
- Re-work this proposal by providing an image instead of a `DISTRO`.
- The `refpolicy-mcs` is now using the `meta-selinux` include file instead of creating `bbappend`, so that multiple policies can be configured an shipped in different images.
- A `test` flavored `refpolicy-mcs` is installed for test images with SElinux enabled.
- Some moving around between image recipes and `packagegroup` was done to avoid duplication.

recipes-security/refpolicy/refpolicy-mcs-2.20170204/policy/modules/contrib/bats.te
```diff
@@ -0,0 +1,63 @@
...
+
+# Run xl.
+xen_domtrans_xm(bats_t)
+
+# Run xec/xec-vm
+xec_domtrans(bats_t)
+
+# Manage VM disk images.
+xc_manage_client_disk_files(bats_t)
+
```
dozylynx:
```
# The access_control test suite needs to examine SELinux status
selinux_get_fs_mount(bats_t)
seutil_read_config(bats_t)
```

eric-ch: Thanks for looking into this.

recipes-devtools/bats-suite/files/openxt-bats-suite.initscript
dpsmith: Ack

jandryuk:
I think the image approach is better than the distro one.  I was wondering if we should make the bats suite & refpolicy a configuration option to the regular dom0-image.  That will help prevent them getting out of sync and bit rotting.  Since you make xenclient-dom0-test-image.bb and xenclient-dom0-image.bb share most stuff through an include, maybe it's not much of a big concern.

eric-ch:
@jandryuk Adding a new image was only motivated by offering the possibility to install one or the other through a choice in the installer, but that comes at a cost (two images, possibility of installing to test stuff by mistake, etc).

jandryuk:
@eric-ch Ah, yes, your approach allows an installation time selection.  You structured the images nicely, so my synchronization concern isn't really a worry.

dozylynx:
PR tested with the below applied. The BATS suites are run at boot and the output is present in `/var/log/messages`.

The access_control tests reported failures:
```
Oct 16 19:27:33.816549 bats-suite: not ok 1 check SELinux enforcing
Oct 16 19:27:33.828120 bats-suite: # (in test file /usr/libexec/dom0/access_control.bats, line 4)
Oct 16 19:27:33.842169 bats-suite: #   `[ "$(getenforce)" = "Enforcing" ]' failed
Oct 16 19:27:33.961270 bats-suite: # getenforce:  getenforce() failedok 2 check XSM enforcing
Oct 16 19:27:34.022852 bats-suite: not ok 3 no kernel classes/permissions left undefined in the SELinux policy
Oct 16 19:27:34.035773 bats-suite: # (in test file /usr/libexec/dom0/access_control.bats, line 14)
Oct 16 19:27:34.049066 bats-suite: #   `[ "${count}" -eq 0 ]' failed
Oct 16 19:27:34.107842 bats-suite: not ok 4 no avc messages in /var/log/messages
Oct 16 19:27:34.114574 bats-suite: # (in test file /usr/libexec/dom0/access_control.bats, line 20)
Oct 16 19:27:34.127698 bats-suite: #   `[ "${count}" -eq 0 ]' failed
```
Testing was with the following changes:

* additional `bats_t` permissions for the access_control suite:
```
# The access_control test suite needs to examine SELinux status
selinux_get_fs_mount(bats_t)
seutil_read_config(bats_t)

```

* bats-suite PR#4 applied https://github.com/OpenXT/bats-suite/pull/4

* additional `bats_t` permissions for Argo:
```
# Permissions for the Argo tests, using sshargo and name resolution
ssh_exec(bats_t)
dev_read_urand(bats_t)
dev_read_rand(bats_t)
ssh_manage_home_files(bats_t)
xc_files_rw_argo_chr(bats_t)
files_read_etc_files(bats_t)
sysnet_read_config(bats_t)
allow bats_t self:capability { setuid setgid };
```

To complete the build with the OpenXT container build system, modifications were made to the manifest files and the xenclient-oe layer in the build output directory to handle the renamed dom0 image.

eric-ch:
Stale PR, closing.

PR #1260, opened 2019-09-18
xen: Add missing dummy XSM init code to quiet Argo XSM check patch

dozylynx:
The recent change to quiet the AVC issued when a domain does not have
XSM argo enabled unfortunately omitted a change to add initialization
logic to the dummy policy, so when running with the dummy policy active,
it results in a crash in Xen on boot.
This change adds the needed initialization.

Fixes OXT-1692.


eric-ch:
Build [$6757](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6757)

PR #1263, opened 2019-09-20
Update openxt.cfg to set runlevel for UEFI boot

jandryuk:
commit c11da6014c6e1fb33ec2ba951edfb69765653a69 "[efi] Turn on efi
runtime services in the kernel" reverted setting of "runlevel=3" to the
older "3" (originally done in commit
e2a14f972b9ad022633f5e4db7964a7498b5c5f9 "dom0-tweaks: change runlevel
declaration" for the initramfs-framework conversion).  This breaks
booting console mode since we don't run in init level 3, so surfman
starts and a getty isn't started.  (autostart is not set, so VM aren't
started).

Correct the runlevel setting.


eric-ch:
Build [$6756](http://openxt-builder.ainfosec.com:8010/builders/openxt64/builds/6756)

jean-edouard:
I tried to test this but when I pick the UEFI console option it just fails to boot ("Press F1 key to retry boot.")
Any idea about that? I'm using a Latitude E7470.

jean-edouard:
Meh, console AMT works and I was able to test this.
LGTM, merging soon.

PR #1265, opened 2019-09-23
OXT-1688: [blktap3] Fix blktap3 vhd encryption

crogers1:
Upstream commits broke our vhd encryption implementation. The
  patch header in this commit provides in-depth detail of each
  problem and their proposed solution.

  OXT-1688

### DIFF
--- a/recipes-extended/xen/blktap3/fix-encryption.patch
+++ b/recipes-extended/xen/blktap3/fix-encryption.patch
@@ -1,0 +1,116 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Fix vhd-encryption failures caused by upstream blktap3 changes
+
+################################################################################
+LONG DESCRIPTION:
+################################################################################
+After upstreaming the OpenXT blktap encryption logic, several changes were
+merged into the upstream blktap repository that broke vhd encryption on OpenXT.
+These commits were 032df683d4d86502e2193337d9546e5b07429e77 and
+63e7784aaa05f8c2f96a09af50f6131af11f26b9
+A summary of the problems are as follows:
+
+ 1. The find key logic was if-def'd out. We now set this in blktap3.bb
+ 2. Because of 1, the redefinition of key causes a variable redefine error when
+    we turn on the if-def. Rename the var in function def to avoid this.
+ 3. 63e7784 only loads libblockcrypto if -E is set when calling tap-ctl-open.
+    OpenXT uses tap-ctl-create and always wants crypto on, so use the existing 
+    ifdef to set an always true conditional that loads crypto no matter what
+    cmdline tool invokes the encryption based tap routines.
+ 4. findkey() sets keysize to 512 or 256, but an xts case deals with bytes,
+    better name the variables to indicate whether they contain keysize in bytes
+    or bits, and make sure we convert chain_find_keyed_vhd to bytes before
+    calling xts_aes_setkey()
+
+################################################################################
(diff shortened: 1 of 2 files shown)


recipes-extended/xen/blktap3/fix-encryption.patch
```diff
@@ -0,0 +1,169 @@
...
+-}
+-
+ static int
+ __load_crypto(struct td_vbd_encryption *encryption)
+ {
+@@ -695,41 +683,35 @@ __load_crypto(struct td_vbd_encryption *
+ 		return -ENOMEM;
+ 	}
+ 
+-	if (encryption->encryption_key == NULL) {
```
jandryuk: You could make a smaller patch by changing this to just:
```
-	if (encryption->encryption_key == NULL) {
+	if (0 /* OXT - force else path to always load encryption support */) {
```
and not changing the indenting?

crogers1: 100% correct.  ```If (0)``` always seemed hacky to me but I can absolutely change it if we want a smaller patch.

recipes-extended/xen/blktap3/fix-encryption.patch
```diff
@@ -0,0 +1,169 @@
...
++vhd_open_crypto(vhd_context_t *vhd, const uint8_t *keyin, size_t key_bytes, const char *name)
+ {
+ 	struct vhd_keyhash keyhash;
+ 	int err;
+@@ -402,7 +402,7 @@ vhd_open_crypto(vhd_context_t *vhd, cons
+ 		return err;
+ 	}
+ 
+-	xts_aes_setkey(vhd->xts_tfm, key, key_bytes);
++	xts_aes_setkey(vhd->xts_tfm, key, keysize / 8);
```
BobBall: As the keysize is pulled in within the #ifdef earlier, can we set key_bytes to equal keysize/8 within the #ifdef code? (or have another variable which is initialised to key_bytes and overridden to keysize/8 within the ifdef)

recipes-extended/xen/blktap3/fix-encryption.patch
BobBall: Can the patch work with the #ifdef?
Also, no value in removing dummy_open_crypto is there?

Thinking:
load_crypto = encryption->encryption_key != NULL
#ifdef OPEN_XT
load_crypto = 1
#endif

recipes-extended/xen/blktap3/fix-encryption.patch
BobBall: (Also, IMO, keysize should be keysize_bits or keysize_bytes to be clear - as there are mismatches throughout)

crogers1: Yea the inconsistencies threw me off a bit when I was debugging.  I can do the variable renaming to note bits/bytes, that's not a problem.  Dividing by 8 after the return from ```chain_find_keyed_vhd``` will also drop the need for this hunk.

recipes-extended/xen/blktap3/fix-encryption.patch
crogers1: Yes that is do-able.  I will make the changes and test.

crogers1:
Updated PR from feedback.  Now using ifdef to better handle the ```load_crypto``` and ```key_bytes``` logic

eric-ch:
Will merge soon.

PR #1268, opened 2019-09-26
xen: Remove locking.sh dependency on perl

jandryuk:
Add cmp-fd-file-inode tool to remove dependency on perl.

We put it in its own package and make xen-scripts-common depend on it.
We can't just include it in the xen-scripts-common package because
cmp-fd-file-inode is built in xen-lixl.bb and xen-scripts-common is
built in xen.bb.

This removes the perl dependency in https://github.com/OpenXT/xenclient-oe/pull/1255


jean-edouard:
Built here with https://github.com/OpenXT/xenclient-oe/pull/1255: http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7010

jean-edouard:
https://github.com/OpenXT/xenclient-oe/pull/1255 will be merged soon.
It will break this PR because of a small conflict.
Once https://github.com/OpenXT/xenclient-oe/pull/1255 and the conflict is resolved, I will merge this.

jandryuk:
I noticed an extra `install -d` after pushing - I'll test removing it and re-push.

jandryuk:
Cleaned up version pushed.

PR #1269, opened 2019-10-03
Remove unused qemu-wrappers

jandryuk:
qemu-wrappers provides /usr/lib/xen/bin/qemu-dm_alt, which is a wrapper
to start /usr/lib/xen/bin/qemu-dm, which no longer exists.  It also
provides /usr/share/xenclient/qemu-dm-wrapper_alt which is not useful
without qemu-dm_alt.

Also remove qemu-dm-wrapper-old from the refpolicy to go along with https://github.com/OpenXT/toolstack/pull/39


jean-edouard:
Built here with https://github.com/OpenXT/toolstack/pull/39:
http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7020

PR #1270, opened 2019-10-07
OXT-1696: refpolicy: General cleaning of the recipe and patches.

eric-ch:
While working on https://github.com/OpenXT/xenclient-oe/pull/1259 and upgrading the OE layer. Some changes in the refpolicy recipe from xenclient-oe seems to be shared.

- Add a step to provide configuration files (`modules.conf` and `booleans.conf`). This lets derived policy configure, from the layer, which module to include in the built policy and switch booleans.
- Remove deprecated/unused modules from the patch-queue: `vgmch`, `txtstat`, `get-edid`.
- Fix invalid patch documentation.
- Remove patches one the build system.
- Refresh the patch-queue following https://openxt.atlassian.net/wiki/spaces/DC/pages/20905986/Dealing+with+quilt to avoid further noise.


stacktrust:
Can we tag this PR with the ticket number for OE uprev and/or BATS improvements?

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/booleans.conf
```diff
@@ -0,0 +1,12 @@
...
+# 
+# 
+# 
+# 
+# This should be enabled when all programs
+# are compiled with ProPolice/SSP
+# stack smashing protection.  All domains will
+# be allowed to read from /dev/urandom.
+# 
+global_ssp = false
```
jandryuk: I think we want `global_ssp = true`.  That's what `policy.booleans.diff` was setting.

recipes-security/refpolicy/refpolicy-mcs_2.%.bbappend
```diff
@@ -202,7 +202,7 @@ POLICY_MLS_CATS = "256"
 EXTRA_OEMAKE += "PKGNAME=${POLICY_NAME}-${PV}"
 
 do_srctree_copy() {
-        cp -r ${WORKDIR}/policy ${S}/
+        cp -nr "${WORKDIR}/policy/modules" "${S}/policy"
```
jandryuk: I'm concerned about the no-clobber option.  If we change a module in the layer, and bitbake triggers a re-build, will no-clobber result in the old, stale file still being in `${S}`?

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7018

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/booleans.conf
eric-ch: I thought I commented on that in the commit message, I'll amend that.

`global_ssp` documentation says:
> This should be enabled when all programs are compiled with ProPolice/SSP stack smashing protection.

It does not appear to be the case in our builds (could not find `-fstack-protector*`), which matches the absence of new AVCs with that Boolean set to `false`.

That being said, giving read permissions to `/dev/urandom` to all domains should not be a huge risk either. So I can switch it back.

recipes-security/refpolicy/refpolicy-mcs_2.%.bbappend
eric-ch: Good point, I was afraid of things in the layer overwriting upstream files, but the former is worse.
I think we can leave it up to the developers if they want to overwrite policy files from the layer or not. Additional logic to handle only overwriting files provided in the `policy` directory from the layer would not be worth it.

eric-ch:
### V2:
- Removed `no-clobber` option from `cp` in `do_srctree_copy`
- Moved the `global_ssp = false` change to a separate commit until SSP is enabled for OpenXT image. This last part triggers a bunch of compilation failures and requires more work, it will be addressed in a different changeset.

eric-ch:
#### V3:

- Rebase on top of master.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7026

PR #1271, opened 2019-10-08
blktap3: Update another LIBBLOCKCRYPTO_NAME variable

jandryuk:
LIBBLOCKCRYPTO_NAME exists in vhd-util-copy.c as well as the other
places.  Update that copy as well.  Otherwise, you'll receive an error:
"failed to load crypto support library"

### DIFF
--- a/recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
+++ b/recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
@@ -241,3 +241,14 @@ PATCHES
  
  #define MAX_KEY_SIZE 512
  int CRYPTO_SUPPORTED_KEYSIZE[] = { 512, 256, -1};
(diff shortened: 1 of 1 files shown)


eric-ch:
@crogers1 stable-9.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7021

jean-edouard:
Merging soon.

jandryuk:
Well, there are a bunch of changes from stable-9's v3.20.0 to master where they are merged.  It would be more than just this simple fix.  Another option is to bring in the two patches I upstreamed and re-work the patch queue.  That's adding the two patches and dropping 2 or three hunks.

jean-edouard:
Ok, I'll just merge this and we can worry about rebasing later. Thanks

jandryuk:
Updated commit message to state `vhd-util copy` triggers the failure.

PR #1272, opened 2019-10-09
S9: blktap3: Update another LIBBLOCKCRYPTO_NAME variable

jandryuk:
Stable-9 version of https://github.com/OpenXT/xenclient-oe/pull/1271

LIBBLOCKCRYPTO_NAME exists in vhd-util-copy.c as well as the other
places.  Update that copy as well.  Otherwise, you'll receive an error:
"failed to load crypto support library"

### DIFF
--- a/recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
+++ b/recipes-extended/xen/blktap3/fix-run-time-errors-and-memory-leaks.patch
@@ -241,3 +241,14 @@ PATCHES
  
  #define MAX_KEY_SIZE 512
  int CRYPTO_SUPPORTED_KEYSIZE[] = { 512, 256, -1};
+--- a/vhd/lib/vhd-util-copy.c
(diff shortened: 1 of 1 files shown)


jean-edouard:
Merging soon.

stacktrust:
When does a user see this failure, e.g. which CLI or UIVM operations will trigger that error message?

jandryuk:
I hit it following this release note: https://github.com/OpenXT/docs/blob/master/releases/OpenXT_9.0.2_ReleaseNotes.adoc#618-vhds-created-in-openxt-90-before-902-are-not-encrypted
running `vhd-util copy`

I've added that to the commit message.

stacktrust:
Do we need to update the 9.0.2 release note to indicate that the vhd-util copy step will fail?

jandryuk:
The workaround is to create the symlink: `rw ; ln -s libblockcrypto.so.0.0.0 /usr/lib/libblockcrypto.so ; ro` with the obvious side effect of breaking measured launch.

PR #1273, opened 2019-10-11
qemu: Avoid QMP events filling the argo ring

jandryuk:
RFC since we may want to drop the debugging fprintfs, but they are useful for testing.
----

We've observed a hang where qemu is blocked in argo_send for a QMP
RTC_CHANGE event.  This seems to happen with NTP time updates on Windows
10.

qmp_helper, when not connected, sits blocked in an accept() call and
does not service its Argo connection.  That means events like the above
will fill up the argo ring.  qemu's argo socket is blocking, so when an
additional event comes in, argo_send blocks in the kernel and qemu
hangs.

We already have some connected state logic for QMP, so re-use that.
When not connected, just drop the writes.  Normal writes happen as
response to requests while a client is connected, and a client will stay
connected until it receives its response.  Events may come through while
a client is connected or not.  With a client connected, it's up to the
client to handle them.  Without a client, we can just drop them.

To make this work, argo_chr_read had to move forward the setting &
clearing of connected.  Without that change, on connect, argo_chr_write
could be called while connected=0 and valid responses were dropped.
Setting connected=1 before argo_chr_update_read_handler(chr) and
qemu_chr_be_event() seems to avoid that.

### DIFF
--- a/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
+++ b/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
@@ -7,7 +7,7 @@
 +chardev-obj-$(CONFIG_XEN) += char-argo.o
 --- /dev/null
 +++ b/chardev/char-argo.c
-@@ -0,0 +1,345 @@
+@@ -0,0 +1,354 @@
 +/*
 + * QEMU System Emulator
 + *
@@ -88,6 +88,11 @@
 +    int write_len = len;
 +    int ret;
 +
++    if (!s->connected) {
++        /* This is important to drop events when not connected. */
++        return len;
++    }
++
 + again:
 +    if (s->stream) {
 +        ret = argo_send(s->fd, buf, write_len, 0);
@@ -167,6 +172,9 @@
 +    if (!s->stream) {
 +        if (s->connected == 0 && !strncmp((char*)(s->buf), ARGO_MAGIC_CONNECT, 4)) {
 +            fprintf(stderr, "argo_recvfrom() returned ARGO_MAGIC_CONNECT, connecting.\n");
++            /* update "connected" first since qemu_chr_be_event calls monitor
++               code which can call back into argo_chr_write to send data. */
++            s->connected = 1;
 +            argo_chr_update_read_handler(chr);
 +            qemu_chr_be_event(chr, CHR_EVENT_OPENED);
 +            if (!chr->gsource) {
@@ -174,10 +182,12 @@
 +                                                 argo_chr_read_poll,
 +                                                 argo_chr_read, chr, chr->gcontext);
 +            }
-+            s->connected = 1;
 +            return FALSE;
 +        }
 +        if (s->connected == 1 && !strncmp((char*)(s->buf), ARGO_MAGIC_DISCONNECT, 4)) {
++            /* update "connected" first since qemu_chr_be_event calls monitor
++               code which can call back into argo_chr_write to send data. */
++            s->connected = 0;
 +            argo_chr_update_read_handler(chr);
 +            qemu_chr_be_event(chr, CHR_EVENT_CLOSED);
 +            if (!chr->gsource) {
@@ -186,7 +196,6 @@
 +                                                 argo_chr_read, chr, chr->gcontext);
 +            }
 +            fprintf(stderr, "argo_recvfrom() returned ARGO_MAGIC_DISCONNECT, closing.\n");
-+            s->connected = 0;
 +            return FALSE;
 +        }
 +    }


stacktrust:
> For ease of constructing a test case for this: do you know if the hang is reproducible with in-guest clock changes on a HVM Linux guest?

`hwclock -w` should generate a RTC_CHANGE event

jandryuk:
@stacktrust how did you get `hwclock -w` to generate the event?  It's not doing it for me.

To slightly clarify the original description, it seems Windows 10 is writing to the RTC as NTP makes time changes.  I also noted the RTC_CHANGE when I manually set the time in Windows 7.

I only tried this once, but running `xl qemu-monitor-command $VM 'info version'` unblocked qmp_helper and drained the argo ring.  QEMU resumed execution.

jean-edouard:
Built here: http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7028

dozylynx:
There are several complex state machines involved around this code that are difficult to reason about, as are the possible effects of changing them or of silently dropping writes as proposed in this PR.

I have an alternative proposal: to add a check for sufficient available space in the ring before issuing the send, and since this is the only place where writes are issued into that ring, this can ensure that it will not block. The call site already appears to handle the case where less than the full amount of data available is sent, so that can be used in conjunction with the check.

To enable the 'sufficient space' check that this requires, the hypervisor's Argo 'notify' op provides support for querying a ring for sufficient space, so in order to make this available for this qemu caller, we can plumb a method of access through the Linux driver and userspace libargo with a small amount of straightforward logic.

The proposed change for qemu is here:
https://github.com/dozylynx/xenclient-oe/commit/7500254e34dd9d44ea5c95ee0a5d252041292b03

and the kernel and userspace changes to implement the new check function are here:
https://github.com/dozylynx/linux-xen-argo/commit/f17f70a8efb536df7ca464050dcfb739e9e51fb8

I've combined the two into a single branch for ease of testing here:
https://github.com/dozylynx/xenclient-oe/commits/oxt-1699-qmp-send-when-ready
(though for a PR it will be changes to both linux-xen-argo and xenclient-oe repos, rather than adding a linux argo patch into xenclient-oe as per the above test branch)

I've performed some initial limited testing that it does not break access to `hwclock -r` from a Linux HVM guest, but at this point further testing is still needed to verify that it addresses the original issue reported that motivated this PR.

eric-ch:
> qmp_helper, when not connected, sits blocked in an accept()

IIUC this happens when the `xl` daemon is not yet connected, or drops the connection, with the UNIX socket ([qmp_helper:257](https://github.com/OpenXT/xctools/blob/master/qmp_helper/src/qmp_helper.c#L257), [qmp_helper:341](https://github.com/OpenXT/xctools/blob/master/qmp_helper/src/qmp_helper.c#L341))?

So this PR is mitigation for a design flaw in the `qmp_helper` (another one of which is being already addressed in https://github.com/OpenXT/xctools/pull/56). QEMU handling this without hanging is fine, but I cannot stop thinking the `qmp_helper` should be redesigned to not have this kind of problems, for example by having a single `select` loop, or use one of the asynchronous event libraries already pulled in dom0, like `libevent`, `libuv`, `libev`.

@dozylynx This can also be addressed by non-blocking operations without having to introduce new changes to Argo or the driver. `write` would return `EAGAIN` or `EWOULDBLOCK` and QEMU can drop or try again later depending on the case. Now, non-blocking operations are far from ideal when not absolutely necessary IMHO. In this case, redesigning the `qmp_helper` and handling the connection accordingly in QEMU sounds preferable.

jandryuk:
@dozylynx thanks for taking a look.

I think we should maintain the "only send when connected - drop otherwise" behaviour.  This is how the QEMU char-socket code works: https://git.qemu.org/?p=qemu.git;a=blob;f=chardev/char-socket.c;h=185fe38dda734c6d51ceba69b3437344135d9dfb;hb=HEAD#l160 (char-argo.c started as a copy of char-udp.c.  Since UDP is connectionless, there is no connection tracking.  Also, UDP will just fire off packets, so it can't block like argo does.)

The thing we have to be careful about is QMP sends JSON.  JSON parsing will go pear shaped if we have an incomplete write of the JSON data.  When connected, the EMSGSIZE fallback code will send partial messages, but they will be followed up by sending the remainder which should work as qmp_helper will be draining the ring.

@eric-ch in regular operation, `xl` is not holding open the QMP socket.  It connects and disconnects as needed.  And during normal running, nothing is connected.  https://github.com/OpenXT/xctools/pull/56 fixes the same flaw and moves to the single loop like you say.  However the qmp_helper change is on the receive side while this char-argo change is on the sending side.  If we know we are not connected, we can just drop the messages since no one will be there to process them.

I would prefer to avoid non-blocking operations, too.

I prefer just using the connected state and dropping messages when not connected.  It's simple and matches the behaviour of other QEMU chardevs.

brendank310:
Have hit a similar issue, and have been able to resolve the issue by disabling the following viridian enlightenment: "HV_X64_MSR_TIME_REF_COUNT". Just providing this as additional information.

dozylynx:
I think there are several pieces involved to fixing this.

I've filed [OXT-1699: Qemu is blocked in argo_send for a QMP RTC_CHANGE event](https://openxt.atlassian.net/browse/OXT-1699) for the specific issue of the QMP RTC_CHANGE event generation causing problems, and I think [OXT-1680: fix qemu argo character device handling when destination ring space is not available](https://openxt.atlassian.net/browse/OXT-1680) should cover the need to introduce some form of flow control between QEMU and qmp-helper.

Since LibXL doesn't act on the QMP Events even if they make it there:
https://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=tools/libxl/libxl_qmp.c;h=9aabad74fabd7bfd27b2bac153973c612a46a3c2;hb=HEAD#l278

it seems appropriate to discard the events earlier to reduce exposure to at least part of the problem.
I've put together a patch that extends the Qemu monitor to add control over event generation, with runtime control over which event types are enabled or not. The patch is here:
https://github.com/dozylynx/xenclient-oe/commit/86908cb02ba2e10dd9fc4bf42c30a275239cd13c

Once applied, the monitor command is accessible from the xl command line, eg:
`xl qemu-monitor-command $VM 'event-enable RTC_CHANGE off'`

There is also a second example patch on the same branch that switches the default for RTC_CHANGE events to drop, at this link:
https://github.com/dozylynx/xenclient-oe/commit/75f6caad1743f4eb6de27944a305f8f9eb2783a3
and the same change of default could be applied to other event types with the same technique.

It is possible that for some workloads this is sufficient for the Qemu->qmp-helper connection to remain free of blocking, if it is only guest-triggered events (which can now be discarded closer to their source) that are causing the connection buffer to be fully consumed. If not, then adding a method of flow control to the connection could be an option to prevent Qemu from issuing blocking sends, as an alternative to dropping messages.

jandryuk:
> I think there are several pieces involved to fixing this.

Implementing "drop when disconnected" behavior is sufficient to fix this.  That also matches regular QEMU behavior, so I think it's the right solution to this problem.

This was only marked RFC since I didn't clean out the debugging print statements.  Otherwise I think the gist of the change is appropriate.

> it seems appropriate to discard the events earlier to reduce exposure to at least part of the problem.

Disabling QEMU event generation is fine.  I would do it through a command line flag instead of a monitor command (though you could leave the monitor command for runtime control if you desire).

> ..., and I think [OXT-1680: fix qemu argo character device handling when destination ring space is not available](https://openxt.atlassian.net/browse/OXT-1680) should cover the need to introduce some form of flow control between QEMU and qmp-helper.

If you want to re-work `argo_chr_write` to query available space, that could be okay.  It may be smarted than the current `/= 2` best effort.  Adding an additional system call (hypercall?) to each write operation should be considered.

The larger problem, as I see it, is QEMU as the sender can't know if out-of-space is a transient failure or a permanent one.  If nothing is emptying the ring, we cannot make progress.  Given the combination of OS & Hypervisor scheduling, it seems to me transient failures should be considered as a possible situation.

This is where tracking the connection state is useful.  Send failure for lack of space is most likely transient when connected.  If we think we are connected, blocking in a write would be okay.  But if we are "connected" and the remote end has crashed, we would hang.

It looks like QEMU uses non-blocking sockets and a 100 microsecond sleep on EAGAIN.  That won't totally handle a crashed remote end.

In separate conversation with Marek, we may have discovered a QEMU flaw that sends the QMP events when it shouldn't.  That is when the chardev is "disconnected".

Also, I have a QEMU 4.1 uprev for OpenXT, but it is only minimally tested.

eric-ch:
Build [$7031](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7031)

eric-ch:
> This was only marked RFC since I didn't clean out the debugging print statements. Otherwise I think the gist of the change is appropriate.

I think so too.
With the debug statement being trivial to trigger from the guest, they would have to be macro'ed out for debugging or removed to avoid log spamming though.

Some tests examples:
- Turning off/on NTP in Windows 10 will trigger `RTC_CHANGE`, so does setting the time manually:
```
[argo-chardrv] not connected, dropping 113 bytes
[argo-chardrv] not connected, dropping 113 bytes
```
- `xl qemu-monitor-command` will exercise the connection and trigger the connection states.

I like @dozylynx feature `xl qemu-monitor-command $VM 'event-enable RTC_CHANGE off'` though, do you plan to submit that change upstream?

stacktrust:
@jandryuk several tickets were filed as a result of the discussion in this PR.  Could you please pick the most appropriate ticket (or create a new one) and add that ticket number to the description of the PR?

jandryuk:
Pushed a version dropping the debugging output and removed the RFC label.

eric-ch:
Build [7040](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7040)

eric-ch:
Will merge soon.

PR #1275, opened 2019-10-15
layer: Remove unused patches against bitbake.

eric-ch:
No functional change.

These were backported long ago and are not used in the project.
All of these should be upstream already.


jean-edouard:

jean-edouard:
Merging soon.

PR #1276, opened 2019-10-15
linux: Remove kernel no longer maintained.

eric-ch:
No functional change.

Linux 4.9 is no longer maintained, shipped or tested with OpenXT. Its last use was in OpenXT 7.
Linux 4.14 is no longer maintained, shipped or tested with OpenXT. Its last use was in OpenXT 8.


jean-edouard:

jean-edouard:
Merging soon.

PR #1277, opened 2019-10-15
class: Remove always_rebuild... class.

eric-ch:
No functional change.

This was maybe intended to have `PSTAGING_DISABLED` for OpenXT recipes (`SRC_URI` has `OPENXT_GIT_MIRROR` source). This is not used and it is not quite clear what `PSTAGING` was.

This class is not inherited anywhere, it will still be parsed by bitbake when the layer is added though. Since it is no longer used, it should be removed.


jean-edouard:

jean-edouard:
Merging soon.

PR #1278, opened 2019-10-15
layer: Retire unused nilfvm and classes.

eric-ch:
- Retire NILFVM. The machine no longer has any image or recipe use.
- Retire the Debian packaging classes. This is redundant with upstream functions and its only use was in NILFVM.
- Remove `xenclient.bbclass`. This currently has no effect, the intended use is redundant with upstream features as well.
- Remove `null.bbclass`. That class was a work-around to conditionally inherit the `xenclient-deb*` classes, which are removed in this PR as well.


dpsmith:
Will need to double check but I believe we are using the NILFVM type in the derivative.

dpsmith:
> Will need to double check but I believe we are using the NILFVM type in the derivative.

Correction, we have moved off of nilfvm type, so no objection to it being retired.

eric-ch:
@dpsmith Thanks for the verification!

@jandryuk Indeed, that will have to be taken care of as well.

@stacktrust Image assembly wise, the layer does not provide a way more specific than what OE provides to build netfilter or VPN service VM. That being said, that would require some discussion with people currently doing this to define what should and can be factorized (e.g, using Debian packages might be controversial) above what OE already provides.

This PR intends to remove dead recipes and classes hanging around un-maintained and may trigger errors and overhead during upgrades.

eric-ch:
@stacktrust Actually there is already a ticket about this, I'll tag the commits shortly.

jean-edouard:

jean-edouard:
Merging soon.

PR #1281, opened 2019-10-31
OXT-1703: Uprev OpenXT Xen version to tip of Xen stable-4.12 and AUTOREV

dozylynx:
For OpenXT's master branch:

* Uprev the Xen revision to the current commit at the tip of stable-4.12
    * removes 3 patches which are now present in the upstream branch
    * rebases 1 patch where it was needed

* Switch the Xen recipe in OpenXT's master branch to AUTOREV, so it will track the tip of the upstream Xen stable-4.12 branch automatically
    * XSA patches will be picked up when they are added to the upstream branch

JIRA ticket [OXT-1703](https://openxt.atlassian.net/browse/OXT-1703) has further notes.


recipes-extended/xen/xen-version.inc
```diff
@@ -1,3 +1,3 @@
 XEN_PV = "4.12.2-pre"
 XEN_BRANCH = "stable-4.12"
-XEN_SRCREV = "c28853456b51516514d36659494fbe7c1d983e2c"
+XEN_SRCREV = "${AUTOREV}"
```
jandryuk: As AUTOREV moves forward, we can get out of sync with XEN_PV.  I don't know how we can handle that in an automated fashion, but XEN_PV could be set to just 4.12 or 4.12+.

recipes-extended/xen/xen-libxl.bb
```diff
@@ -21,9 +21,7 @@ python () {
                 'libxlutil',
                 'libxlutil-dev',
                 'libxenlight',
-                'libxenlight-dev'
-                'cmp-fd-file-inode',
-                ]:
+                'libxenlight-dev']:
```
jandryuk: Sorry about that.

eric-ch:
Build [$7038](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7038)

eric-ch:
Will merge soon.

recipes-extended/xen/xen-version.inc
eric-ch: We can also fallback on the standard `_git+${SRCPV}` versioning of OE since we will now track a branch susceptible to change.

dozylynx: @jandryuk , @eric-ch : ack, I'm looking into this.

dozylynx: Branch refreshed with an update to the above logic to set `XEN_PV` with the `+git${SRCPV}` that is common in OE, to reflect that it is tracking the branch.

I've also removed the reference to `${XEN_PV}` where `FLASK_POLICY_FILE` was set in `xen-libxl.bb`, because that recipe does not package the flask policy file or require that variable. The flask policy filename from a policy built within the Xen tree has a filename that includes the full Xen version:
`xenpolicy-$(XEN_VERSION).$(XEN_SUBVERSION)$(XEN_EXTRAVERSION)`
-- which is affected by the switch to a tracking branch, since that version advances as the branch does. This can be addressed by changing the `FILES_${PN}-flask` definition to add globbing and it looks like the policy file handling in meta-virtualization's Xen recipe could be improved.

eric-ch:
Build [$7046](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7046)

eric-ch:
Will merge soon.

PR #1284, opened 2019-11-04
Stable-9: OXT-1703: Uprev Xen version to tip of Xen stable-4.12 and apply XSA patches

dozylynx:
- Uprev the Xen revision to the current commit at the tip of stable-4.12 as of 31st October
    - remove patches which are now present in the upstream branch
    - rebase one patch where needed

- Add patches for issued XSAs that apply to x86 Xen that are not present in that branch at current revision
    - XSAs 296, 298, 299, 301, 302

Refs: [OXT-1703](https://openxt.atlassian.net/browse/OXT-1703)


crogers1:
@eric-ch if we could get a test build of this just to make sure nothing OpenXT-specific breaks on top of it, that would be great.  Code wise, there's not much more for me to say since the Xen community presumably had enough eyes on it.

eric-ch:
Build [$7048](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7048)

PR #1285, opened 2019-11-06
OXT-1705: linux: Upgrade to 4.19.81.

eric-ch:
From the log:
```
ccb02adf3293 xen/netback: fix error path of xenvif_connect_data()
90a886b68faa xen/efi: Set nonblocking callbacks
975859bb69b2 xen/xenbus: fix self-deadlock after killing user process
a1afd826e549 xen-netfront: do not use ~0U as error return value for xennet_fill_frags()
371357770c8b x86/uaccess: Don't leak the AC flags into __get_user() argument evaluation
47288968eebd xen-netfront: do not assume sk_buff_head list is empty in error handling
eb535aaf3cd7 xen/blkback: fix memory leaks
b3410f0f8505 xen/netback: Reset nr_frags before freeing skb
e72e6ba17ab4 xen/pciback: remove set but not used variable 'old_state'
04fdca1f2f65 xen/swiotlb: fix condition for calling xen_destroy_contiguous_region()
740e0167a382 x86/paravirt: Fix callee-saved function ELF sizes
11cb9f8700c9 xen/pv: Fix a boot up hang revealed by int3 self test
007e5aaf287c xen/events: fix binding user event channels to cpus
e73db096691e xen: let alloc_xenballooned_pages() fail if not enough memory free
4acce744284c xenbus: Avoid deadlock during suspend due to open transactions
66f33b2bd2d8 xen/pvcalls: Remove set but not used variable
```


jean-edouard:

jean-edouard:
LGTM, merging soon.

PR #1286, opened 2019-11-06
qemu: Avoid QMP events filling the argo ring

jandryuk:
We've observed a hang where qemu is blocked in argo_send for a QMP
RTC_CHANGE event.  This seems to happen with NTP time updates on Windows
10.

qmp_helper, when not connected, sits blocked in an accept() call and
does not service its Argo connection.  That means events like the above
will fill up the argo ring.  qemu's argo socket is blocking, so when an
additional event comes in, argo_send blocks in the kernel and qemu
hangs.

We already have some connected state logic for QMP, so re-use that.
When not connected, just drop the writes.  Normal writes happen as
response to requests while a client is connected, and a client will stay
connected until it receives its response.  Events may come through while
a client is connected or not.  With a client connected, it's up to the
client to handle them.  Without a client, we can just drop them.

To make this work, argo_chr_read had to move forward the setting &
clearing of connected.  Without that change, on connect, argo_chr_write
could be called while connected=0 and valid responses were dropped.
That is because qemu_chr_be_event() -> monitor_qmp_event() ->
qmp_send_response()-> argo_chr_write().  Setting connected=1 before
argo_chr_update_read_handler(chr) and qemu_chr_be_event() avoids that.

OXT-1699

### DIFF
--- a/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
+++ b/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
@@ -7,7 +7,7 @@
 +chardev-obj-$(CONFIG_XEN) += char-argo.o
 --- /dev/null
 +++ b/chardev/char-argo.c
-@@ -0,0 +1,345 @@
+@@ -0,0 +1,354 @@
 +/*
 + * QEMU System Emulator
 + *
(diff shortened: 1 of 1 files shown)


jandryuk:
Closing to leave the comment trail all together in https://github.com/OpenXT/xenclient-oe/pull/1273

PR #1288, opened 2019-11-07
ovmf: New mirror available for PREBOOT.EXE.

eric-ch:
Use the currently available mirror for version 24.2.

Attend [recent build failure](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7039) when `PREBOOT.EXE` is not in download cache yet.


jean-edouard:
Nice find, merging soon to fix the build.

PR #1290, opened 2019-11-07
xen-libxl patch cleanup

jandryuk:
I noticed some unnecessary changes in our libxl patches which are removed in this series.

@crogers1 Is there a particular reason for changing "/local/domain/N/control" to rw perms https://github.com/OpenXT/xenclient-oe/commit/db804d91b4d16e498ef8656094ddfc403933c546#diff-83c581625019ac0b6ff2913055c602acR52?  You introduced it originally, but I don't think it's necessary.

### DIFF
--- a/recipes-extended/xen/files/hvmloader-legacy-seabios-optionroms.patch
+++ b/recipes-extended/xen/files/hvmloader-legacy-seabios-optionroms.patch
@@ -145,7 +145,7 @@ PATCHES
  {
      unsigned int bios_dest = 0x100000 - bios_length;
  
-@@ -140,12 +155,79 @@ static void seabios_load(const struct bi
+@@ -140,12 +155,76 @@ static void seabios_load(const struct bi
      memcpy((void *)bios_dest, bios_addr, bios_length);
      seabios_config.bios_address = bios_dest;
      seabios_config.image_size = bios_length;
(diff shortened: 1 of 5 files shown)


jean-edouard:

jean-edouard:
@crogers1, please review.

crogers1:
@jandryuk Sorry for the late review, I missed these notifications.

For the /local/domain/N/control changes, I popped back to the xen 4.6.1 release and looked around for them and couldn't find error, drivers, attr, or messages, so I likely introduced them for that reason.  Xen 4.12 looks like it doesn't have "error" or "messages" set, so you may need to keep those?  Or perhaps they aren't needed at all anymore.

At some point it looks like they split out the vkb/vfb creation, so that commit is fine to drop the redundant code.  Yea overall, looks good.  Once again, sorry for the delay.

jandryuk:
Thanks, @crogers1 

For "error" and "messages", yes, those get dropped.  They are unused on my systems, and are a relic of xenvm, AFAICT.  I should update the commit message to mention that.

jandryuk:
Pushed with an amended commit message for "xen: Remove duplicate xenstore writes " mentioning the removal of "error" and "messages" nodes.

eric-ch:
With no diff post force-push: Will merge soon.

jandryuk:
Making `control` read-only breaks qemu-dm's acpi power-off.  qemu-dm cannot write to `/local/domain/$domid/control/hvm-powerbutton-enable`.  Xenmgr doesn't see that node, so it doesn't write `"/local/domain/" ++ domid ++ "/control/hvm-shutdown") "poweroff"` to trigger the acpi path and defaults to `"xl shutdown -c -w " ++ domid`.

Options:
 - Revert the control permission change.
 - Have libxl create `control/hvm-powerbutton-enable` so it's writable with value "0".  Xenmgr would then need to be a little smarter to ensure the value is "1" before using that shutdown path.
 - Drop "control/hvm-powerbutton-enable" and just make it always available.  That means ACPI must always be available.  Not totally unreasonable since ACPI is basically always used nowadays.
 - Have qemu-dm write to its own xenstore, and have Xenmgr check there instead of in the guest.  Use a writable path under "device-model"?

I think I'm leaning toward the last option.

PR #1291, opened 2019-11-08
refpolicy: Fix typos in patches.

eric-ch:
There are a few typos in some patches that left some leading '+' characters, likely from reject files.

For some reason, building the policy does not fail almost all of the time.


jean-edouard:

PR #1292, opened 2019-11-08
ovmf: Upgrade PREBOOT.EXE to 24.3.

eric-ch:
24.2 mirror changed again since 24.3 is provided.
Also add the md5sum of the archive, making this easier to check with the website.

This changed sometime since yesterday with 24.2 now being provided at https://downloadmirror.intel.com/29166/eng/PREBOOT.EXE.

`E3522X2.EFI` hasn't changed in the new version, this is the only file extracted from the archive.
```
7862a12529539905d4f57c94920d52b4  PREBOOT.0/APPS/EFI/EFIx64/E3522X2.EFI
```


jean-edouard:
Thanks, merging soon.
I assume the plan for S9 is to use the 24.2 URL?

eric-ch:
Sure, the only extracted file is the E1000 NIC driver, which has not changed between 24.2 and 24.3.

PR #1293, opened 2019-11-12
S9: ovmf: Update PREBOOT.EXE 24.2 mirror.

eric-ch:
New mirror provided since 24.3 release.
The E1000 driver provided in 24.3 is the same as 24.2, so there is no reason to update for now. This is the only file extracted and used by this recipe.

Adding `md5sum` for easier verification when consulting the download web-page.


eric-ch:
Build [$7048](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7048)

PR #1298, opened 2019-11-15
packagegroup-dom0: Remove explicit lib depends

jandryuk:
With package dependency tracking, we shouldn't be explicitly installing
libraries.  Just install executables and let their dependencies bring in
their needed libraries.

For the dropped libraries:
xen-xl -> xen-libxenguest, xen-libxenlight, xen-libxlutil
xen-xenstat -> xen-libxenstat
surfman -> libedid
rpc-proxy -> libxchargo -> libargo


eric-ch:
Build [$7056](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7056)

PR #1301, opened 2019-12-10
Upgrade dojo from 1.7.2 to 1.7.12

jean-edouard:
1.7.12 is the latest micro release of the 1.7 line.
1.8 releases break the UI, so minor upgrades are not an option.
1.7.12 was released in March 2017 (vs May 2013 for 1.7.2).

One notable improvement is that 1.7.12 fixes scroll-down lists
with more recent libwebkitgtk versions.


jean-edouard:
Goes with https://github.com/OpenXT/toolstack-data/pull/60

eric-ch:
Testing is easier with comparison on previous Zeus build as it uses libwebkitgtk 2.24 which repro the issue in the first place. Existing builds can be used for comparison (say, [$10013](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10013) for repro).

jandryuk:
LGTM, but one request.  Would you please include any details on "1.8 releases break the UI" in the commit message.  Capturing that now will help when people understand why 1.8 is unusable in the future.  Thanks.

jean-edouard:
@jandryuk honestly I'm not sure how to describe it... I built it there if you want to have a look:
http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7067
When built with dojo 1.8 the UI is just ... a broken mess

jandryuk:
My firewall blocks port 8010 on the builder, and I don't see a public posting of 7067 on https://openxt.ainfosec.com/builds/.  Are the widgets not rendered properly but input works?  Or is input broken and the rendering fine?  Both?  I was hoping for a little description what is broken since broken on its own doesn't describe the issue.

jean-edouard:
Here: https://openxt.ainfosec.com/builds/custom/master/7067/
Some things work, some don't, not sure what a widget is...

jandryuk:
Widgets are a term for the UI elements.

The broken UI fails to render icons along the top bar.  Instead it prints the text labels oriented vertically.  The labels cannot be clicked.  The lower status bar doesn't render the measured launch or power icons.  A VM can be created through the wizard, but the details button cannot be clicked on the resulting VM.

jean-edouard:
Thank you for having a look, I'll add your description to the commit message.

PR #1303, opened 2019-12-16
icedtea: fix compilation on recent compilers

jean-edouard:
icedtea is compiled with -Werror, and compilation fails on recent
versions of gcc which introduce new warnings.
This PR keeps the warnings but avoids error-ing on them.


eric-ch:
This appears to be addressed in `meta-java` recent master: http://git.yoctoproject.org/cgit/cgit.cgi/meta-java/commit/?id=3832a3a74dc987224365ae9500bf2dbecd5e4d50

PR #1304, opened 2019-12-18
qemu-dm: acpi-pm: Use device-model xenstore directory

jandryuk:
6dde8833f7de "xen: Remove duplicate xenstore writes" changed a domain's
control xenstore directory to read-only.  This broke stubdom qemu's
ability to write to "control/hvm-powerbutton-enable".  Without that
node, xenmgr wouldn't use the acpi shutdown paths.

Switch to writing hvm-powerbutton-enable and hvm-shutdown in the
device-model directory of qemu-dm's domain.  That is either the stubdom
or dom0.  Logically it makes sense to put this device-model information
underneath the device-model instead of the guest.  It also side steps
the permission issue.

This needs a corresponding change to xenmgr.

Goes along with https://github.com/OpenXT/manager/pull/175

### DIFF
--- a/recipes-extended/qemu-dm/qemu-dm/acpi-pm-feature.patch
+++ b/recipes-extended/qemu-dm/qemu-dm/acpi-pm-feature.patch
@@ -1111,7 +1111,7 @@ PATCHES
 +   char base[XEN_BUFSIZE];
 +   int err;
 +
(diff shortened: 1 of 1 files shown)


eric-ch:
Build [$7070](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7070)

eric-ch:
Will merge soon.

PR #1305, opened 2019-12-18
xen: Build hypevisor in do_compile

jandryuk:
Right now, do_compile only regenerates the .config file.  The hypervisor
is actually built in do_install since it is called implicitly as a
dependency on install.  Add an explicit call to build xen in do_compile.


eric-ch:
Build [$7072](http://openxt-builder.ainfosec.com:8010/#/builders/1/builds/7072)

eric-ch:
Will merge soon.

PR #1308, opened 2019-12-31
OXT-1281: NetworkManager upgrade to version 1.4.4

mahantesh-ais:
OXT-1281


mahantesh-ais:
To be merged with:
https://github.com/OpenXT/network/pull/17

recipes-connectivity/network-manager-applet/network-manager-applet-1.4.4/com.openxt.xml
```diff
@@ -0,0 +1,10 @@
+<node>
+    <interface name="com.openxt.nmapplet">
+        <method name="popup_network_menu">
+            <arg name="x_ordinate" type="u" direction="in" />
```
dpsmith: IMHO x_ordinate/y_ordinate are a bit unorthodox, unless your a mathematician. Did you consider x_coord/y_coord which I personally seen more often as the short hand for x/y coordinate variables.

recipes-connectivity/network-manager-applet/network-manager-applet-1.4.4/default-certs-dir.patch
```diff
@@ -19,14 +20,15 @@
...
+ 	                                   _("Choose a Certificate Authority certificate"));
 --- a/src/wireless-security/eap-method-tls.c
 +++ b/src/wireless-security/eap-method-tls.c
-@@ -304,9 +304,15 @@ setup_filepicker (GtkBuilder *builder,
+@@ -361,9 +361,16 @@ setup_filepicker (GtkBuilder *builder,
  	GtkWidget *widget;
  	GtkFileFilter *filter;
  	const char *filename = NULL;
 +	const char *default_certs_dir;
++
```
dpsmith: <nit> is the extra white necessary?</nit>

recipes-connectivity/network-manager-applet/network-manager-applet-1.4.4/disable-auto-ethernet.patch
```diff
@@ -33,12 +33,33 @@
...
+--- a/src/applet.c
++++ b/src/applet.c
+@@ -1354,8 +1354,7 @@ nma_menu_device_get_menu_item (NMDevice
+ 	}
+ 
+ 	if (!managed) {
+-		item = gtk_menu_item_new_with_label (_("device not managed"));
+-		gtk_widget_set_sensitive (item, FALSE);
++		return NULL;
+ 	}
```
dpsmith: <nit> since the change resulted in a single line in the `if` statement, you could drop the brackets as well </nit>

recipes-connectivity/network-manager-applet/network-manager-applet-1.4.4/disable-auto-ethernet.patch
```diff
@@ -33,12 +33,33 @@
...
+ 
+ 	return item;
+@@ -1372,7 +1371,7 @@ add_device_items (NMDeviceType type, con
+ 	for (i = 0; all_devices && (i < all_devices->len); i++) {
+ 		NMDevice *device = all_devices->pdata[i];
  
- 	g_slist_free (connections);
+-		if (nm_device_get_device_type (device) == type) {
++		if (nm_device_get_device_type (device) == type && nm_device_get_state (device) != NM_DEVICE_STATE_UNMANAGED ) {
```
dpsmith: <nit> wrap the test condition after the `&&` to keep from excessively exceeding 80 cols</nit>

recipes-connectivity/network-manager-applet/network-manager-applet-1.4.4/openxt-menus.patch
```diff
@@ -0,0 +1,469 @@
...
++
++       		flags |= G_DBUS_CONNECTION_FLAGS_MESSAGE_BUS_CONNECTION;
++		flags |= G_DBUS_CONNECTION_FLAGS_AUTHENTICATION_CLIENT;
++
++		address = g_dbus_address_get_for_bus_sync (G_BUS_TYPE_SYSTEM,
++							   NULL,
++							   &error);
++
++		if (address == NULL){
++			if(error != NULL){
```
dpsmith: <nit> brackets are not necessary here </nit>

recipes-connectivity/network-manager-applet/network-manager-applet-1.4.4/openxt-menus.patch
```diff
@@ -0,0 +1,469 @@
...
++		if (menu_bus != NULL) {
++			g_bus_own_name_on_connection(menu_bus,
++						     "com.openxt",
++						     0,
++						     on_name_acquired,
++						     NULL,
++						     applet,
++						     NULL);
++		} else {
++			if(error != NULL){
```
dpsmith: <nit>braces</nit>

recipes-connectivity/network-manager-applet/network-manager-applet-1.4.4/openxt-menus.patch
```diff
@@ -0,0 +1,469 @@
...
++
+ 	/* Initialize device classes */
+ 	applet->ethernet_class = applet_device_ethernet_get_class (applet);
+ 	g_assert (applet->ethernet_class);
+@@ -3358,5 +3491,23 @@ static void nma_class_init (NMAppletClas
+ 	GObjectClass *oclass = G_OBJECT_CLASS (klass);
+ 
+ 	oclass->finalize = finalize;
++
++        g_warning("register openxt signals");
```
dpsmith: <nit>alignment</nit>

recipes-connectivity/network-manager-applet/network-manager-applet-1.4.4/openxt-menus.patch
```diff
@@ -0,0 +1,469 @@
...
+ 
+--- a/src/applet.h
++++ b/src/applet.h
+@@ -52,6 +52,9 @@
+ 
+ typedef struct {
+ 	GApplicationClass	parent_class;
++	void (*popup_network_menu) (AppletAgent *self, unsigned int x, unsigned int y);
++	void (*close_network_menu) (AppletAgent *self);
++
```
dpsmith: <nit> unnecessary whitespace</nit>

recipes-connectivity/networkmanager/networkmanager-1.4.4/xc-nutty-network.patch
```diff
@@ -0,0 +1,366 @@
...
+ 		nm_device_master_release_one_slave (self, slave, configure, reason);
+ 		/* Bridge/bond/team interfaces are left up until manually deactivated */
+ 		if (priv->slaves == NULL && priv->state == NM_DEVICE_STATE_ACTIVATED)
+--- a/src/devices/nm-device-ethernet-utils.c
++++ b/src/devices/nm-device-ethernet-utils.c
+@@ -31,6 +31,7 @@ nm_device_ethernet_utils_get_default_wir
+ 	guint j;
+ 	int i;
+ 
++	return NULL;
```
eric-ch: That shortcuts a few things, so no ethernet is returned to be the "default" connection?

recipes-connectivity/networkmanager/networkmanager-1.4.4/xc-nutty-network.patch
```diff
@@ -0,0 +1,366 @@
...
++
++	iface = nm_platform_link_get_name(platform, ifindex);
++	if (iface) {
++		int table = 30 + ifindex;
++		guint32 oxt_metric = metric;
++		NMPObject oxt_obj_id;
++		nm_auto_nlmsg struct nl_msg *oxt_nlmsg = NULL;
++
++		if(strncmp(iface, "brbridged", 9) != 0)
++			metric = metric + 1;
```
eric-ch: Not part of this PR: This should be configurable.

mahantesh-ais:
> Generally seems okay. Looks like 1.4.8 is the latest 1.4.x release. Is this 1.4.4 since it was against the pyro version?
> 
> I'd prefer when you add patches, that you put the text from the commit messages at the start of those patch files. That way if someone looks at the file they have the reason right there.

Yes, this 1.4.4 is because it was against pyro version.

Makes sense, I will add the commit message at the start of patch files.

recipes-connectivity/networkmanager/files/com.openxt-nmapplet.conf
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE busconfig PUBLIC
+ "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
+ "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
+<busconfig>
+	<policy user="root">
+		<allow own="com.openxt"/>
+		<allow send_destination="com.openxt"/>
```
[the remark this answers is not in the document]
mahantesh-ais: Yes, I will change it to `org.openxt`. Thanks

recipes-connectivity/networkmanager/files/com.openxt-nmapplet.conf
```diff
@@ -0,0 +1,13 @@
...
+    </policy>
+    <policy context="default">
+		<allow own="com.openxt"/>
+		<allow send_destination="com.openxt"/>
```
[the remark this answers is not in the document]
mahantesh-ais: Yes, this has to be changed. root should only have `allow` rule to send_destination. For default context there should be `allow` rule to own as the context of nm-applet connecting over argo cannot be verified and a `deny` rule to send_destination. Thanks

recipes-connectivity/networkmanager/networkmanager-1.4.4/dbus-system.conf
```diff
@@ -0,0 +1,21 @@
...
+These directives can override D-Bus or OS defaults.
+
+For upstream or distribution-wide defaults that can be overridden
+by a local sysadmin, create files matching
+/usr/share/dbus-1/system.d/*.conf instead.
+-->
+<!DOCTYPE busconfig PUBLIC "-//freedesktop//DTD D-Bus Bus Configuration 1.0//EN"
+ "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
+<busconfig>
+<allow_anonymous/>
```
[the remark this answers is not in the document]
mahantesh-ais: Yes, I will drop the above comment and put a rationale.

As I understand DBUS_COOKIE_SHA1, the cookie stored is updated periodically and nm-applet is required to prove access to the cookie file by reading the cookie. I'm not sure this can be disabled to have one static cookie. Also glib (nm-applet) should be patched to read cookie from domstore or otherwise. This could be researched more and can be taken up as a separate task.

recipes-connectivity/networkmanager/networkmanager-1.4.4/fix-device-type-property-exposure.patch
```diff
@@ -0,0 +1,75 @@
...
+-		{ "device-type", NULL },
+ 
+ 		{ NULL },
+ 	};
+@@ -395,6 +395,36 @@ constructed (GObject *object)
+ 	g_warn_if_fail (G_OBJECT_TYPE (object) == _nm_device_gtype_from_dtype (priv->device_type));
+ }
+ 
++static NMDeviceType
++coerce_type (NMDeviceType type)
```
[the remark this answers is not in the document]
mahantesh-ais: libnm (and hence the clients) track the properties of the device using `property_info`, the device-type property for a device is missing, this patch fixes that. This is useful when configuring a network using nm-applet or other clients.

I will add reason in the patch. Thanks

recipes-connectivity/network-manager-applet/network-manager-applet-1.4.4/com.openxt.xml
mahantesh-ais: Ack, will change to x_coord/y_coord. Thanks

recipes-connectivity/network-manager-applet/network-manager-applet-1.4.4/openxt-menus.patch
```diff
@@ -0,0 +1,469 @@
...
++static void nma_context_menu_create (GtkMenuShell *menu, NMApplet *applet)
+ {
+-	GtkMenuShell *menu;
+ 	guint id;
+ 	static gboolean icons_shown = FALSE;
+ 
+-	g_return_val_if_fail (applet != NULL, NULL);
+-
+-	menu = GTK_MENU_SHELL (gtk_menu_new ());
++	g_assert (applet != NULL);
```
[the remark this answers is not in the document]
mahantesh-ais: The return type of the function is changed to void, keeping `g_return_val_if_fail` will result in error. In order to not proceed without checking `applet` being `NULL`, `g_assert` is introduced instead of `g_return_val_if_fail`

recipes-connectivity/networkmanager/networkmanager-1.4.4/xc-nutty-network.patch
```diff
@@ -0,0 +1,366 @@
...
+ gboolean
+ nm_dns_manager_add_ip4_config (NMDnsManager *self,
+                                const char *iface,
+--- a/src/nm-policy.c
++++ b/src/nm-policy.c
+@@ -472,8 +472,10 @@ update_ip4_routing (NMPolicy *self, gboo
+ 	}
+ 	g_assert ((best || vpn) && best_ac);
+ 
++#if 0
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, will split the current patch into appropriate patches to make the diff explicit. Thanks

recipes-connectivity/networkmanager/networkmanager-1.4.4/xc-nutty-network.patch
mahantesh-ais: Allowing the creation of 'Wired connection %d' is of no use, eth0 on discovering it as a slave for brbridged assumes a slave connection and this connection is left as is instead of being cleaned up. In fact this creates a problem when the ethernet cable is unplugged and plugged back.
On unplug/plug of ethernet cable eth0 simply adopts this connection instead of being a slave connection for brbridged.

recipes-connectivity/networkmanager/networkmanager-1.4.4/xc-nutty-network.patch
mahantesh-ais: Ack, I will remove it. Thanks

recipes-connectivity/networkmanager/networkmanager-1.4.4/dbus-system.conf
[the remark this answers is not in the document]
dpsmith: This is beyond the scope of this work, but it would be good if we expanded the dbus auth cookie code to be able to put it in a location accessible by domains that would like to use dbus. Something to consider when suggesting domstore is that domstore is really a dbus service and thus you are publishing the auth cookie, who's premise is out-of-band exchange, over dbus.

recipes-connectivity/networkmanager/networkmanager-1.4.4/fix-device-type-property-exposure.patch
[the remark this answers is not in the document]
jandryuk: @mahantesh-ais I was more specifically asking why you needed `coerce_type`.  It seems to list every `NM_DEVICE_TYPE_*`, so it's not immediately obvious what it's converting.  Does `type` start as -1 when uninitialized and you need to coerce that to `NM_DEVICE_TYPE_UNKNOWN`?

recipes-connectivity/networkmanager/networkmanager-1.4.4/dbus-system.conf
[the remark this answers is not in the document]
jandryuk: Oh, I didn't realize `DBUS_COOKIE_SHA1` changed. Yes, that doesn't make it a good fit.

recipes-connectivity/networkmanager/networkmanager-1.4.4/fix-device-type-property-exposure.patch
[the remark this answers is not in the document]
mahantesh-ais: Oh okay, my bad, I misunderstood the question. `coerce_type` is to fix reporting of `NM_DEVICE_TYPE_UNKNOWN`, as per the setting this type is never set.

eric-ch:
Build [$10035](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10035)

conf/distro/openxt-main.conf
```diff
@@ -57,7 +57,7 @@ PREFERRED_PROVIDER_console-tools = "console-tools"
 PREFERRED_PROVIDER_virtual/libx11 = "libx11"
 
 # Force specific package versions
-PREFERRED_VERSION_networkmanager = "0.9.8.%"
+PREFERRED_VERSION_networkmanager = "1.4.4.%"
```
eric-ch: `networkmanager` version is `1.4.4`, this will not match, it should be `1.4.4` or `1.4.%` if the goal is to follow micro releases of the `1.4` line.
```
preferred version 1.4.4.% of networkmanager not available (for item networkmanager)
```

recipes-connectivity/networkmanager/networkmanager-1.4.4/xc-nutty-network.patch
eric-ch: Thanks, the delta between the metrics would not change the order anyway.
```
default via 192.168.0.1 dev brbridged proto static metric 425 
default via 192.168.0.1 dev wlan0 proto static metric 600 
```

recipes-connectivity/networkmanager/networkmanager-1.4.4/update-resolv-conf.patch
```diff
@@ -0,0 +1,176 @@
...
+-		update_resolv_conf (self, searches, nameservers, options, NULL, NM_DNS_MANAGER_RESOLV_CONF_MAN_UNMANAGED);
++		update_resolv_conf (self, RESOLV_CONF, searches, nameservers, options, NULL, NM_DNS_MANAGER_RESOLV_CONF_MAN_UNMANAGED);
+ 
+ 	/* signal that resolv.conf was changed */
+ 	if (update && result == SR_SUCCESS)
+@@ -1286,6 +1291,89 @@ forget_data (NMDnsManager *self, NMDnsIP
+ }
+ 
+ static gboolean
++xc_rewrite_resolv_conf (NMDnsManager *self,
```
eric-ch: The return value of this function is incorrect and triggers warnings (see further comment):
```
Jan 13 23:01:07.885387 VM Network (6): NetworkManager: <trace> [1578956467.8698] dns-mgr: update-resolv-conf: write internal file /var/volatile/etc/resolv.conf.brbridged succeeded (rc-manager=file)
Jan 13 23:01:07.885397 VM Network (6): NetworkManager: <warn>  [1578956467.8698] dns-mgr: Could not commit DNS changes: (-1) (unknown)
```

recipes-connectivity/networkmanager/networkmanager-1.4.4/update-resolv-conf.patch
```diff
@@ -0,0 +1,176 @@
...
++       } else
++               g_ptr_array_free (rc.options, TRUE);
++
++       nis_domain = rc.nis_domain;
++
++       strcpy(resolv_path, RESOLV_CONF);
++       strcat(resolv_path, ".");
++       strcat(resolv_path, iface);
++
++       if (success == FALSE)
```
eric-ch: `success` is initialized to `FALSE` and is not used until it is tested here. No need to test.

recipes-connectivity/networkmanager/networkmanager-1.4.4/update-resolv-conf.patch
```diff
@@ -0,0 +1,176 @@
...
++               success = update_resolv_conf (self, resolv_path, searches, nameservers, options, error, NM_DNS_MANAGER_RESOLV_CONF_MAN_FILE);
```
eric-ch: This function does not return a `gboolean`, yet `success` is used as one.
```
static SpawnResult
update_resolv_conf (NMDnsManager *self,
                    const char *resolv_path,
                    char **searches,
                    char **nameservers,
                    char **options,
                    GError **error,
                    NMDnsManagerResolvConfManager rc_manager)
```
It should be compared to the enum value `SR_SUCCESS`
```
typedef enum {
	SR_SUCCESS,
	SR_NOTFOUND,
	SR_ERROR
} SpawnResult;
```

recipes-connectivity/networkmanager/networkmanager-1.4.4/update-routing-tables.patch
```diff
@@ -0,0 +1,128 @@
+#OpenXT:Update the routing tables for different bridges as setup by network-slave.
+#       These routing tables are reffered when routing packets for different networks.
+--- a/src/nm-policy.c
++++ b/src/nm-policy.c
+@@ -472,8 +472,10 @@ update_ip4_routing (NMPolicy *self, gboo
+ 	}
+ 	g_assert ((best || vpn) && best_ac);
+ 
++#if 0
```
eric-ch: Should be removed by the patch instead of defined out.

conf/distro/openxt-main.conf
mahantesh-ais: Yes, I will change it to `1.4.4` as I think our next goal would be to upgrade against zeus. Thanks

recipes-connectivity/networkmanager/networkmanager-1.4.4/fix-eth0-slave-issues.patch
```diff
@@ -0,0 +1,65 @@
...
+ 		nm_device_master_release_one_slave (self, slave, configure, reason);
+ 		/* Bridge/bond/team interfaces are left up until manually deactivated */
+ 		if (priv->slaves == NULL && priv->state == NM_DEVICE_STATE_ACTIVATED)
+--- a/src/devices/nm-device-ethernet-utils.c
++++ b/src/devices/nm-device-ethernet-utils.c
+@@ -31,6 +31,7 @@ nm_device_ethernet_utils_get_default_wir
+ 	guint j;
+ 	int i;
+ 
++	return NULL;
```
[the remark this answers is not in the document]
mahantesh-ais: Ack, I will remove the dead code.

recipes-connectivity/networkmanager/networkmanager-1.4.4/update-routing-tables.patch
mahantesh-ais: Ack, will remove it in the patch.

recipes-connectivity/networkmanager/networkmanager-1.4.4/update-resolv-conf.patch
mahantesh-ais: Yes, this is wrong, I will change it appropriately. Thanks

recipes-connectivity/networkmanager/networkmanager-1.4.4/update-resolv-conf.patch
```diff
@@ -0,0 +1,176 @@
...
++	const char *rc_path = resolv_path;
+ 	nm_auto_free char *rc_path_real = NULL;
+ 
+ 	/* If we are not managing /etc/resolv.conf and it points to
+@@ -834,7 +839,7 @@ update_resolv_conf (NMDnsManager *self,
+ 		return SR_ERROR;
+ 	}
+ 
+-	if (symlink (MY_RESOLV_CONF, RESOLV_CONF_TMP) == -1) {
++	if (symlink (RESOLV_CONF, RESOLV_CONF_TMP) == -1) {
```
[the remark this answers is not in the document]
mahantesh-ais: Yes, this is never reached.
Yes, `rc_manager` is set to `NM_DNS_MANAGER_RESOLV_CONF_MAN_FILE`, because we are writing to `/var/volatile/etc/resolv.conf.XXX` (XXX being the name of an interface), which is a regular file.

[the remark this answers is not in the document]
mahantesh-ais: But I made this change thinking if `rc_manager` is ever set to `NM_DNS_MANAGER_RESOLV_CONF_SYMLINK`. I can remove this if you're thinking this will be redundant.

mahantesh-ais:
Build: [10038](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10038)

eric-ch:
Will merge soon.

mahantesh-ais:
> The glib .bbappend config would affect every user of glib. We typically use LD_PRELOAD to override libc for Argo, could you share some details on the glib use case for LD_PRELOAD? E.g. what approach was used by the previous version of NetworkManager to avoid the need for this bbappend?

In the upgraded version of network-manager-applet (nm-applet) GDBus programming is used to handle any DBus communication and GDBus relies on glib. Now, LD_PRELOAD is used to override symbols used internally in glib for Argo.

In the older version of nm-applet dbus-glib was used.

(quoting from : [https://developer.gnome.org/gio/stable/ch35.html](https://developer.gnome.org/gio/stable/ch35.html))
dbus-glib uses the libdbus reference implementation, GDBus doesn't. Instead, it relies on GIO streams as transport layer...

As I understand libdbus uses libc symbols which were overridden using LD_PRELOAD.

PR #1309, opened 2020-01-13
OXT-1732: dbus: limit build dependency to Xen argo header

stacktrust:
Limit OE recipe DEPENDS to socket definition in header file, instead
of libargo. Remove xen as build dependency of dbus. When xen changes,
avoid rebuilding dbus-dependent packages across multiple images.

Signed-off-by: Rich Persaud <rich.persaud@baesystems.com>

OXT-1306
OXT-1732

### DIFF
--- a/recipes-core/dbus/dbus_1.%.bbappend
+++ b/recipes-core/dbus/dbus_1.%.bbappend
@@ -5,10 +5,10 @@ DEPENDS_append_class-native += " \
 # DBus will not link with libargo in the native case.
 # The argo kernel headers cannot be expected in the native environment since
 # libargo depends on Xen, and the hypervisor headers are not separated from the
-# main recipe.
+# main recipe. Limit dependency to argo header.
 DEPENDS_append_class-target += " \
     libselinux \
-    libargo \
+    argo-module-headers \
 "
 FILESEXTRAPATHS_prepend := "${THISDIR}/patches:"
 


recipes-core/dbus/dbus_1.%.bbappend
```diff
@@ -5,10 +5,10 @@ DEPENDS_append_class-native += " \
 # DBus will not link with libargo in the native case.
 # The argo kernel headers cannot be expected in the native environment since
 # libargo depends on Xen, and the hypervisor headers are not separated from the
-# main recipe.
+# main recipe. Limit dependency to argo header.
 DEPENDS_append_class-target += " \
     libselinux \
-    libargo \
+    argo-module-headers \
```
jandryuk: This isn't doing what you think it is.

argo.patch patches configure to look for libargo.h and determine family=argo support.  libargo.h is packaged in libargo - not argo-module-headers - so changing the dependency removes libargo.h from the sysoot.  libargo.h is not found and autoconf doesn't enable the family=argo (PF_XENARGO) support.

FYI, the configure output is wrong since:
Building argo support:     ${enable_argo}
should be
Building argo support:     ${have_argo}

but you could check where argo support is compiled with:
`grep ARGO tmp-glibc/work/core2-64-oe-linux/dbus/1.10.14-r0/build/config.h`

stacktrust: Thanks for the review.  Will provide fix(es).

stacktrust: How does the OpenXT dbus recipe current enable argo support without passing --enable-argo to configure?

eric-ch: It is "enabled" [by default](https://github.com/OpenXT/xenclient-oe/blob/master/recipes-core/dbus/patches/argo.patch#L9)
```
AC_ARG_ENABLE(argo, AS_HELP_STRING([--enable-argo],[build with argo support]),enable_argo=$enableval,enable_argo=auto)
```
[AC_ARG_ENABLE](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Package-Options.html) is setting `enable_argo` to `auto` if `--enable-argo` is not passed to configure. The rest of the configure tests if argo is disabled against `$enable_argo = "no"`.

Unless `--enable-argo=no` is given, or if `libargo.h` is not found, HAVE_ARGO will be `1`.

PR #1312, opened 2020-01-23
busybox: Enable poweroff command

jandryuk:
The stubdom uses the busybox poweroff command to shutdown, but it is
currently missing.  In that case, the /init script ends and the kernel
panics instead of shutting down.  Install poweroff to restore graceful
shutdown.

### DIFF
--- a/recipes-core/busybox/files/poweroff.cfg
+++ b/recipes-core/busybox/files/poweroff.cfg
@@ -1,0 +1,2 @@
+# Used by stubdom
+CONFIG_POWEROFF=y
--- a/recipes-core/busybox/busybox_%.bbappend
+++ b/recipes-core/busybox/busybox_%.bbappend
@@ -10,6 +10,7 @@ SRC_URI += " \
     file://login.cfg \
     file://misc.cfg \
     file://netutils.cfg \
+    file://poweroff.cfg \
     file://process.cfg \
     file://shells.cfg \
     file://syslog.cfg \


eric-ch:
Build [$10043](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10043)

recipes-core/busybox/files/poweroff.cfg
```diff
@@ -0,0 +1,2 @@
+# Used by stubdom
+CONFIG_POWEROFF=y
```
eric-ch: Nit: `CONFIG_POWEROFF` is defined under `Init Utilities` in the `defconfig`, so it would be defined in `init.cfg` if we follow OE upstream standard for configuration artefacts.

jandryuk: Oh, I didn't know about the convention.  That makes sense.

Hmmm, one wrinkle is oe-core already has `init.cfg` which is conditionally included:
`${@["", "file://init.cfg"][(d.getVar('VIRTUAL-RUNTIME_init_manager') == 'busybox')]}`

I don't think we want to set, `VIRTUAL-RUNTIME_init_manager`.  I guess we could just extended SRC_URI with `init.cfg`.  What do you want?

eric-ch: That would conflict if we ended up fiddling with `VIRTUAL-RUNTIME_init_manager`...
I think the easier is to keep `poweroff.cfg` for now until we end up using `VIRTUAL-RUNTIME_init-manager` for real.

eric-ch:
Will merge soon.

PR #1316, opened 2020-02-04
debootstrap: Remove defunct recipe.

eric-ch:
Former recipe to force a staging to build debian packages, or so it seems. This is no longer useful and OE can generate Debian packages in a much cleaner way should the need arise.


eric-ch:
Build [$10047](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10047)

jandryuk:
Will merge soon

PR #1317, opened 2020-02-04
grub-efi: Amend FILESEXTRAPATH to use local cfg

eric-ch:
The cfg file is used by grub-mkimage to let grub know where to find its configuration.
For the iso installer, the default one pointing to the ESP for grub.cfg will not work.


eric-ch:
Build [$10047](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10047)

jandryuk:
Will merge soon

jandryuk:
Is it worth amending the commit message to state symptom - UEFI USB booting goes to a grub prompt instead of launching into the installer?

eric-ch:
@jandryuk Probably, it is fairly obscure at first glance.

eric-ch:
#### v2:
- Commit message amended.

jandryuk:
Thanks

PR #1318, opened 2020-02-10
module-compat: Replace by module.bbclass

eric-ch:
Remove module-compat.
This class will not work with EXTERNALSRC. It can be avoided with more
careful Makefiles for out-of-tree modules and does not provide
equivalent configuration to module.bbclass.

#### Depends on:
- https://github.com/OpenXT/xenfb2/pull/9
- https://github.com/OpenXT/fbtap/pull/5


eric-ch:

jandryuk:
Will merge soon

PR #1320, opened 2020-02-10
carrier-detect: git PV and patch to repo

eric-ch:
Apply the patch to the repository. There is little to no reason to keep
patches against OpenXT repositories in the OpenXT layer.
This is tracking the tip of the repository, so version as _git, which in
turn will set PV from AUTOREV.

#### Depends on:
- https://github.com/OpenXT/xctools/pull/62


eric-ch:

jandryuk:
Will merge soon

PR #1321, opened 2020-02-10
dbd: Unify and rewrite recipe

eric-ch:
dbd and db-tools are both built from the same repository and from the
same path in the directory. The OCaml toolchain does not separate source
and build artifacts. This leads to a world of troubles when using
externalsrc since both rely on the local findlib database.

There is little reason dbd and db-tools are built by separate recipes.
Unify all that and take the opportunity to fix the following:
- db-cat is overwritten by db-write due to a typo in the recipe
- remove haskell.bbclass, nothing uses it
- Use B where appropriate, but...
- brokensep (B = ${S}) is still required as the Makefile references
  autogen directly
- Fix RDEPENDS on libargo


eric-ch:

recipes-openxt/dbd/dbd_git.bb
```diff
@@ -9,48 +10,103 @@ DEPENDS = " \
...
+    ${bindir}/db-cmd \
+    ${bindir}/db-exists \
+    ${bindir}/db-ls \
+    ${bindir}/db-nodes \
+    ${bindir}/db-read \
+    ${bindir}/db-rm \
+    ${bindir}/db-write \
+    ${bindir}/db-cat \
+"
+RDEPENDS_${PN}-tools += "bash"
```
jandryuk: db-$foo and db-$foo-dom0 are both `#!/bin/sh`, so I don't think there is a bash RDEPENDS for `${PN}-tools`.  dbd.initscript is `#!/bin/bash` with bashisms though.

Other than that, LGTM.

eric-ch:
#### v2:
- Removed `RDEPENDS` on `bash` for `dbd-tools`

jandryuk:
Will merge soon

PR #1322, opened 2020-02-10
idl: Fetch files from ${S}

eric-ch:
When installing the RPC .xml descriptions, they are fetched from the
sources, which should be ${S}, not ${WORKDIR}/git.
This breaks EXTERNALSRC otherwise.


eric-ch:

jandryuk:
Will merge soon

PR #1326, opened 2020-02-11
OXT-1489 : cryptsetup: set parameters for swap file encryption

stacktrust:
Change swap file encryption from default cipher (plain dm-crypt) to the
cipher used for the config partition (aes-xts-plain64). Options specified
here are passed from cryptdisks.initscript to cryptsetup on each boot.

OXT-1489

Signed-off-by: Rich Persaud <rich.persaud@baesystems.com>

### DIFF
--- a/recipes-openxt/xenclient-cryptdisks/xenclient-cryptdisks/crypttab
+++ b/recipes-openxt/xenclient-cryptdisks/xenclient-cryptdisks/crypttab
@@ -1,4 +1,4 @@
 # <target name>	<source device>		<key file>                      <options>
 log             /dev/xenclient/log      /config/keys/log.key
 cores           /dev/xenclient/cores    /config/keys/cores.key
-swap            /dev/xenclient/swap     SWAP
+swap            /dev/xenclient/swap     SWAP                            --cipher aes-xts-plain64 --key-size 512 --hash sha512


recipes-openxt/xenclient-cryptdisks/xenclient-cryptdisks/crypttab
```diff
@@ -1,4 +1,4 @@
 # <target name>	<source device>		<key file>                      <options>
 log             /dev/xenclient/log      /config/keys/log.key
 cores           /dev/xenclient/cores    /config/keys/cores.key
-swap            /dev/xenclient/swap     SWAP
+swap            /dev/xenclient/swap     SWAP                            --cipher aes-xts-plain64 --key-size 512 --hash sha512
```
eric-ch: If, like myself, your first thought it "How is this possibly working?". [Here](https://github.com/OpenXT/xenclient-oe/blob/master/recipes-openxt/xenclient-cryptdisks/xenclient-cryptdisks/cryptdisks.initscript) is why.
This does not follow [crypttab(5)](https://www.freedesktop.org/software/systemd/man/crypttab.html).

While systemd has [systemd-cryptsetup-generator](https://www.freedesktop.org/software/systemd/man/systemd-cryptsetup-generator.html), for sysvinit I am only aware of [cryptdisk_start](https://packages.debian.org/sid/cryptsetup) which seems to be provided by Debian.

It would be nice to move away from this custom initscript if there is a replacement somewhere.

stacktrust: Had a similar reaction to learning that we have a bespoke implementation that combines a "standard" name with similar-but-different file format.  This PR is compatible with the current implementation. Will file a ticket to look for a more standard alternative.

stacktrust:
> Why not also change log and cores? I guess that would introduce some backward compatibility issues unless you can convert on the fly.

Yes they should be changed also, just added a note to the ticket.  But that requires changes to the init script and more test coverage that I currently have available.  This change is relatively self-contained and low risk, because swap is recreated on each boot.

eric-ch:
Build [$10050](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10050)

eric-ch:
Will merge soon.

PR #1328, opened 2020-02-12
tpm2-tss: OXT-1689: remove bash-isms from recipe

dozylynx:
Fixes the OpenXT build on Devuan Beowulf with default shell.

### DIFF
--- a/recipes-security/tss2/tpm2-tss_2.0.0.bb
+++ b/recipes-security/tss2/tpm2-tss_2.0.0.bb
@@ -28,7 +28,7 @@ FILES_resourcemgr = " \
 do_configure_prepend () {
     # Creates the src_vars.mk file used by automake to handle source-files for
     # each component. Modified to not call autotools and let OE handle that.
-    pushd ${S}
+    cd ${S}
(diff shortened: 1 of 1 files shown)


eric-ch:
Build [$10053](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10053)

dozylynx:
> The change is fine.
> 
> Devuan's default is dash?

$ ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Jan 17  2019 /bin/sh -> dash

> Bash as /bin/sh is expected elsewhere, I imagine:
> https://github.com/OpenXT/openxt/blob/d46013674edfa10bd4c694cbf03e98b02ea8867a/build-scripts/oe/setup.sh#L68 Have you run into other issues?

Not so far, no.
I've been focussed on building some dom0 components so it is possible that there are some more to be found.

jandryuk:
Thanks.  /bin/sh -> bash may be more specific to the scripts in openxt/openxt.

eric-ch:
Will merge soon.

PR #1330, opened 2020-02-14
rpcgen: Fix paths in do_install

eric-ch:
Copy files from $B an $S where appropriate.

Referencing `$WORKDIR` does not sit well with `externalsrc` and should be avoided.


eric-ch:
Build[$10057](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10057)

jandryuk:
Will merge soon.

PR #1334, opened 2020-02-19
qemu-dm: Wire up chardev-argo domid option

jandryuk:
The domid option is not current allowed since it is not listed in
qemu_chardev_opts.  qemu fails with "Invalid parameter 'domid'".

Add domid into qemu_chardev_opts so it is accepted as a option.

We also need to bind a connected ring pointed at the remote domid.
Otherwise when the remote tries to reply, it'll fail since the ring is
bound to dom0.

hypervisor: (XEN) d23v0 argo: vm23 connection refused, src (vm23:13ed) dst (vm32:3afd)
VM Windows10-dm (32): qemu-system-i386: [argo-chardrv] failed to connect argo socket - err: 111

### DIFF
--- a/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
+++ b/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
@@ -304,7 +304,7 @@
 +        return;
 +    }
 +
-+    if (argo_bind(fd, &s->local_addr, 0)) {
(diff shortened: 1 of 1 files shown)


eric-ch:
Build [$10064](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10064)

eric-ch:
Will merge soon.

PR #1335, opened 2020-02-19
tpm-functions: Fix tpm_is_owned for tpm2

jandryuk:
This function, on an un-owned tpm2, throws an error:
[: -eq: unary operator expected

When the tpm2 is un-owned, `grep -q` return 1.  The conditional is
skipped, and state is still "" when it gets to the inversion on the way
out of the function.  Initialize the state to 0 to avoid this issue.

The tpm1 path is fine since it always sets state.


eric-ch:
Build [$10064](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10064)

eric-ch:
Will merge soon.

PR #1338, opened 2020-02-20
OXT-1748: gene3fs: Remove associated patches and recipes

eric-ch:
`vhd.patch` applied to `e2fsprogs` introduces the option -Z, which is no longer used in OpenXT: OpenXT/openxt@b5797f1.

Removing that patch ends up removing the only consumer of gene3fs and libbudgetvhd provided in https://github.com/OpenXT/gene3fs, so remove these recipes as well.


stacktrust:
OXT-1748 

We can move the gene3fs repo to https://github.com/openxt-extras

eric-ch:
##### V2;
- Flagged the commits with the ticket number.

jandryuk:
Will merge soon.

PR #1340, opened 2020-02-26
Refactored Xen recipes : OXT-1694

dozylynx:
This PR is a RFC to enable code review and testing. It has currently been compile-tested only, prior to being rebased onto the current OpenXT xenclient-oe master branch.

In this PR, the Xen recipes for OpenXT have been rewritten for compatibility with the refactored recipes for the meta-virtualization layer. A new Linux Argo header has been introduced to enable breaking a dependency cycle and allow for the separate xen-libxl recipe to be retired and its function be provided by the xen-tools recipe. The Xen recipes are made compatible with python3.

To support building and testing of the new OpenXT Xen recipes, copies of the proposed new meta-virtualization recipes have been applied in a new dynamic-layers directory within this xenclient-oe repository: this enables compatibility with the Zeus branch meta-virtualization layer currently used by OpenXT, while allowing the new OpenXT Xen recipes to work with the new meta-virtualization Xen recipes. This can be changed to an alternative method of obtaining the new meta-virtualization Xen recipes, according to the preference of the OpenXT community.

To reduce the volume of change in this series, the Xen patches have not been rebased to handle patch fuzz.

See also corresponding required changes in these repos:
- [linux-xen-argo # 7](https://github.com/OpenXT/linux-xen-argo/pull/7)
- [xctools # 63 ](https://github.com/OpenXT/xctools/pull/63)


recipes-extended/seabios/patches/only-boot-selected-devices.patch
```diff
@@ -71,11 +75,11 @@ PATCHES
 -                  , desc ? MAKE_FLATPTR(seg, desc) : "Unknown");
 -    DefaultBEVPrio = DEFAULT_PRIO;
 +    if (SelectedBEV) {
-+        bootentry_add(IPL_TYPE_BEV, defPrio(prio, DefaultBEVPrio)
-+                      , SEGOFF(seg, bev).segoff
-+                      , desc ? MAKE_FLATPTR(seg, desc) : "Unknown");
-+        DefaultBEVPrio = DEFAULT_PRIO;
-+    }
++		bootentry_add(IPL_TYPE_BEV, defPrio(prio, DefaultBEVPrio)
++					  , SEGOFF(seg, bev).segoff
```
jandryuk: Your refreshed patch introduces hardtabs.  Please revert them to spaces to match the rest of the file.  The changes add noise and makes reviewing a little harder.  Thanks.

recipes-support/lvm2/lvm2_%.bbappend
```diff
@@ -22,6 +26,10 @@ do_install_append() {
     install -m 0644 ${WORKDIR}/volatiles.99_cryptsetup ${D}${sysconfdir}/default/volatiles/99_cryptsetup
 }
 
+ldconfig_postinst_fragment() {
+}
+ALTERNATIVE_lvm2_remove = "lvm.conf"
```
eric-ch: That should not be a problem with the following:
```
PACKAGES =+ "${PN}-conf"
RRECOMMENDS_${PN}_append += "${PN}-conf"

FILES_${PN}-conf = " \
    ${sysconfdir}/lvm/lvm.conf \
"
```
I do not see `ALTERNATIVE` definitions in `meta-openembedded` either, what reason triggered this change?

recipes-openxt/xenclient-toolstack/xenclient-toolstack_git.bb
```diff
@@ -2,8 +2,8 @@ DESCRIPTION = "XenClient toolstack"
 LICENSE = "LGPLv2.1"
 LIC_FILES_CHKSUM = "file://COPYING;md5=321bf41f280cf805086dd5a720b37785"
 DEPENDS += "ocaml-dbus camomile xen xz"
-RDEPENDS_${PN} = "xen-xenstore xen-xenstored"
-RDEPENDS_${PN}_xenclient-ndvm += " dbd-tools-vm"
+RDEPENDS_${PN} = "xen-tools-xenstore virtual/xenstored"
+RDEPENDS_${PN}_xenclient-ndvm += " db-tools-vm"
```
eric-ch: `db-tools-vm` is `dbd-tools-vm` since https://github.com/OpenXT/xenclient-oe/pull/1321, seems like a rebase induced error.

recipes-extended/seabios/patches/only-boot-selected-devices.patch
dozylynx: ack - that was not intended. I've just refreshed this branch to with the hardtabs removed.

recipes-extended/xen/xen-common.inc
```diff
@@ -1,4 +1,4 @@
-FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+FILESEXTRAPATHS_prepend := "${THISDIR}/../../../meta-virtualization/recipes-extended/xen/files:${THISDIR}/files:"
```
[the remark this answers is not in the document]
dozylynx: > If that is specific to `xen-ocaml-libs`, it probably belongs in that recipe.

It could go there but the intent was to make sure in an obvious way that the search path for the SRC_URI is the same for each of the OpenXT Xen recipes to avoid surprises. Do you need it to be moved or does this justify it being there?

> Why are the OCaml pieces requiring meta-virt patches if meta-virt does not support OCaml btw? Headers I presume? (introduced by [f439396](https://github.com/OpenXT/xenclient-oe/commit/f439396ec678bec79daa9c98b7ff80ddecfaee8e), which seems to point at it being ocaml specific)

IIRC it's build system stuff: there are python script fixes needed to make the build happy since python2 is not used in the build now.

conf/layer.conf
```diff
@@ -26,6 +28,11 @@ LAYERSERIES_COMPAT_xenclient-oe = "zeus"
 
 LICENSE_PATH += " ${LAYERDIR}/files/additional-licenses/"
 
+BBFILES_DYNAMIC += " \
+    virtualization-layer:${LAYERDIR}/dynamic-layers/meta-virtualization/*/*/*.bb \
+    virtualization-layer:${LAYERDIR}/dynamic-layers/meta-virtualization/*/*/*.bbappend \
+"
```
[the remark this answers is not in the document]
dozylynx: I'm flexible on which mechanism we agree to go with for having backport recipes and patches being made available, and also on the timing of how to bring these new `meta-virtualization` Xen recipes into OpenXT.

This was added to this series in support of it being a "RFC PR" and to enable testing while leaving our meta-virtualization layer as is for now.

This dynamic-layer mechanism is new to me too: I don't think it disables the referenced layer entirely - my understanding is that it just allows for selective recipe override, so in this case it should only replace Xen, iPXE and Seabios from the Zeus branch of meta-virt.

conf/distro/openxt-main.conf
```diff
@@ -100,3 +104,24 @@ STUBDOMAIN_KERNEL ?= "bzImage"
 
 # Set our root home
 ROOT_HOME = "/root"
+
+# multilib is required for building Xen's hvmloader, which needs
```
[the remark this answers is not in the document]
dozylynx: There are 32-bit header files needed by the hvmloader build that are provided by 32-bit glibc. multilib is how OE makes the 32-bit glibc available to build with.

recipes-extended/qemu-dm/qemu-dm/atapi-pass-through.patch
```diff
@@ -1073,7 +1073,7 @@ Ross Philipson, <philipsonr@ainfosec.com>, 10/13/2016
 +
 +#define ATAPI_CDROM_PORT 5000
 +#define ARGO_ATAPI_PT_RING_SIZE \
-+  (XEN_ARGO_ROUNDUP((((4096)*64) - sizeof(xen_argo_ring_t)-XEN_ARGO_ROUNDUP(1))))
++  (XEN_ARGO_ROUNDUP((((4096)*64) - (8 * XEN_ARGO_ROUNDUP(1)))))
 +
 +#define MAX_ARGO_MSG_SIZE (ARGO_ATAPI_PT_RING_SIZE)
```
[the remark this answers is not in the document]
dozylynx: Agree that there is something to look at here; there's discussion on the related `xctools` and `linux-xen-argo` PRs.
My work on this is on hold for a little while.

recipes-extended/xen/xen-openxt.inc
```diff
@@ -0,0 +1,61 @@
...
+do_configure() {
+    echo "debug := n" > .config
+    echo "XSM_ENABLE := y" >> .config
+    echo "FLASK_ENABLE := y" >> .config
+
+    cp "${WORKDIR}/defconfig" "${B}/xen/.config"
+    #Define CONFIG_TXT_OP in the hypervisor build to export tboot evtlog data
+    #It's stubbed out for the pv-shim since it's not supported, but uses the
+    #same hypercall headers
+    echo "CONFIG_TXT_OP=y" >> "${B}/xen/.config"
```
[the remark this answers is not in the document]
dozylynx: Let's fix that in a separate PR, separate from this refactoring.

recipes-extended/xen/xen-openxt.inc
```diff
@@ -0,0 +1,61 @@
...
+
+    # Enable argo
+    echo "CONFIG_ARGO=y" >> "${B}/xen/.config"
```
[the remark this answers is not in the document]
dozylynx: Let's fix that in a separate PR, separate from this refactoring.

recipes-extended/xen/xen-tools-blktap3.inc
```diff
@@ -0,0 +1,161 @@
...
+
+INITSCRIPT_PACKAGES =+ "${PN}-tapback "
+INITSCRIPT_NAME_${PN}-tapback = "tapback"
+INITSCRIPT_PARAMS_${PN}-tapback = "defaults 61"
+
+TARGET_CPPFLAGS += "-DTAP_CTL_NO_DEFAULT_CGROUP_SLICE -DOPEN_XT"
+
+do_configure_append() {
+    # blktap3 uses autotools but since we're splicing it into a subtree
+    # of the Xen tree which has its own autotools material, we need to
```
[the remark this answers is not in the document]
dozylynx: In order to resolve the cyclic dependency that the previous recipes suffered from, blktap has to be in-tree for the libxl recipe to be folded into the tools recipe.

recipes-extended/xen/xen-ocaml-libs.bb
```diff
@@ -1,84 +1,106 @@
...
+            = "blktap3" ]
+    then
+        # The xen-tools recipe includes the xen-tools-blktap3.inc which will
+        # cause ${S}/tools/blktap3 to be populated for that recipe; this recipe
+        # needs an alternative.
+        # We just need a Makefile there sufficient to pass this configure step.
+	mkdir -p ${S}/tools/blktap3
+        cat >${S}/tools/blktap3/Makefile <<EOF
+clean:
+EOF
```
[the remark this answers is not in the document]
dozylynx: As per prior comment, it's folded in to resolve the issue of cyclic dependency where libxl depends on blktap

recipes-extended/xen/xen-openxt.inc
```diff
@@ -0,0 +1,61 @@
...
+    --exec-prefix=/usr \
+    --prefix=/usr \
+    --host=${HOST_SYS} \
+    --with-systemd=${systemd_unitdir}/system \
+    --with-systemd-modules-load=${systemd_unitdir}/modules-load.d \
+    --disable-stubdom \
+    --disable-ioemu-stubdom \
+    --disable-pv-grub \
+    --disable-xenstore-stubdom \
+    --disable-rombios \
```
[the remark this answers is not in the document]
dozylynx: Let's fix that in a separate PR, separate from this refactoring.

recipes-support/lvm2/lvm2_%.bbappend
dozylynx: This issue is surfaced when multilib is enabled. I haven't tested your proposed alternative here - does it coexist ok with the `lvm2-conf-initramfs` OpenXT recipe?

recipes-openxt/xenclient-toolstack/xenclient-toolstack_git.bb
dozylynx: ack, have refreshed the branch to fix this

dozylynx:
My work on this is on hold for a little while.

conf/distro/openxt-main.conf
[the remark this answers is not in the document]
dozylynx: Xen README states:

> there are a number of prerequisites for building a Xen source release. Make sure you have all the following installed, either by visiting the project webpage or installing a pre-built package provided by your OS distributor:

```
    * Libc multiarch package (e.g. libc6-dev-i386 / glibc-devel.i686).
      Required when building on a 64-bit platform to build
      32-bit components which are enabled on a default build.

```
https://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=README;h=7f31e6b8174706713d551a3c655c3f9abf7b3c63;hb=a2e84d8e42c9e878fff17b738d8e5c5d83888f31#l65

The above instruction was introduced in 2015 here:
https://xenbits.xen.org/gitweb/?p=xen.git;a=commit;h=84b4f356db6a2181fda1a9cb17a876f6d773970b

eric-ch:
Build [$10077](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10077)

recipes-support/lvm2/lvm2_%.bbappend
eric-ch: `lvm2-conf-initramfs` was introduced for the above sleigh of hand (https://github.com/OpenXT/xenclient-oe/commit/8950739df & https://github.com/OpenXT/xenclient-oe/commit/104ddd73389740513d88ea3ea1b6ed918ae151bc)

recipes-extended/xen/xen-openxt.inc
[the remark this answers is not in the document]
eric-ch: This PR introduces building these components. Actually is this even built or just configure frobbing to pass the configuration step?

dozylynx:
I'm about to start looking at this again. I've just posted some work to update upstream meta-virtualization, including a Xen version uprev and Xen 4.14 is now approaching release which will likely involve some churn there too, so this PR will take some work to refresh, but I think the general approach of this PR will remain similar.

recipes-support/lvm2/lvm2_%.bbappend
dozylynx: ok, I now have an alternative proposal for this that doesn't do the above and still builds successfully: the lvm2 recipe has some multilib-aware logic and a small similar piece needs adding to the `lvm2-conf-initramfs` recipe:
- add `inherit multilib_script`
- add `MULTILIB_SCRIPTS = "${PN}:${sysconfdir}/lvm/lvm.conf"`

conf/layer.conf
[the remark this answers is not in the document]
dozylynx: This use of dynamic layers is not going to be needed and will not be included in the next version of this work.

recipes-extended/xen/xen-common.inc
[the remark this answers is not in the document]
dozylynx: This isn't needed with the current version of meta-virtualization and a ported xenclient-oe to use it.

recipes-core/images/xenclient-installer-part2-image.bb
```diff
@@ -47,3 +47,12 @@ IMAGE_PREPROCESS_COMMAND += " \
...
 LDCONFIGDEPEND = ""
+
+# With ${IMAGE_ROOTFS}/etc removed, prelink_image will fail, so bracket it
+# to provide a temporary etc directory:
+prelink_image_prepend() {
+    mkdir ${IMAGE_ROOTFS}/etc
+}
+prelink_image_append() {
+    rmdir ${IMAGE_ROOTFS}/etc
+}
```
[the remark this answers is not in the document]
dozylynx: This change will be removed from this PR and handled separately.

recipes-extended/xen/xen-openxt.inc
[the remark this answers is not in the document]
dozylynx: have removed the remove of disable-rombios in the lastest version of the PR branch

recipes-extended/xen/xen-openxt.inc
```diff
@@ -0,0 +1,61 @@
...
+    --disable-ocamltools \
+    --with-initddir=${INIT_D_DIR} \
```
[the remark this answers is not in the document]
dozylynx: replaced with an explicit `${sysconfdir}/init.d` which is easier to follow

recipes-extended/xen/xen-openxt.inc
```diff
@@ -0,0 +1,61 @@
...
+    --with-sysconfig-leaf-dir=default \
+    --with-system-qemu=/usr/bin/qemu-system-i386 \
```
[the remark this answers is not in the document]
dozylynx: yep

recipes-extended/xen/xen-tools-openxt.inc
```diff
@@ -0,0 +1,124 @@
...
+
+PACKAGES += " \
+    ${PN}-cmp-fd-file-inode \
+    ${PN}-toolstack-headers \
+    ${PN}-xenstored-c \
+    "
+
+PROVIDES =+ "virtual/xenstored"
+
+RPROVIDES_${PN}-toolstack-headers = "${PN}-toolstack-headers"
```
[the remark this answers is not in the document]
dozylynx: ack, dropped

recipes-extended/qemu-dm/qemu-dm/atapi-pass-through.patch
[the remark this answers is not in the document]
dozylynx: Switched to use the new `ARGO_RING_OVERHEAD` macro, defined for this purpose.

dozylynx:
PR branch refreshed and all comments addressed.

.gitignore
```diff
@@ -1,2 +1,3 @@
 *~
 patches/bitbake/*.APPLIED
+*.sw*
```
dpsmith, line 3: I know was likely to deal with issues you were having, but does it belong in this RFC PR?

conf/distro/openxt-main.conf
```diff
@@ -73,9 +73,13 @@ PREFERRED_PROVIDER_virtual/java-native = "jamvm-native"
...
-# Set the preferred version to c-xenstored.
-# xen-xenstored-c is a PROVIDE alias to xen.bb.
-PREFERRED_PROVIDER_xen-xenstored = "xen-xenstored-c"
+# Set the preferred version to C xenstored from xen-tools
+PREFERRED_PROVIDER_virtual/xenstored = "xen-tools"
+PREFERRED_RPROVIDER_virtual/xenstored = "xen-tools-xenstored"
+
+# The C and OCaml xenstored packages can be installed together
+# and this whitelisting is necessary to allow both to be built
+MULTI_PROVIDER_WHITELIST += " virtual/xenstored"
```
dpsmith, line 82: Why do we need to build both simultaneously?

recipes-core/images/xenclient-stubdomain-initramfs-image.bb
```diff
@@ -14,7 +14,7 @@ IMAGE_INSTALL = " \
     busybox \
     bridge-utils \
     initramfs-stubdomain \
-    xen-xenstore \
+    xen-tools-xenstore \
```
dpsmith, line 17: Wait, I though you were switching to virtual/xenstore? Shouldn't you include virtual/xenstore and then allow preferred provider to kick in?

recipes-extended/xen/xen-openxt.inc
```diff
@@ -10,7 +10,6 @@ EXTRA_OECONF_remove = " \
     --disable-ioemu-stubdom \
     --disable-pv-grub \
     --disable-xenstore-stubdom \
-    --disable-rombios \
```
dpsmith: nit: an explanation in the commit message of why we want rombios would be helpful

recipes-extended/xen/xen-openxt.inc
```diff
@@ -11,7 +11,7 @@ EXTRA_OECONF_remove = " \
     --disable-pv-grub \
     --disable-xenstore-stubdom \
     --disable-ocamltools \
-    --with-initddir=${INIT_D_DIR} \
+    --with-initddir=${sysconfdir}/init.d \
```
dpsmith, line 14: Why? This is the value that is set by the update-rc.d bbclass. Is it getting trampled elsewhere?

recipes-extended/xen/xen-openxt.inc
```diff
@@ -0,0 +1,60 @@
...
+
+    # do configure
+    oe_runconf EXTRA_CFLAGS_XEN_CORE="${EXTRA_CFLAGS_XEN_CORE}" \
+               EXTRA_CFLAGS_XEN_TOOLS="${EXTRA_CFLAGS_XEN_TOOLS}" \
+               PYTHON="${PYTHON}"
+}
+
+do_install_append() {
+    install -d ${D}/boot
+    ln -sf "`basename ${D}/boot/xen-*xc.gz`" ${D}/boot/xen-debug.gz
```
eric-ch, line 55: Maybe `XEN_VENDORVERSION` should be defined in the recipe, nit and out-of-scope though.

recipes-extended/xen/xen-tools-openxt.inc
```diff
@@ -0,0 +1,124 @@
...
+    ${sysconfdir}/udev/rules.d/xen-vbd3-backend.rules \
+    ${sysconfdir}/xen/scripts/block-vbd3 \
+"
+
+# Export local headers for external toolstack
+FILES_${PN}-toolstack-headers = "\
+    ${includedir}/xc_dom.h \
+    ${includedir}/xen/libelf/libelf.h \
+    ${includedir}/xen/libelf/elfstructs.h \
+    "
```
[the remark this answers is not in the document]
eric-ch: I'd like to echo that sentiment: Is there a reason to package `FILES_${PN}-toolstack-headers` outside of `-dev`?

recipes-extended/xen/xen-tools_git.bbappend
```diff
@@ -0,0 +1,12 @@
...
+LIC_FILES_CHKSUM = "file://COPYING;md5=bbb4b1bdc2c3b6743da3c39d03249095"
+
+# OpenXT's Xen recipes share a common patchqueue so reset SRC_URI
+SRC_URI = "git://xenbits.xen.org/xen.git;branch=${XEN_BRANCH}"
+
+require xen-common.inc
+require ${@bb.utils.contains('DISTRO_FEATURES', 'blktap2', 'xen-tools-blktap2.inc', 'xen-tools-blktap3.inc', d)}
+require xen-tools-openxt.inc
+
+DEFAULT_PREFERENCE = "1"
```
eric-ch, line 12: Not sure I understand why `DEFAULT_PREFERENCE` is set to `1` here, it shouldn't change anything afaict.

.gitignore
dozylynx: sure, will drop it

conf/distro/openxt-main.conf
dozylynx: We need both the `xen-tools` and `xen-ocaml-libs` recipes to be built and the two xenstored imlementations are built by those recipes. Without whitelisting this, the build will fail since it believes that the two packages collide.

recipes-core/images/xenclient-stubdomain-initramfs-image.bb
dozylynx: `xenstore` and `xenstored` are different packages, for the client and daemon respectively. Only `xenstored` is virtual.

recipes-extended/xen/xen-openxt.inc
dozylynx: this is removing a remove of a disable... so I think that means that the result is that we _don't_ have rombios

recipes-extended/xen/xen-openxt.inc
dozylynx: This was in response to feedback from @eric-ch on the last round, where he'd had to chase down where the definition for `INIT_D_DIR` was set and what it was. I don't actually care either way, so I'll drop this commit that touches this, and leave it as `INIT_D_DIR`.

recipes-extended/seabios/patches/only-boot-selected-devices.patch
```diff
@@ -112,8 +116,19 @@ PATCHES
...
+     if (GET_GLOBAL(PlatformRunningOn) & PF_QEMU) {
+         // We want short boot times.  But on physical hardware even
+         // the test unit ready can take several seconds.  So do media
+@@ -572,15 +583,21 @@ boot_add_cd(struct drive_s *drive, const char *desc, int prio)
+             free(extra);
+         }
+     }
 -    bootentry_add(IPL_TYPE_CDROM, defPrio(prio, DefaultCDPrio)
 -                  , (u32)drive, desc);
++#endif
```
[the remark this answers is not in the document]
dozylynx: np - will do

recipes-extended/xen/xen-common.inc
```diff
@@ -102,32 +122,13 @@ SRC_URI_append = " \
     file://ocaml-makefiles-sysroot.patch \
 "
 
-COMPATIBLE_HOST = 'i686-oe-linux|(x86_64.*).*-linux|aarch64.*-linux'
```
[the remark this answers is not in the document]
dozylynx: No, I am not building on ARM. That is an entertaining idea though.
I'm removing this line from the OpenXT recipe since I don't think there's a need for OpenXT's Xen recipes to explicitly diverge from the meta-virtualization definitions of COMPATIBLE_HOST.

recipes-extended/xen/xen-ocaml-libs.bb
```diff
@@ -1,84 +1,106 @@
...
-    file://oxenstored.conf \
-    "
+require recipes-extended/xen/xen.inc
+require xen-common.inc
+
+inherit ocaml findlib
 
 PACKAGES = " \
-    xen-xenstored-ocaml \
+    ${PN}-xenstored \
```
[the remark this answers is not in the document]
dozylynx: Aiming for consistency of approach to naming so the relevant recipe can be quickly identified from package name, hence: `<recipe>-<package>`.
Agree it's a mouthful though and if it weren't for the aforementioned reason, I wouldn't have picked that.

recipes-extended/xen/xen-ocaml-libs.bb
```diff
@@ -1,84 +1,106 @@
...
+
+do_configure() {
+    if [ ${@bb.utils.contains('DISTRO_FEATURES', 'blktap2', 'blktap2', 'blktap3', d)} \
+            = "blktap3" ]
+    then
+        # The xen-tools recipe includes the xen-tools-blktap3.inc which will
+        # cause ${S}/tools/blktap3 to be populated for that recipe; this recipe
+        # needs an alternative.
+        # We just need a Makefile there sufficient to pass this configure step.
+	mkdir -p ${S}/tools/blktap3
```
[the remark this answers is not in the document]
dozylynx: thanks, nice catch

recipes-extended/xen/xen-openxt.inc
```diff
@@ -0,0 +1,60 @@
...
+    --with-system-qemu=${bindir}/qemu-system-i386 \
+    --disable-qemu-traditional \
+"
+
+# Force the target architecture for the hypervisor build to be 64-bit
+# irrespective of the current MACHINE settings.
+# Also supply a custom vendor suffix.
+EXTRA_OEMAKE += " \
+    XEN_TARGET_ARCH=x86_64 \
+    XEN_VENDORVERSION=-xc \
```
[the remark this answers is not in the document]
dozylynx: It might be nice to swap that in, but I think this PR has enough churn in it already. I don't know what breaks if that gets changed here.

recipes-extended/xen/xen-openxt.inc
dozylynx: good observation but I'm going to avoid changing this here

recipes-extended/xen/xen-tools_git.bbappend
dozylynx: Since Xen 4.14 is now released, I'm about to swap the `DEFAULT_PREFERENCE` value in the upstream meta-virtualization Xen recipes, which will make the `xen_git.bb` and `xen-tools_git.bb` recipes non-default.

It'll only matter if OpenXT's Xen major version happens to match the major stable version in meta-virt, which at some point it hopefully will. The aim is to avoid any surprise accidental use of the non-"_git.bb" Xen recipes from meta-virt.

recipes-extended/xen/xen-openxt.inc
eric-ch: > That looks fine though, INIT_D_DIR is defined in updated-rc.d.bbclass to ${sysconfdir}/init.d.

I don't quite remember what I meant... Maybe because this `.inc` does not inherit `update-rc.d`?

recipes-extended/xen/xen-common.inc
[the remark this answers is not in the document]
eric-ch: > That is an entertaining idea though.

+1!

Agreed with not diverging, thanks!

recipes-extended/xen/xen-tools_git.bbappend
eric-ch: Makes sense.

recipes-extended/xen/xen-openxt.inc
dozylynx: oh... you're right, @eric-ch ! nice catch, thanks. I need to look at this.

conf/distro/openxt-main.conf
dpsmith: Ack

recipes-core/images/xenclient-stubdomain-initramfs-image.bb
dpsmith: Is this related to above where we are building both C and Ocaml to mix and match xenstore and xenstored implementations?

recipes-extended/xen/xen-openxt.inc
dpsmith: Then IMHO if we are touching/manipulating init in the `.inc` then this should inherit update-rc.d.

dozylynx: Patch just submitted to meta-virtualization here to clean this up. Moves the tools-specific configuration for the init script handling over into the tools recipe and out of the common include, so the hypervisor recipe doesn't need that variable defined:
https://lists.yoctoproject.org/g/meta-virtualization/message/5663

dozylynx:
I have been busy with other tasks since my last update but would like to unblock progress on this. I have limited time available this week as I'm attending a conference, but will take another look at this and hopefully determine what the next steps will be.

dozylynx:
I am close to having a new version of this work to push here. Since @nkrasnoff is targetting dunfell with this, and that branch of meta-virtualization doesn't include 0db3901a484, I need to look at the handling of the initddir setting again for this.

recipes-extended/xen/xen-openxt.inc
dozylynx: Looking into this now: the dunfell branch of meta-virtualization doesn't include my change above, so `xen.inc` there contains:
`    --with-initddir=${INIT_D_DIR} \
`
and `INIT_D_DIR` is not defined when `update-rc.d` is not included: in dunfell meta-virt, the xen-tools recipe has it set, but the xen (hypervisor) recipe does not. This `xen-open.inc` file is only included by the hypervisor recipe in OpenXT, not any of the others.

I'm going to drop this change since the existing logic does the right thing for setting the `EXTRA_OECONF` variable: it removes the `--with-initddir` setting for the Xen hypervisor recipe.

I've also just found that the `xen-ocaml-libs` recipe needs to inherit `update-rc.d` to set `INIT_D_DIR` correctly.

_(Note that the Zeus branch of meta-virtualization does not contain the split xen / xen-tools recipes, but the Dunfell branch of meta-virtualization builds Xen OK with the Zeus branch of the other upstream repos; hence looking at Dunfell as the stable target for meta-virtualization for this PR.)_

recipes-core/images/xenclient-stubdomain-initramfs-image.bb
dozylynx: > Is this related to above where we are building both C and Ocaml to mix and match xenstore and xenstored implementations?

No, this specific change is just an update after the packages were renamed as part of the recipe separation.

However, on a similar topic, I've just pushed a few added commits with cleanups to the handling of the C xenstore daemon and the preference selection config which determines which is installed. I think I've addressed what's needed for this here.

dozylynx:
ok, this should now be ready for further testing with: zeus branches of the upstream repos except for meta-virtualization, which should be dunfell

recipes-extended/xen/xen-openxt.inc
```diff
@@ -0,0 +1,61 @@
...
+
+    # Enable argo
+    echo "CONFIG_ARGO=y" >> "${B}/xen/.config"
+
+    unset CFLAGS
+
+    # do configure
+    oe_runconf EXTRA_CFLAGS_XEN_CORE="${EXTRA_CFLAGS_XEN_CORE}" \
+               EXTRA_CFLAGS_XEN_TOOLS="${EXTRA_CFLAGS_XEN_TOOLS}" \
+               PYTHON="${PYTHON}"
```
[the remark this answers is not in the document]
eric-ch: My bad, it is defined in `EXTRA_OEMAKE`, nothing to do with the configuration.

recipes-support/lvm2/lvm2-conf-initramfs_1.0.bb
```diff
@@ -7,6 +7,9 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 SRC_URI = "file://lvm.conf"
 
 inherit allarch
+inherit multilib_script
+
+MULTILIB_SCRIPTS = "${PN}:${sysconfdir}/lvm/lvm.conf"
```
jandryuk, line 12: This change seems unexpected.  I'd be interested to see the actual error message (and maybe it should be in the commit message).

I'm surprised this recipe is a problem since it is marked allarch.  When I look up MULTILIB_SCRIPTS, it says 

> "Whilst the package managers handle multilib ELF binaries well, they don't handle scripts in the *bindir directories well. This adds support for marking these up so that they can be handled using update-alternatives."

This is a conf file and not a script/ *bindir file.  The recipe isn't using update-alternatives since it's supposed to override the lvm2 RRECOMMENDS.  A different package is installed which populates the same file system name.

Oh, is the issue that since lvm2 is multilib compatible, it's looking for package `lvm2-conf-${MULTILIB_SUFFIX}`, but this recipe (lvm2-conf-initramfs) isn't providing that name?  The MULTILIB_SCRIPT takes care of that?

recipes-devtools/qemu/qemu_%.bbappend
```diff
@@ -0,0 +1,5 @@
+# This temporary override enables compatibility of older openembedded-core
+# with newer meta-virtualization + xenclient-oe.
+# When OpenXT's openembedded-core is uprev'd to a version that already includes this
+# defintion in meta/recipes-devtools/qemu/qemu.inc then this can be removed here.
+PACKAGECONFIG[xen] = "--enable-xen,--disable-xen,xen-tools,xen-tools-libxenstore xen-tools-libxenctrl xen-tools-libxenguest"
```
jandryuk, line 5: Is this change to qemu needed?  We use qemu-dm and qemu-dm-stubdom for our device emulators.

recipes-support/lvm2/lvm2_%.bbappend
```diff
@@ -6,6 +6,10 @@ SRC_URI += " \
     file://volatiles.99_cryptsetup \
 "
 
+# OpenXT does not use LVM thin-provisioning and dropping this dependency
+# accelerates the build by avoiding a lengthy compilation of boost.
+PACKAGECONFIG_remove = "thin-provisioning-tools"
```
jandryuk, line 11: This commit seems to be independent of the PR?

recipes-support/lvm2/lvm2-conf-initramfs_1.0.bb
jandryuk: Could it be that "${PN}-conf" should be replaced with a fixed string here: https://github.com/OpenXT/xenclient-oe/blob/d24250170500f42bdffb7e02a7361e03d0ccb05f/recipes-support/lvm2/lvm2_%25.bbappend#L25 ?

That may prevent multilib from changing the variable? (If that is the issue).

recipes-support/lvm2/lvm2_%.bbappend
dozylynx: Yes I think it is. Am running a build now with it removed and will drop it from the PR if it looks ok

recipes-devtools/qemu/qemu_%.bbappend
dozylynx: Ack, I don't think it's needed. An incremental build without it works fine, so I'll drop this commit.

recipes-support/lvm2/lvm2_%.bbappend
dozylynx: Builds fine without it, will drop the commit from this PR.

jandryuk: It's a perfectly fine commit to go in independently.

recipes-support/lvm2/lvm2-conf-initramfs_1.0.bb
dozylynx: I need to dig into this to refresh my understanding of it and will either update the commit message or replace the commit, depending on what I find. I have observed that the upstream lvm2 recipe inherits the `multilib_script` class and applies `MULTILIB_SCRIPTS` to the `lvm.conf` file, here:

https://git.openembedded.org/meta-openembedded/tree/meta-oe/recipes-support/lvm2/lvm2_2.03.02.bb?h=zeus&id=663493e98883ada2a13a6e873700ed0eda654bb3#n13

and the change in this commit is mirroring that into the `lvm2-config-initramfs` recipe.

The error observed without this change is:

```
     WARNING: xenclient-initramfs-image-1.0-r0 do_rootfs:
        lvm2.postinst returned 1, marking as unpacked only, configuration
        required on target.
    
     ERROR: xenclient-initramfs-image-1.0-r0 do_rootfs: Postinstall
        scriptlets of ['lvm2'] have failed. If the intention is to defer
        them to first boot, then please place them into
        pkg_postinst_ontarget_${PN} ().
        Deferring to first boot via 'exit 1' is no longer supported.
```

and the contents of the script that fails are:

```
        #!/bin/sh
        set -e
            update-alternatives --install /etc/lvm/lvm.conf lvm.conf /etc/lvm/lvm.conf-lib 10
        if [ x"$D" = "x" ]; then
            if [ -x /sbin/ldconfig ]; then /sbin/ldconfig ; fi
        fi
```

jandryuk: If you put the error message in the commit message, and the part about "lvm2 recipe inherits the multilib_script class and applies MULTILIB_SCRIPTS to the lvm.conf file", I think that is fine.  A comment in the file would be good, too since it's not obvious why its there without looking at the changelog.  Don't spend too much time on this.  Thanks.

dozylynx:
@jandryuk : have just pushed a new version to the PR branch to address your comments. Thanks for the reviews.

@eric-ch : am now looking into your comment about whether `xen-tools-openxt.inc` still needs to package `toolstack-headers` -- it doesn't look like it is necessary any more, since a quick grep in xenclient-oe doesn't indicate anything referencing it. Am running an incremental build now with that packaging removed, ie. build-testing without it.

dozylynx:
@eric-ch : Unfortuntately that build test fails:
`xenclient-oe/recipes-openxt/xclibs/libxchxenstore_git.bb:do_configure`
reports:
`| ghc: could not execute: /build/tmpfs/tmp/xen-refactor-oxt-1694-glibc/hosttools/gcc`

I haven't dug into it, but it looks like there's some work needed in xclibs or the haskell toolchain to fix that. Is this OK to merge without that?

stacktrust:
We need to get this merged soon, to avoid the cost of rebasing caused by months of new delay, and to gain the benefit of alignment with upstream meta-virt.  Please file tickets for build optimizations that have been identified as a side effect of this PR, and retitle the PR once it has moved beyond RFC state and is suitable for merge.

eric-ch:
@dozylynx That's fine, we can attend the `ghc` error you see later, it builds with the toolstack headers for now. This is the last unresolved issue, so I'm fine with this PR.

@stacktrust What cost of rebasing? This is up to date with OE Dunfell. Merging it in mater is already considered unstable as it crosses OE `zeus` and `dunfell` branches in our layer set. Should there be further upstream developments on the Xen recipes, this layer would still need to be updated.

stacktrust:
@eric-ch potential rebasing cost would be in the event of OpenXT changes, rather than OE.  Thanks for the help with this PR.

eric-ch:
@stacktrust Ah, my bad. Given this is part of the upgrade to dunfell (https://github.com/OpenXT/xenclient-oe/pull/1355), I think it will be the next logical step for master. See [OXT-1796](https://openxt.atlassian.net/browse/OXT-1796) for the tickets related to the upgrade. I moved [OXT-1694](https://openxt.atlassian.net/browse/OXT-1694) in there as well.

eric-ch (approved):

recipes-extended/xen/xen-tools-openxt.inc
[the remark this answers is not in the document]
dozylynx: I agree that the headers should be in the `-dev` package. I've filed this to track fixing that: https://openxt.atlassian.net/browse/OXT-1799

dozylynx:
@stacktrust : PR retitled and new tickets:

- [OXT-1799 : Fix libxchxenstore and drop xen-tools-toolstack-headers package](https://openxt.atlassian.net/browse/OXT-1799)
- [OXT-1800 : Fix Xen and Xen-tools recipes python3 dependency in meta-virtualization](https://openxt.atlassian.net/browse/OXT-1800)
- [OXT-1801: Accelerate OpenXT build by avoiding boost dependency](https://openxt.atlassian.net/browse/OXT-1801)

eric-ch:
Will merge soon.
[$zeus-quick-4](https://openxt.ainfosec.com/OpenXT/auto/zeus/zeus-custom-quick-legacy/4/)

jandryuk:
With this change, xenclient-oe will depend on the dunfell branch of meta-virt? Does anyone know if that can be expressed in conf/layer.conf?

@eric-ch how did you do your zeus-quick-4 build?

eric-ch:
@jandryuk Indeed, I used the dunfell branch of meta-virtualization which supports both `zeus` and `dunfell` in its conf/layer.conf (for now at least).

dozylynx:
@eric-ch : do you need a PR to openxt.git to set the revisions for the layers (submodules) to go with this change?

eric-ch:
@dozylynx That would be good if people are using this repository's build-scripts.

PR #1341, opened 2020-02-27
installer: Use do_deploy instead of IMAGE_POSTPROCESS to deploy files

eric-ch:
images should not have an SRC_URI or process local files. This is supposed to be handled by packages and the sstate.

#### V1 (Deprecated):
Introduce:
- network-installer-answerfiles_1.0.bb
- syslinux-pxelinux-conf_1.0.bb
- syslinux-isolinux-conf_1.0.bb
- grub-efi-installer-conf_1.0.bb
- network-answerfiles_1.0.bb

These recipes will provide the files that used to be shoved in the installer image to be then deployed in the DEPLOY_DIR_IMAGE.
Just as well, installing from image built rootfs to DEPLOY_DIR_IMAGE should be avoided as it does not use the sstate. Instead of using IMAGE_POSTPROCESS, inherit deploy and implement
do_deploy for each recipe to provide the expected deployment files.

#### V3:
Append syslinux, grub-efi and xenclient-installer, acms, tboot to have
each implement, or extend the existing, do_deploy task and have it ship
the files necessary to assemble the bootable media.

This replaces the use of IMAGE_POSTPROCESS in xenclient-installer-image
and leverages the sstate for the deployed resources.

#### All:

It also fixes a race while building `xenclient-installer-image`, `do_unpack` will copy files passed in `SRC_URI` in `${WORKDIR}`, but `do_rootfs` does not depend on it so even if the `noexec` flag is removed from the step, it is not assured to be executed before `do_rootfs`.
Some people may have seen `xenclient-installer-image` fail `do_rootfs` in postprocess because it could not find `pxelinux.cfg` for example. It can be resolved without this changeset (https://github.com/eric-ch/xenclient-oe/commit/f4be3b9719427816bcbbe424e34e6874258c7a6f), but perpetuates the postprocess work-around.

#### Depends on:
- https://github.com/OpenXT/openxt/pull/367
- https://github.com/apertussolutions/bordel/pull/18
... depending on your choice of build-scripts.


eric-ch:
Build [$10076](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10076)

recipes-devtools/syslinux/syslinux_6.%.bbappend
```diff
@@ -1,3 +1,26 @@
...
+    for bin in isolinux.bin ldlinux.c32 mboot.c32 libcom32.c32; do
+        install -m 644 "${D}${datadir}/syslinux/${bin}" "${DEPLOYDIR}/iso"
+    done
+    # ISO-hybrid packaging (see build-scripts).
+    install -m 644 "${D}${datadir}/syslinux/isohdpfx.bin" "${DEPLOYDIR}/"
+}
+do_deploy_class-native() {
+    :
+}
+do_deploy[cleandirs] = "${DEPLOYDIR}"
```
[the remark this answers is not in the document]
eric-ch: I think you are correct, `do_deploy[cleandirs]` should not be necessary. Looking at other recipes doing similary things:
- [grub-efi](https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-bsp/grub/grub-efi_2.04.bb) does not `cleandirs`,
- [ovmf](https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-core/ovmf/ovmf_git.bb#L223) does both.

eric-ch:
> [About do_install] I don't think we actually need them packaged

We don't, for some reason I assumed they had to be, but that's not true.

eric-ch:
Build [$10076](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10076)

eric-ch:
#### V2:
- Using `DEPENDS` in image is not great, but it does work.

#### V3:
- Refactor this PR.
Instead of creating independent recipes  that would not package files and only provide a `do_deploy` step, add relevant `.bbappend` with `do_deploy` implementation to provide the deployable files.

recipes-core/microcode/intel-microcode_%.bbappend
```diff
@@ -1,10 +1,19 @@
...
+# from /boot to the boot partition.
 do_install_append() {
-	install -d ${D}/boot
-	install ${WORKDIR}/microcode_${PV}.bin ${D}/boot/microcode_${PV}.bin
-	cd ${D}/boot
-	ln -sf microcode_${PV}.bin microcode_intel.bin
+    install -d "${D}/boot"
+    # intel-microcode recipe installs in ${WORKDIR}.
+    install "${WORKDIR}/microcode_${PV}.bin" "${D}/boot/microcode_${PV}.bin"
+    ln -sfr "${D}/boot/microcode_${PV}.bin" "${D}/boot/microcode_intel.bin"
```
jandryuk:
```suggestion
    install "${WORKDIR}/microcode_${PV}.bin" "${D}/boot/microcode_intel.bin"
```

recipes-core/microcode/intel-microcode_%.bbappend
```diff
@@ -1,10 +1,19 @@
...
 }
 
-FILES_${PN} += " /boot"
+# Override do_deploy to suit OpenXT existing bootstrap.
+do_deploy() {
+    # intel-microcode recipe installs in ${WORKDIR}.
+    install "${WORKDIR}/microcode_${PV}.bin" "${DEPLOYDIR}/microcode_${PV}.bin"
+    ln -sfr "${DEPLOYDIR}/microcode_${PV}.bin" "${DEPLOYDIR}/microcode_intel.bin"
```
jandryuk:
```suggestion
    install "${WORKDIR}/microcode_${PV}.bin" "${DEPLOYDIR}/microcode_intel.bin"
```

recipes-openxt/answerfiles/network-installer-answerfiles_1.0.bb
```diff
@@ -0,0 +1,33 @@
+SUMMARY = "OpenXT installer answerfiles for network boot."
```
jandryuk: Delete this file since it should be unused.

recipes-core/microcode/intel-microcode_%.bbappend
eric-ch: Don't we want the versioned file and the symlink?

recipes-openxt/answerfiles/network-installer-answerfiles_1.0.bb
eric-ch: Yes.
I think we should move the answer files in https://github.com/OpenXT/installer as well, where the other answerfiles are, but since this PR is already dragging along I would like to do it in a different one.

recipes-core/microcode/intel-microcode_%.bbappend
jandryuk: You are replacing this, so I think the version is unnecessary:
```
microcode_install() {
    cp -f ${IMAGE_ROOTFS}/boot/microcode_intel.bin ${DEPLOY_DIR_IMAGE}/microcode_intel.bin
}
```

recipes-openxt/answerfiles/network-installer-answerfiles_1.0.bb
jandryuk: Yes, but you already moved the answerfiles into the xenclient-installer do_deploy, so this recipe is unused?

eric-ch: Yes, that recipe can go.

eric-ch:
Build [$10081](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10081)

jandryuk:
Will merge soon

PR #1344, opened 2020-03-10
qemu: Fix initial qmp timeout issues

jandryuk:
VM startup is always delayed 5 seconds when libxl fails to connect to
QMP.  Logs show:
libxl_qmp.c:530:qmp_next:Domain 1:timeout
libxl_qmp.c:823:libxl__qmp_initialize:Domain 1:Failed to connect to QMP

The root cause is a mismatch between the argo chardev being connected,
but the qmp monitor still being closed.  qmp-helper defaults to open
(skipping sending "live"), argo chardev receives the message, but the
qmp monitor is disabled.  The request isn't processed and request times
out.  qmp helpers sends "dead" and the argo chardev is closed which
brings the monitor and chardev into sync.  When the next connection
comes in, argo transitions to connected and the monitor is started by
qemu_chr_be_event(chr, CHR_EVENT_OPENED).

The oneline fix is
-+    *be_opened = false;
++    *be_opened = true;

so connected and be_opened are synchronized from the beginning.

However, conceptually it's unusual to start open when you have the
live/dead protocol.  Change the code so datagram connections (used by
QMP) start connected=0 and be_opened=false and streams start connected=1
and be_opened=true.  qmp-helper must be changed to send an initial live
message now.

Depends on https://github.com/OpenXT/xctools/pull/64

### DIFF
--- a/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
+++ b/recipes-extended/qemu-dm/qemu-dm/qmp-argo-char-driver.patch
@@ -7,7 +7,7 @@
 +chardev-obj-$(CONFIG_XEN) += char-argo.o
 --- /dev/null
 +++ b/chardev/char-argo.c
-@@ -0,0 +1,354 @@
+@@ -0,0 +1,355 @@
 +/*
 + * QEMU System Emulator
 + *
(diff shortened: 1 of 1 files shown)


eric-ch:
Build[$10085](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10085)

eric-ch:
Build [$10095](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10095)

PR #1345, opened 2020-04-15
[xen] Update kconfig-grant-table patch

crogers1:
The patch context is now incorrect due to a change in upstream commit
  46bde0561b4a6eca4deee2dc1797086353e95be9. This fixes a paren error
  causing our patch to fail to apply.


eric-ch:
Tested in local build.
Will merge soon.

PR #1347, opened 2020-04-28
S9: ocaml: use sysroot for dynamic libraries

eric-ch:
Builds are now failing on recently created containers using the recommended Debian 8.11 environment for OCaml recipes producing dynamic libraries.

- xen-ocaml-libs should not rely on a handler defining CC, since OE   passes more than just the compiler path quote the $(CC) Makefile variable. As well, ocamlmklib should pass the flags down to the linker.
```
| x86_64-oe-linux-ar rcs libxenmmap_stubs.a  xenmmap_stubs.o && ocamlmklib -o `basename libxenmmap_stubs.a .a | sed -e 's/^lib//'`  xenmmap_stubs.o
| [...]/xen-ocaml-libs/4.12.2-pre-r0/recipe-sysroot-native/usr/bin/x86_64-oe-linux/../../libexec/x86_64-oe-linux/gcc/x86_64-oe-linux/6.4.0/ld: cannot find crti.o: No such file or directory
```
- xenclient-toolstack: ocamlmklib should pass the flags down to the linker.
```
| x86_64-oe-linux-ar rcs libstdext_stubs.a  unixext_stubs.o && ocamlmklib -o `basename libstdext_stubs.a .a | sed -e 's/^lib//'`  unixext_stubs.o
| [...]/xenclient-toolstack/0+gitAUTOINC+a5ca565cb6-r0/recipe-sysroot-native/usr/bin/x86_64-oe-linux/../../libexec/x86_64-oe-linux/gcc/x86_64-oe-linux/6.4.0/ld: cannot find crti.o: No such file or directory
```
This might be why master had to be switched to static libraries and could be forward ported, so it might be worth considering revisiting OXT-1704?

Requires: https://github.com/OpenXT/meta-openxt-ocaml-platform/pull/19

### DIFF
--- a/recipes-extended/xen/files/ocaml-makefiles-sysroot.patch
+++ b/recipes-extended/xen/files/ocaml-makefiles-sysroot.patch
@@ -1,0 +1,13 @@
+Index: git/tools/ocaml/common.make
+===================================================================
+--- git.orig/tools/ocaml/common.make
++++ git/tools/ocaml/common.make
+@@ -12,7 +12,7 @@ OCAMLFIND ?= ocamlfind
+ CFLAGS += -fPIC -Werror -I$(shell ocamlc -where)
+ 
+ OCAMLOPTFLAG_G := $(shell $(OCAMLOPT) -h 2>&1 | sed -n 's/^  *\(-g\) .*/\1/p')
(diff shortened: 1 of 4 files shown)


jandryuk:
This wants to be closed, right?

eric-ch:
@jandryuk absolutely, I thought I had closed it already.

PR #1353, opened 2020-08-13
dbd: make dbd RRECOMMENDS dbd-tools

jandryuk:
commit 3b15d672292d "dbd: Unify and rewrite recipe" changed the
packaging of db-* tools.  Previously, dbd included those tools, but now
they aren't included in dom0.  Make dbd RRECOMMENDS dbd-tools so they
are once again installed.


eric-ch:
Build [$10145](http://openxt-builder.ainfosec.com:8010/#/builders/3/builds/10145)

PR #1354, opened 2020-08-21
layer: Point recipes to upstream repositories

eric-ch:
Remove SRC_URI legacy customization.
Recipes should point to their upstream repository and it is up to site setup to provide or not local clones.

This PR moves away from using local clones through a (single) variable passed to each recipe and more than likely requires changes to auto-builders.
People interested in setting up a development tree can build from only cloning  the layers.
CI setup might want to find an alternative to build from local clones to test ongoing development, a possibility is to use [`externalsrc.bbclass`](https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#ref-classes-externalsrc), e.g: https://github.com/eric-ch/meta-openxt-externalsrc.

PR #1356, opened 2020-09-11
OXT-1694: ZEUS: qemu-dm: updates following switch to dedicated Linux argo header

eric-ch:
This change was pushed to master and affected https://github.com/OpenXT/linux-xen-argo.
The recipes have not been switched to track sub-project revisions yet, so `zeus` builds will break.

Given the scope of this change and since OpenXT does not maintain a versioned release in its `zeus` branch at the moment I recommend this change to be cherry-picked in `zeus`.

> xen_argo_ring_t is no longer automatically exposed to userspace via a libargo include, so don't base the ring size calculation on it; use the ARGO_RING_OVERHEAD macro for the calculation.
> char-argo.c now needs its own include of xen.h rather than relying
on transitive inclusion.

### DIFF
--- a/recipes-extended/qemu-dm/qemu-dm/atapi-pass-through.patch
+++ b/recipes-extended/qemu-dm/qemu-dm/atapi-pass-through.patch
@@ -1073,7 +1073,7 @@ Ross Philipson, <philipsonr@ainfosec.com>, 10/13/2016
 +
 +#define ATAPI_CDROM_PORT 5000
 +#define ARGO_ATAPI_PT_RING_SIZE \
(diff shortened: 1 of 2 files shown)


jandryuk (approved):
LGTM

jandryuk:
If xenclient-oe specified SRCREV for all the Xen recipes, then we would have avoided this issue.  It would also make it easy to rebuild at a given point in OpenXT history.

eric-ch:
@jandryuk I agree, it is the intended next step after switching to https://github.com/OpenXT/xenclient-oe/pull/1354 in the versioning changes. In the original proposal I suggested tagging all sub-repositories to their current documented version or `1.0` otherwise.

dpsmith (approved):

jandryuk:
Will merge Monday.

PR #1357, opened 2020-09-14
vbetool: remove the utility

eric-ch:
vbetool is no longer referenced or used anywhere.
The recipe also relies on a mirrored tarball instead of fetching the
sources from the repository. That mirror no longer exists and fetching
the tarball depends on MIRRORS to be set to a proper backup.

Remove the recipe and RDEPENDS from dom0 image.


jandryuk (approved):
LGTM

jandryuk:
Will merge soon.

PR #1360, opened 2020-09-25
kernel-module-signing: Add signing key to SSTATE_HASH

jandryuk:
The contents of the signing key are important to knowing if sstate can
be re-used.  There have been cases where the modules and kernel keys &
signatures have gotten out of sync.  Ensure the key is added to the
sstate to make modules depend on it.

I'm put some scripts that I've used in the past to check module signatures here.  They are a little rough, but they do the job:
https://gist.github.com/jandryuk/62f18d802eb705f5ecf4ecd866318f3a

### DIFF
--- a/classes/kernel-module-signing.bbclass
+++ b/classes/kernel-module-signing.bbclass
@@ -17,3 +17,9 @@ do_configure_append() {
                ${B}/.config
     fi
 }
+
+def get_signing_key(d):
+    path = d.getVar("KERNEL_MODULE_SIG_CERT") or os.path.join(d.getVar("STAGING_KERNEL_BUILDDIR"),"certs","signing_key.x509")
+    return path + ":" + str(os.path.exists(path))
+
+do_shared_workdir[file-checksums] = "${@get_signing_key(d)}"


eric-ch (approved):
This probably solves the reported occurrences of OXT-1523 still happening.
LGTM.

eric-ch:
[Build $mcql-10](https://openxt.ainfosec.com/buildbot/#/builders/14/builds/10)

jandryuk:
OXT-1523 talks about no signatures, but I've seen builds with the wrong signature.  I'm not sure how it happens, but signature with the wrong cert have been seen.  Having the sstate track the key (via its contents) seems like a step in the right direction.

eric-ch:
@jandryuk Indeed, my bad I jumped on the ticket too quickly.

eric-ch:
Will merge soon.

dpsmith (approved):

eric-ch:
@jandryuk This should go in the `zeus` branch as well, do you mind opening a PR against it?

jandryuk:
Done https://github.com/OpenXT/xenclient-oe/pull/1361

PR #1366, opened 2020-10-22
ZEUS: layer/zeus: fix sub-project versions

eric-ch:
Fix the revisions of sub-project to their stable version for this layer stable branch.

In order to simplify version management, since multiple recipes can pull a common repository, move the source/revision configuration to a matching include file and sort the recipes per repository provider.

The current revisions are the tips of the master branches for each subprojects.

Should the need for numbered version management per subproject arise, it is now a matter of changing the relevant .inc file and submitting a PR against this branch.


dpsmith:
Looks good to me, probably good to let the public builder run through it to make sure it will result in install-able build.

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/3/builds/20): [$zeus-stable-20](https://openxt.ainfosec.com/OpenXT/auto/zeus/zeus-stable/20/)

jandryuk (approved):

jandryuk:
I'm okay with merging this.  Do we want to coordinate with master so that the files paths are kept in sync?  Or just put this one in now and fix up master later?

eric-ch:
@jandryuk I think we should forward-port the recipe path changes into master. It might be worth doing after the Dunfell upgrade gets in to minimize conflicts.

jandryuk:
Ok.  Will merge soon.

PR #1369, opened 2020-11-06
blktap2 blk-mq fixes

jandryuk:
These two patches address hangs seen with blktap tapdevs.

The first came about because hangs were observed where the tapdisk was sleeping in select(), and mkfs.ext3 was blocked waiting for responses.   /sys/class/blktap2/blktapN/debug showed no pending requests.  Code inspection showed we may have stopped the hardware queues but never restarted them

Then second came about from code inspection.  RING_PUSH_REQUESTS wasn't called before we signaled a wake up via poll.

With both patches IO operations from mkfs.ext3 make better progress and seemingly fairer progress when multiple commands are run in parallel.

### DIFF
--- a/recipes-kernel/linux/5.4/patches/blktap2.patch
+++ b/recipes-kernel/linux/5.4/patches/blktap2.patch
@@ -245,7 +245,7 @@ PATCHES
 +int blktap_device_try_destroy(struct blktap *tap);
 +int blktap_device_pause(struct blktap *);
 +int blktap_device_resume(struct blktap *tap);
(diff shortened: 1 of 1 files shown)


jandryuk:
This is insufficient.  I have mkfs.ext3 waiting, tapdisk in select(), and cat /sys/class/blktap2/blktap14/debug shows 1 pending request.

jandryuk:
https://github.com/OpenXT/xenclient-oe/pull/1369/commits/2acfe289e34de545969619247a361703774a0f28 is broken.  It causes poll() to never return work.

PR #1376, opened 2020-12-11
Restore allarch after addition of MULTILIB

jandryuk:
commit b840fe8eff90 added multilib to the openxt-main distro, but an
unnoticed side effect was that allarch no longer worked.
allarch.bbclass doesn't set PACKAGE_ARCH to "all" when MULTILIB_VARIANTS
is set, so the rest of allarch.bbclass doesn't take effect.

Work around this by manually setting PACKAGE_ARCH="all" and then
inheriting allarch.  There could be a problem if any of these packages
depended on further arch specific packages, but none of ours do.

Wrap it into the small multilib-allarch.bbclass to centralize the
change.

### DIFF
--- a/classes/multilib-allarch.bbclass
+++ b/classes/multilib-allarch.bbclass
@@ -1,0 +1,6 @@
+# Since we have multilib enabled for xen, regular allarch doesn't take affect
+# and the packages use the default tune.  Setting PACKAGE_ARCH forces allarch
+# like we want.
+PACKAGE_ARCH = "all"
+
+inherit allarch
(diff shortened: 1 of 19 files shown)


eric-ch:
Upstream patch serie for reference: https://patchwork.openembedded.org/patch/154659/

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/10/builds/14): [mcq-14](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom-quick/14/)

eric-ch (approved):

eric-ch:
Will merge soon.

PR #1383, opened 2020-12-18
RFC: linux-yocto: provide OpenXT Linux kernels using kernel-yocto and config fragments

eric-ch:
This serie is an attempt to leverage `kernel-yocto.bbclass` to provide useful kernel configurations for Xen/OpenXT like service VMs, in a way that would be easier to extend than maintaining an entire defconfig.

`openxt-kmeta` provides:
- fragments for Xen features;
- ktypes for common Xen domain types;
- ktypes (extensions) for common purpose-specific Xen domains (dom0, stubdomain, ndvm, uivm);
- OpenXT patch-series separated in (intended to be) independent config fragments;
- BSP kernel configurations for OpenXT kernels (x86-64 with common desktop/laptops features);
- a BSP kernel configuration for an OpenXT installer with matching hardware support (dom0+ndvm combined);

To make experimentation easier and facilitate comparing the differences between kernel configurations using only one build tree, this change set defines a new set of MACHINE/images to build an OpenXT, the equivalences should be fairly straight forward (`s/xenclient-/openxt-`, `openxt-installer -> openxt-live-installer `).

https://github.com/OpenXT/xenclient-oe/commit/a699b34c51eb48500faa3324b34bd628dd2d76d6 is still required since MACHINE names are referenced directly in various recipes.


recipes-kernel/linux/files/openxt-kmeta/bsp/common-pc/common-pc-drivers-reduced.cfg
```diff
@@ -0,0 +1,114 @@
...
+
+#
+# Console display support
+#
+CONFIG_VGA_CONSOLE=y
+
+#
+# I2C support
+#
+CONFIG_I2C=y
```
jandryuk, line 95: I've only skimmed.  This is better than the full defconfigs we currently use, but I am a little surprised by how how many lines are required for all the fragments.  Picking I2C as an example, this seems like something we shouldn't need to enable explicit.  Looks like, CONFIG_I2C_I801=y would select I2C_SMBUS... I think that would indirectly turn I2C on?

I'm guessing you started from a full defconfig?  I recently found out about `make savedefconfig`.  That saves only the non-default values, so the full .config is created after running through `make oldconfig`.  For NDVM, `make savedefconfig` creates a 200-something line defconfig.  Making fragments out of that should be much smaller.

eric-ch: Actually I used `bitbake -c diffconfig` quite a bit to try and get as close as possible to the original defconfigs in order to not introduce too much significant changes.

That specific line comes directly from [the original file in yocto-kernel-cache](https://git.yoctoproject.org/cgit/cgit.cgi/yocto-kernel-cache/tree/bsp/common-pc/common-pc-drivers.cfg).

I agree though, that would avoid keeping some configuration enabled should another configuration fragment disable one that would have selected it.

eric-ch:
#### v2:

- Remove selected/default configuration in fragments
- No functional change

Note: There is a segfault triggered in QEMU in a stubdomain for HVM guests with graphical output (NDVM is fine). It seems this comes from the surfman integration.

recipes-kernel/linux/files/openxt-kmeta/patches/openxt-bridge-quirks/openxt-bridge-quirks.scc
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: MIT
+define KFEATURE_DESCRIPTION "Apply OpenXT bridge quirks."
+define KFEATURE_COMPATIBILITY all
+
+# Change the bridge carrier state according to the priority 0 forwarding port.
+patch bridge-carrier-follow-prio0.patch
+
+# Cmdline option to have the bridge pass copy of the skb and release the initial one.
+## TODO: determine if this is necessary (not enabled by default in any vm)
```
eric-ch, line 9: This is not set anywhere, it should be deprecated.

eric-ch:
##### v3:

- Rebase on vglass and recent OpenXT changes ([changes pre-squash](https://github.com/eric-ch/xenclient-oe/commits/linux-yocto-4)

I initially did the renaming of the machines/images (the kernels have to be named differently) to be able to build this side-by-side in a tree. It should be easy to switch entirely or revert just the renaming. What would be the preferred path here?

recipes-openxt/xenclient-dom0-tweaks/xenclient-dom0-tweaks/grub.cfg
```diff
@@ -31,7 +31,7 @@ fi
 
 TBOOT_COMMON_CMD="min_ram=0x2000000 loglvl=all serial=115200,8n1,0x3f8 logging=serial,memory"
 XEN_COMMON_CMD="dom0_mem=min:768M,max:768M,768M mbi-video vga=current flask=enforcing loglvl=debug guest_loglvl=debug ucode=-2 smt=0 bootscrub=1 argo=yes,mac-permissive=1"
-LINUX_COMMON_CMD="console=hvc0 root=/dev/mapper/xenclient-root ro boot=/dev/mapper/xenclient-boot swiotlb=16384 xen_pciback.passthrough=1 consoleblank=0 video.delay_init=1 vt.global_cursor_default=0 rootfstype=ext3 bootfstype=ext3"
+LINUX_COMMON_CMD="console=hvc0 root=/dev/mapper/xenclient-root ro boot=/dev/mapper/xenclient-boot swiotlb=16384 xen_pciback.passthrough=1 consoleblank=0 video.delay_init=1 vt.global_cursor_default=0 rootfstype=ext3 bootfstype=ext3 i915.enable_psr=0"
```
jandryuk, line 34: vglass commit f3286fe0382c55a4f4f50ce972f98ef1c6e103af "Mark framebuffer dirty to update contents" makes this unneeded now?

eric-ch: Yes indeed.

eric-ch:
#### v4:
- Legacy boot require the installer livecd to be Xen capable, add a new ktype for this.
- Add nouveau and radeon fragments since vGlass supports the GPUs in dom0
- Remove the txt_info module from the layer and instead maintain it as a patch

eric-ch:
#### v4:

- Add the missing BSP to support the legacy livecd installer.

eric-ch:
#### v5:

- Add missing crypto modules
- Fix config fragment rename that was not applied.

PR #1385, opened 2021-01-06
argo-module-headers: Stub compile/configure for externalsrc

jandryuk:
Using meta-openxt-externalsrc, xen-tools was being rebuilt during each
build.  If you ran bitbake-diffsigs it would complain that here was no
hash for argo-module-header:do_compile and do_configure.  While the
tasks were marked noexec, bitbake wouldn't write an sstate entry and it
would think it needed to re-run the whole dependency chain.

Execing placeholder stubs for do_compile & do_configure let bitbake
generate and find sstate to avoid the issue.

### DIFF
--- a/recipes-openxt/argo/argo-module-headers_git.bb
+++ b/recipes-openxt/argo/argo-module-headers_git.bb
@@ -14,6 +14,12 @@ do_install() {
     oe_runmake INSTALL_HDR_PATH=${D}${prefix} headers_install
 }
 
-# Skip build steps.
-do_compile[noexec] = "1"
-do_configure[noexec] = "1"
+# Skip build steps.  Exec empty functions so externalsrc can find sstate.
+#do_compile[noexec] = "1"
+do_compile() {
+    :
+}
+#do_configure[noexec] = "1"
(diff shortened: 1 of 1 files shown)


jandryuk:
There are still issues with externalsrc.  I'm seeing `Checksum for file oe-devtool-tree-sha1-argo-module-headers changed from a6d778f643ef5bbc9deeb2545088b305 to b1f163083725eaf70dcd701acf7fe0ab` when MACHINE changes.

eric-ch:
I am not seeing this anymore. It might have been due to the `git add -A` that externalsrc does to take a revision from the current source state, which should be fixed with https://github.com/OpenXT/linux-xen-argo/pull/12.

jandryuk:
OpenXT/linux-xen-argo#12 should be the fix for "Checksum for file oe-devtool-tree-sha1-argo-module-headers changed from a6d778f643ef5bbc9deeb2545088b305 to b1f163083725eaf70dcd701acf7fe0ab when MACHINE changes." which was observed even with this PR.  I'll close this for now.

PR #1386, opened 2021-01-06
OXT-1753: Run udev in initramfs to fix forward seal

jandryuk:
Forward seal is broken: https://openxt.atlassian.net/browse/OXT-1753

It seems to be that the blocknodes created by lvm without udev in the initramfs cause problems during an upgrade.  The lvm operations fail to change the all block nodes, and forward seal hashes root.old instead of the new dom0 rootfs.

Running udev in the initramfs, and using the default lvm script, creates symlinks instead of block nodes.  These are updated properly, and OTA forward seal works again.

lvm2 initramfs conf becomes unnecessary and can then be removed.

/dev/mapper/config is still a block node.  I haven't fully investigated, but I think this may be because cryptsetup runs from /sbin/init.root-ro.   /sbin/init.root-ro runs after the initramfs udev stops, but before the rootfs udev starts.  I haven't measured, but there could be a delay from cryptsetup failing to find udev before falling back to manually creating a device node.

Another thing I haven't investigated - running udev in the initramfs, more devices are probably automatically enabled.  For instance, usb drivers may be automatically loaded instead of their special, delayed handling in /sbin/init.root-ro.

So maybe this shouldn't be the end solution, but it at least restores forward-seal.


eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/10/builds/25): [mcq-25](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom-quick/25/)

eric-ch:
Looks good to me.

I think the efforts required to have LVM handled correct without udev will not be worth just having udev with a stringent configuration in the initramfs (which seems to be a [common](https://wiki.archlinux.org/index.php/Minimal_initramfs#Udev_requirement) opinion).

jandryuk:
> I haven't measured, but there could be a delay from cryptsetup failing to find udev before falling back to manually creating a device node.

@eric-ch Do you recall how long the timeout may have been when you made 8950739df69b43c0b9731aab27ea51e4b74c1b45 ?

On my todo list is seeing if an initramfs with only the device mapper rules will boot...

eric-ch:
@jandryuk I think it was 90s.

jandryuk:
Thanks, @eric-ch .  We boot slow enough already...  It's hard to know what's going on with our black screen mystery boot, but I'm not seeing anything that long.

jandryuk:
I did some tests with the full udev rule set versus a minimal udev rules of just the device mapper (dm) ones.  There isn't much differences except the full set gets a couple of extra symlinks:
/dev/disk/by-uuid for the fat ESP
/dev/disk/by-id for the disk & partitions by name/serial numer
/dev/disk/by-id for the disk & partitions by wwn numer
/dev/disk/by-partuuid for the 
/dev/input/by-path/platform-i8042-serio-0-event-kbd
/dev/rtc

Since OE packages the udev rules in the udev package, I'd just leave them instead of playing games with the packaging/image.

The kernel modules installed in the initramfs is rather limited, so not many things that can be autoloaded.  Looking at lsmod, only the tpm drivers are loaded, and those are explicitly modprobed in the tpm initramfs-framework files.  We have USB modules, but those are not loaded... that seems to be because we install mainly the *hci-hcd modules, but only the ehci-pci module.  Only an ehci PCI device would cause autoloading, and my test system doesn't have that.  The USB modules are loaded in /sbin/init.root-ro off the dom0 rootfs, so they it seems they can be dropped from initramfs.

eric-ch (approved):
LGTM.

jandryuk:
Pushed an additional commit to remove the USB kernel modules.  This is "defense in depth" to prevent udev from doing more than we want and expect during early boot.

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/10/builds/35): [$mcq-35](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom-quick/35/)

jandryuk:
Force pushed with re-worded commits, but no code change:
[Run udev in initramfs to fix forward seal](https://github.com/OpenXT/xenclient-oe/pull/1386/commits/4ce3e297b4f63ec4806b8cd07dc4c9f6aa6f11de) now explains that it is to fix forward seal.
[Remove lvm2-conf-initramfs](https://github.com/OpenXT/xenclient-oe/pull/1386/commits/5cff1ddfd0a6c6b26975743c43efeb18545091ee) s/installer/initramfs/ since it is removing the initramfs one.  Also mentioned that it removes the mention of installer PREFERRED_PROVIDER.

That second commit made me realize that there is more lvm2-conf stuff that can be cleaned up.  I think that can be done as a follow up.

eric-ch (approved):
LGTM.
Tested with https://github.com/OpenXT/xenclient-oe/pull/1390.

eric-ch:
Will merge soon.

PR #1387, opened 2021-01-08
installer/answer: package custom answerfiles

eric-ch:
package the custom answerfiles provided by the layer.

When using `externalsrc` and `rmwork`, the tmpdir where the answer files are placed by `do_unpack` and used in `do_deploy`, will be wiped. This can lead to issues when the recipe used is targeting an image that needs `do_deploy` to re-run under `externalsrc`, `do_unpack` is not always re-run.

Placing the files in the package tracked by the sstate seems to resolve this issue.

### DIFF
--- a/recipes-openxt/xenclient-installer/xenclient-installer_git.bb
+++ b/recipes-openxt/xenclient-installer/xenclient-installer_git.bb
@@ -21,10 +21,6 @@ inherit multilib-allarch deploy
 do_install () {
     ${S}/install part1 ${D}/install
     ${S}/install part2 ${D}
-}
-
-do_deploy() {
-    install -m 0755 -d "${DEPLOYDIR}/netboot"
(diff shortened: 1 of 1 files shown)


eric-ch:
Here is an example of this happening: https://openxt.ainfosec.com/buildbot/#/builders/18/builds/6. I suspect the issue comes from `do_deploy ` using installing unpackaged files from `WORKDIR`. It is unclear why `do_deploy` was run again in this case.

jandryuk (approved):

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/10/builds/28): [$mcq-28](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom-quick/28/)

jandryuk:
Will merge soon.

PR #1388, opened 2021-02-01
libxl: Wait for argo-firewall-ready asynchronously

troyc:
Wait for argo-firewall-ready asynchronously using xswait


eric-ch:
Could you move this logic out in its own patch?
`libxl-xenmgr-support.patch` conflates 3 changes mostly unrelated to each other:
- Disable QEMU for PV guest
- Have `shutdown_domain` to issue hard shutdown
- Wait for the Argo firewall rules to be written by `xenmgr`

recipes-extended/xen/files/libxl-xenmgr-support.patch
```diff
@@ -108,31 +111,75 @@ PATCHES
...
++    if (ret) goto out;
+ 
+     return;
+ 
+@@ -2366,6 +2380,30 @@ out:
+     spawn_stubdom_pvqemu_cb(egc, &sdss->pvqemu, ret);
+ }
+ 
++static void argo_xswait_cb(libxl__egc *egc, libxl__xswait_state *xswait,
++                           int rc, const char *p)
```
jandryuk, line 157: p here is the value of the watched xenstore node.  Do you want to verify it's the expected value?

jandryuk (commented):
This generally looks reasonable.  However, there is no explanation in the commit message for *why* this change is being made.  Why convert to async?  Also, the timeout increase from 4 to 20 seconds is kinda hidden inside.  Why was that changed - is 4 seconds too short?  Please explain the reason for all changes as precisely as possible.

troyc:
> Could you move this logic out in its own patch?
> `libxl-xenmgr-support.patch` conflates 3 changes mostly unrelated to each other:
> 
> * Disable QEMU for PV guest
> * Have `shutdown_domain` to issue hard shutdown
> * Wait for the Argo firewall rules to be written by `xenmgr`

Sure, I could split up the patch like this.

> This generally looks reasonable. However, there is no explanation in the commit message for _why_ this change is being made. Why convert to async? Also, the timeout increase from 4 to 20 seconds is kinda hidden inside. Why was that changed - is 4 seconds too short? Please explain the reason for all changes as precisely as possible.

The change is being made to preclude any issues caused by blocking and to better position the patches for upstreaming (depending on how the argo-viptables make their way upstream). There's no reason to change the timeout value, so it can remain at 4 seconds in my revision.

recipes-extended/xen/files/libxl-xenmgr-support.patch
troyc: Yes, I don't see any reason why not to verify p's value. However, I don't think there's anything better to do than to log an error message and continue on our merry way, same as in the case of a timeout or other error at this point.

troyc:
I've applied the suggested changes to the branch. What's left in libxl-xenmgr-support.patch will become obsolete with vglass, with the possible exception of a few vif-related changes.

troyc:
I updated the branch to add a commit message.

jandryuk (approved):
Thanks.

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/10/builds/36): [$mcq-36](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom-quick/36/)

eric-ch (approved):

eric-ch:
Will merge soon.

PR #1393, opened 2021-03-25
Fix build SRC_URI's and remove gobi-loader.

troyc:
If gobi-loader is actually still used, we could find an alternative source.


recipes-extended/linux-input/linux-input_1.6.0.bb
```diff
@@ -1,4 +1,4 @@
-SRC_URI = "https://svwh.dl.sourceforge.net/project/linuxconsole/linuxconsoletools-${PV}.tar.bz2 \
+SRC_URI = "https://newcontinuum.dl.sourceforge.net/project/linuxconsole/linuxconsoletools-${PV}.tar.bz2 \
```
jandryuk, line 1: I think you should use "https://download.sourceforge.net/project/linuxconsole/linuxconsoletools-${PV}.tar.bz2" so you go through the load balancer/redirector instead of hardcoding a particular server.

jandryuk (commented):
Thanks. I think dropping gobi_loader is good.  Please split the SRC_URI changes into separate commits.  linux-input is independent from rhboot renaming branches.

troyc:
I split the SRC_URI changes into separate commits and changed the linux-input URI as suggested by @jandryuk.

jandryuk (approved):
LGTM.  Thanks.

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/22/builds/8): [$master-custom-8](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom/8/)

eric-ch (approved):

eric-ch:
Will merge soon.

PR #1395, opened 2021-04-02
Add image-check-module-sigs.bbclass

jandryuk:
image-check-module-sigs.bbclass checks the signatures of modules
installed in an image to ensure they are valid for the associated
kernel.  Usually this is /boot/bzImage, but it will check in deploy dir
for stubdomain's external kernel.

This adds ~7 seconds to dom0 image creation, but it's nicer to check during
the build than have an install that fails to boot properly.

Tie it in to openxt-image so the check is always performed.

The installer doesn't have signed modules, so there is a workaround (if [[ $output =~ "Can't find system certificate list " ]]; then) to skip the check in that case.

pipefail is enabled so we get the exit value of `openssl smime -verify` on failure and not the (always successful) exit status of `sed -n '/Verification successful/!p'`.  I did that so the log will have any error message from smime without spamming the log with Verification successful.  The caller will print the failing module name, so silencing all output from smime is another option.  The error output from smime isn't terribly informative when a signature fails, but there could be value there.

### DIFF
--- a/classes/image-qa-module-sigs.bbclass
+++ b/classes/image-qa-module-sigs.bbclass
@@ -1,0 +1,106 @@
+do_image_qa_module_sigs() {
+	local bzImage="${IMAGE_ROOTFS}/boot/bzImage"
+	if [ ! -f "$bzImage" ]; then
+		bzImage="${DEPLOY_DIR_IMAGE}/bzImage"
+	fi
+
+	if [ ! -f "$bzImage" ]; then
+		bberror "Cannot find bzImage in rootfs or DEPLOYDIR"
+		exit 1
+	fi
+
+	local vers
+	# -L to follow the symlink for deploy dir
+	vers="$( file -L "$bzImage" | sed 's/.*version \([^ ]*\) .*/\1/' )"
+
+	if [ ! -d "${IMAGE_ROOTFS}/lib/modules/$vers" ]; then
+		bbnote "bzImage version is $vers, but module directory is missing"
+		exit 0
+	fi
+
+	local builddir="${STAGING_KERNEL_BUILDDIR}"
+	local scriptdir="${STAGING_KERNEL_DIR}/scripts"
+
+	local vmlinux
+	local keyring
+	vmlinux="$( mktemp -t "vmlinux-$vers.XXXXXXXX" )"
+	keyring="$( mktemp -t "keyring-$vers.XXXXXXXX" )"
+
+	# Extract vmlinux from bzImage
+	"$scriptdir/extract-vmlinux" "$bzImage" > "$vmlinux" 2>/dev/null
+
+	# Extract keyring from vmlinux (needs System.map since bzImage doesn't
+	# have symbols
+	if ! output=$( "$scriptdir/extract-sys-certs.pl" \
+			-s "$builddir/System.map-$vers" "$vmlinux" \
+			"$keyring" 2>&1 1>/dev/null ) ; then
+		if expr "$output" : "^Can't find system certificate list " >/dev/null
+		then
+			bbnote "No certificate list - skipping"
+			exit 0
+		else
+			bberror "$output"
+			exit 1
+		fi
+	fi
+
+	# We need a PEM for smime -certfile option
+	openssl x509 -in "$keyring" -inform DER -out "$keyring.pem" -outform PEM
+	mv "$keyring.pem" "$keyring"
+
+	ret=0
+	# For loop so setting ret is in the main shell, and bitbake doesn't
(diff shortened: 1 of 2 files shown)


classes/image-check-module-sigs.bbclass
```diff
@@ -0,0 +1,89 @@
...
+
+	# Extract vmlinux from bzImage
+	$scriptdir/extract-vmlinux $bzImage > $vmlinux 2>/dev/null
+
+	# Extract keyring from vmlinux (needs System.map since bzImage doesn't
+	# have symbols
+	if ! output=$( $scriptdir/extract-sys-certs.pl \
+			-s $builddir/System.map-$vers $vmlinux \
+			$keyring 2>&1 1>/dev/null ) ; then
+		if [[ $output =~ "Can't find system certificate list " ]]; then
```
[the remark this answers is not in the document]
jandryuk: Yes, it's a bash-ism.  I can change to `expr "$output" : "^Can't find system certificate list "`

classes/image-check-module-sigs.bbclass
```diff
@@ -0,0 +1,89 @@
+do_image_check_module_sigs() {
+	bzImage=${IMAGE_ROOTFS}/boot/bzImage
```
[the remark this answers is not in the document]
jandryuk: Will do.

jandryuk:
I noticed this morning that image.bbclass includes task do_image_qa which executes the commands in IMAGE_QA_COMMANDS.  Should I change the explicit do_image_check_module_sigs task to just add to IMAGE_QA_COMMANDS?  Maybe s/do_image_check_module_sigs/do_image_qa_module_sigs/ as well?

eric-ch:
@jandryuk `IMAGE_QA_COMMANDS` looks a lot more streamlined, I did not know about it, +1 on the renaming of the function as well.

Looking at the history, it could be leveraged more to have builds fail earlier (e.g, checking that `tmpfs` directory mount points are not populated by packages).

jandryuk:
Updated with:

- Quoted variables
- Rename to image_qa_module_sigs
- Replace `[[` bash-ism with `expr`
- Use do_image_qa/IMAGE_QA_COMMANDS
- Added do_image_qa[depends] to ensure we have perl for the linux extract scripts and openssl which is called directly.

jandryuk:
Also, I pass keyring as a parameter to check_module instead of relying on it being set globally.

classes/image-qa-module-sigs.bbclass
```diff
@@ -0,0 +1,95 @@
...
+			sed -n '/Verification successful/!p'
+	ret=$?
+
+	rm "$data"
+	rm "$sigdata"
+
+	return "$ret"
+}
+
+do_image_qa[depends] += "perl-native:do_populate_sysroot openssl-native:do_populate_sysroot"
```
eric-ch, line 91: `$scriptdir` will be provided through the `$tmpdir/work-shared` so this should add `virtual/kernel:do_shared_workdir` to `depends`.

classes/image-qa-module-sigs.bbclass
```diff
@@ -0,0 +1,95 @@
+do_image_qa_module_sigs() {
+	bzImage="${IMAGE_ROOTFS}/boot/bzImage"
```
eric-ch, line 2: I missed it during the first review, variables could be made `local`, yet it does not look like this is idiomatic in OE recipes.

classes/image-qa-module-sigs.bbclass
jandryuk: Yes, it should - both for the scripts and System.map-$vers.  I am concerned that if SSTATE is used, bzImage comes from the linux-openxt package, but System.map will be from a newly compiled copy.  They *should* be the same, but it makes me uncomfortable.  System.map is packaged in kernel-dev with:
boot/Module.symvers-5.10.25
boot/System.map-5.10.25
boot/config-5.10.25

Maybe instead of installing the package, we could do_deploy System.map?

jandryuk:
Updated `do_image_qa[depends]` and made variables local.

(Had a bad force push in there on top of vglass and usbvm).

I'm still unsure what to do about System.map.

eric-ch (approved):
I have been using this a fair amount while testing https://github.com/OpenXT/xenclient-oe/pull/1401.
LGTM.
One potential caveat: (https://github.com/OpenXT/xenclient-oe/pull/1401#discussion_r622414056), although I do not want to block this on such a corner case unless someone has a current use for it.

eric-ch:
Will merge soon.

PR #1396, opened 2021-04-02
[networkmanager] Always use existing connection

crogers1:
profiles. Networkmanager core logic always generates default
  connection profiles for managed devices. These fail to match
  any on-disk (db) saved connection profiles we may have from
  a previous boot and therefore create duplicate entries,
  ignoring any of the modified settings.

  This patch will make NM prefer to use any existing connection
  profiles that share an ID with the default connections NM
  tries to make every boot. This fixes a regression where we
  are unable to save modifications to network settings across
  reboots.

  OXT-1816

### DIFF
--- a/recipes-connectivity/networkmanager/networkmanager-1.18.4/always-use-existing-connections.patch
+++ b/recipes-connectivity/networkmanager/networkmanager-1.18.4/always-use-existing-connections.patch
@@ -1,0 +1,98 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Always used saved connection profiles. If none exist, let networkmanager create
+default ones.
+
+################################################################################
+LONG DESCRIPTION:
+################################################################################
+Now that networkmanager actively manages wired/wireless/bridged devices+
+connections, if we make changes to these connections and save them to disk (db)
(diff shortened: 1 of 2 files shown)


crogers1:
Open to suggestions. Perhaps this isn't the best approach but I avoided modifying core Network Manager logic and managed to keep these devices managed by Network Manager, as opposed to the old way we used to do things.

eric-ch:
@jandryuk The repro I had with this was setting a DNS on `brbridged` and rebooting NDVM. NM would then save the `brbridged` configuration to dbd, but upon reboot re-create a new default `brbridged` configuration (without the DNS configuration) in `NM_KEYFILE_PATH_NAME_RUN` and use it instead of the one read from dbd (`nm_sync.sh`) and saved in `/etc/NetworkManager/system-connection`. Now there are 2 configuration named `brbridged` reported by NM. Setting the DNS again in nm-applet on the default brbridged and rebooting again would then have 3 configurations named `brbridged`.

crogers1:
Been testing with this patch all day, it was doing some things that weren't great.  I found a way to simplify it down to a one line change, going to close this PR and open another.

PR #1397, opened 2021-04-07
[networkmanager] Always use existing connection

crogers1:
profiles. Networkmanager core logic always generates default
  connection profiles for managed devices. These fail to match
  any on-disk (db) saved connection profiles we may have from
  a previous boot and therefore create duplicate entries,
  ignoring any of the modified settings.

  This patch will allow NM to try to match on existing connection
  profiles during its initialization. This fixes a regression where we
  are unable to save modifications to network settings across
  reboots.

  OXT-1816

### DIFF
--- a/recipes-connectivity/networkmanager/networkmanager-1.18.4/always-use-existing-connections.patch
+++ b/recipes-connectivity/networkmanager/networkmanager-1.18.4/always-use-existing-connections.patch
@@ -1,0 +1,59 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Always used saved connection profiles. If none exist, let networkmanager create
+default ones.
+
+################################################################################
+LONG DESCRIPTION:
+################################################################################
+Now that networkmanager actively manages wired/wireless/and brbridged devices+
+connections, if we make changes to these connections and save them to disk (db)
+for future use, due to networkmanager's core logic, it creates new connections
+with the same ID but refuses to attempt to try to match them. This causes
+multiple connections with the same IDs to be shown in the nm-applet.
+
+The keyfile plugin we use to save/parse connection profiles will read and 
+generate connections before NetworkManager runs through its core autogeneration
+logic. This patch removes a flag that is set always set to FALSE during this 
+initialization and lets NetworkManager attempt to match on an existing connection
+profile.
+
+This allows connection setting changes, such as DNS entries or password credentials,
+to persist across ndvm and host platform reboots.
+
+################################################################################
+CHANGELOG
+################################################################################
+Authors:
+Chris Rogers <rogersc@ainfosec.com>
+
+################################################################################
+REMOVAL
+################################################################################
+None
+
+################################################################################
+UPSTREAM PLAN
+################################################################################
+None
+
+################################################################################
+INTERNAL DEPENDENCIES
+################################################################################
+db-nm-settings.patch
+
+################################################################################
+PATCHES
+################################################################################
+--- a/src/nm-manager.c
++++ b/src/nm-manager.c
+@@ -2580,7 +2580,7 @@ get_existing_connection (NMManager *self
+ 		return NULL;
+ 	}
+ 
+-	if (!matched && assume_state_guess_assume) {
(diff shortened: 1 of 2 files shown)


crogers1:
@jandryuk to answer your question on the other PR, commit ```374cd7699d3bac96fa4d6f1d0838230c0c3c8137``` updated NetworkManager.conf to consider all ```eth*, wlan*, brbridged*``` managed.

jandryuk:
Thanks for the reference @crogers1 

> The keyfile plugin we use to save/parse connection profiles will read and 
> generate connections before NetworkManager runs through its core autogeneration
> logic. This patch removes a flag that is set always set to FALSE during this 
> initialization and lets NetworkManager attempt to match on an existing connection
> profile.

Sorry, for my clarification:
1. keyfile plugin creates configuration files from dom-store but they aren't loaded yet
2. NM runs core autogeneration and creates and uses an internal brbridged config
3. keyfile plugin loads configurations but skips brbridged since it is already loaded

This patch modifies step 2 to use the configuration file for brbridged, if found?

Also, I think you want " is set always set to FALSE" -> " is always set to FALSE"?

This patch is much simpler than the previous, so that's nice.

In the other PR I mentioned maybe replacing keyfile with an openxt plugin might be better.  Another approach could be to have network-slave configure NM over dbus.  But these would probably open up their own set of problems :)

jandryuk (approved):
I don't think I have enough knowledge to really review this.  For instance, I don't know if there are potential side effects.  However, since it's a one line change, I am approving.  But I'd like someone else to review and approve as well.

crogers1:
Yea I definitely think an openxt plugin or NM configuration over dbus is a much more ambitious effort.  But to clarify what's going on here:

1. network-slave creates the connection profiles by calling nm_sync.sh, which converts from db format to networkmanager format and stores them in ```/etc/NetworkManager/system-connections```.
2. NM starts up, keyfile plugin reads these and creates in-memory 'connections'.
3. Not sure how it gets here, but at some point during initialization, NM always creates default connections for managed devices. On my box, this is eth0 and brbridged.
4. NM performs complicated matching logic. To quote their inline comments:
```
 /* Now we need to compare the generated connection to each configured
 * connection. The comparison function is the heart of the connection
 * assumption implementation and it must compare the connections very
 * carefully to sort out various corner cases. Also, the comparison is
 * not entirely symmetric.
 *
 * When no configured connection matches the generated connection, we keep
 * the generated connection instead.
 */
```
 Because ```assume_state_guess_assume``` and ```assume_state_connection_uuid``` are FALSE and NULL respectively, we never end up trying to match anything, so NM keeps the generated connections.
 5. Now because keyfile loaded the connection profiles in step 2, we have duplicate connections for ```eth0``` and ```brbridged```.

I don't know how to get the NM state file A) persist, and B) track the connection UUIDs such that ```assume_state_connection_uuid``` has something non-null, but I do know we want these autogenerated connections to try to search through existing connections and find a matching candidate, so by removing ```assume_state_guess_assume``` from the block in the patch here, NM will at least try to find something existing already (ie. our profiles restored from the db).

As for the wireless not autoconnecting, that's a different issue I'm working on currently.

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/22/builds/18): [master-custom-18](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom/18/)

recipes-connectivity/networkmanager/networkmanager-1.18.4/always-use-existing-connections.patch
```diff
@@ -0,0 +1,59 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Always used saved connection profiles. If none exist, let networkmanager create
```
eric-ch, line 4: Nit: s/used/use

eric-ch (approved):
LGTM.

I am also of the opinion that keyfile should be replaced for another plugin that would connect over Argo to the DB, but that's a larger effort.

eric-ch:
Will merge soon.

PR #1401, opened 2021-04-13
module-signing: amend build errors and dependency tracking

eric-ch:
Module signing and sstate mirroring can trigger failures (see [OXT-1523](https://openxt.atlassian.net/browse/OXT-1523)) either at build-time (pseudo database corruption) or runtime (modules signed with a different key than the one embedded in the kernel binary). This changeset should rebuild the kernel and re-sign out-of-tree modules should the signing key change, whether or not the key is provided outside of the tree or the default generated one.

It should be recommended to use `KERNEL_MODULE_SIG_CERT`/`KERNEL_MODULE_SIG_KEY` locally. This will inevitably trigger kernel rebuilds initially without being able to use an sstate as the private key (albeit test ones), should not be shared like that. Kernel build times are not that bad though compared to webkitgtk and other UI components.


classes/module-signing.bbclass
```diff
@@ -17,6 +17,10 @@ export KERNEL_MODULE_SIG_KEY
 # build-time keypair will be generated and used for signing and embedding.
 export KERNEL_MODULE_SIG_CERT
 
+def get_signing_key(d):
+    path = d.getVar("KERNEL_MODULE_SIG_CERT") or os.path.join(d.getVar("STAGING_KERNEL_BUILDDIR"),"certs","signing_key.x509")
```
jandryuk, line 21: I think `STAGING_KERNEL_BUILDDIR` should be `B` here.  Looking at oe-core:
`meta/classes/base.bbclass:do_fetch[file-checksums] = "${@bb.fetch.get_checksum_file_list(d)}"`
The file-checksums are the inputs to the task.  

do_compile will generate the key, so do_configure is too early (today it might inadvertantly be depending on an old key still in STAGING_KERNEL_BUILDDIR?)
do_shared_workdir can depend on B since it is ordered after do_compile, and it make sense since it is copying from B to STAGING_KERNEL_BUILDDIR.

All out of tree modules depend on do_shared_workdir, so they should be covered.

For in tree, the existing task deps are do_install -> do_sizecheck -> do_strip -> do_compile_kernelmodules -> do_shared_workdir, so that should already be covered by do_shared_workdir, I think.

eric-ch: `do_configure` has to at least depend on `KERNEL_MODULE_SIG_CERT` and account if it is set or unset to update the defconfig.
Maybe then vardeps is more appropriate  for the defconfig and we can set `do_compile` to check `file-checksums` of either `KERNEL_MODULE_SIG_CERT` or `d.getVar("B"),"certs","signing_key.x509"`?

jandryuk: Ah, good point about `do_configure` depending on KERNEL_MODULE_SIG_CERT being set or unset.  I just used bitbake-dumpsigs and verified OE already creates a dependency on the variable's value.

Here are the module signing relevant tasks:
```
do_configure:
  [ -n KERNEL_MODULE_SIG_CERT ] && update .config
do_compile
  [ -z KERNEL_MODULE_SIG_CERT ] && generate key/cert
  embed cert/public key
do_shared_workdir
  (implicitly [ -z KERNEL_MODULE_SIG_CERT ] ) cp ${B} cert to STAGING_KERNEL_BUILDDIR
do_install 
  [ -z KERNEL_MODULE_SIG_CERT ] && sign modules ( B for in-tree & STAGING_KERNEL_BUILDDIR for out-of-tree )
do_sign_module
  [ -n KERNEL_MODULE_SIG_KEY ] && sign modules
```

do_configure depends on KERNEL_MODULE_SIG_CERT's setting
do_compile depends on the cert's contents.  This is straightforward for KERNEL_MODULE_SIG_CERT, but tricky for an auto-generated key since the task also creates it.

That implies something like:
```
if KERNEL_MODULE_SIG_CERT
    do_compile[file-checksums] += "${@get_signing_cert(d)}"
else
    do_shared_workdir[file-checksums] += "${@get_autogen_signing_cert(d)}"
```

`get_*signing_key` with an empty return value as appropriate.

We could have a new task between do_configure and do_compile to call `make certs/signing_key.pem signing_key.x509` which would unify on do_compile[file-checksums].

This addition:
`do_sign_modules[file-checksums] += "${@get_signing_key(d)}"`
Yes, we need a dependency there, but I think it should be on KERNEL_MODULE_SIG_KEY and not the cert?

eric-ch: Right, bitbake is able to pickup the vardep since it is explicit in the step. I'll make the changes and try that.

I think at this point the main issue is the sstate being used when we don't want it to.

jandryuk: > I think at this point the main issue is the sstate being used when we don't want it to.

From https://bugzilla.yoctoproject.org/show_bug.cgi?id=14344
"So we don't support changing task hashes part way through the build."

This attempt to hash the signing key/cert doesn't work like we want.

eric-ch: Yes.
I do not see any good way to both leverage the sstate and the auto-generated certificate reliably.
I think our options are:
- pass another certificate that was generated out of the tree (https://github.com/apertussolutions/bordel/pull/51, although we could re-use `repo-{dev,prod}` used to sign the XC-REPOSITORY stuff);
- use a test-signing certificate for development/debug purposes if none is specified out of tree (meta-security does that https://git.yoctoproject.org/cgit/cgit.cgi/meta-security/tree/meta-integrity/data/debug-keys);

classes/module-signing.bbclass
```diff
@@ -10,17 +10,35 @@ INHIBIT_PACKAGE_STRIP = "1"
 export HOST_EXTRACFLAGS = "${BUILD_CFLAGS} ${BUILD_LDFLAGS}"
 
 # Set KERNEL_MODULE_SIG_KEY in local.conf to the filepath of a private key
-# for signing kernel modules.  If unset, signing can be done offline.
+# for signing kernel modules.
```
jandryuk, line 13: I think it can still be valid to have KERNEL_MODULE_SIG_KEY unset for offline signing.

classes/module-signing.bbclass
```diff
@@ -44,3 +62,6 @@ addtask sign_modules after do_install before do_package
 do_install[lockfiles] = "${TMPDIR}/kernel-scripts.lock"
 # Explicit keys sign modules in do_sign_modules
 do_sign_modules[lockfiles] = "${TMPDIR}/kernel-scripts.lock"
+
+do_sign_modules[depends] += "virtual/kernel:do_shared_workdir"
```
jandryuk, line 66: do_sign_modules already transitively depends on "virtual/kernel:do_shared_workdir".  Are you adding the explicit depends for SIGN_FILE?

jandryuk (commented):
[ -z "${KERNEL_MODULE_SIG_CERT}" ] &&  [ -n "${KERNEL_MODULE_SIG_KEY}" ]  is invalid, so maybe bbfatal on that?

Embedding a CERT but doing offline signing is valid, but that would break do_image_qa signature checking.  Maybe that should be mode should be explicit instead of implicit from SIG_KEY bring unset.  i.e. introduce KERNEL_MODULE_OFFLINE_SIGNING?  That would then control do_sign_modules and do_image_qa?

classes/module-signing.bbclass
eric-ch: Yes, and you are correct. Also `get_signing_cert` in `file-checksums` is already satisfied from `do_configure` dependencies.

classes/module-signing.bbclass
eric-ch: I discarded that use-case to unblock the rest, I'll amend and add the `bbfatal`. `KERNEL_MODULE_OFFLINE_SIGNING` sounds fine, it will require changes in https://github.com/OpenXT/xenclient-oe/pull/1395 for `do_image_qa`.

eric-ch:
While there are still things up for discussion, it sounds like https://github.com/OpenXT/xenclient-oe/pull/1401/commits/39a7b5f5ac4b68c290c04be97739fcc6db8e86f6 is not controversial, so I'll split it out of this PR and submit it separately to unbreak some trees using key/cert.

Using certain types of certificates rather than others also requires changes to the kernel configuration (e.g, `default_md = sha256`, or `openssl req -sha256` requires `CONFIG_CRYPTO_SHA256=y`). So I think this should be refactored to guaranty that these conditions are met. This would currently require some `sed` invocation against the defconfig. Should https://github.com/OpenXT/xenclient-oe/pull/1383 be considered, it would only be a `KERNEL_FEATURES` with the relevant `.scc`.

eric-ch:
- Rebased on https://github.com/OpenXT/xenclient-oe/pull/1405
- Allow offline signing, add a `bbnote` if that is detected
- out-of-tree modules needed `STAGING_KERNEL_BUILDDIR` (at least with externalsrc) and not `B`
- re-work the tests to error accurately when given an invalid configuration using out-of-tree certificates

jandryuk (approved):
LGTM.  Thank you

eric-ch:
Rebase for history coherency. No change.

classes/kernel-module-signing.bbclass
```diff
@@ -7,19 +7,20 @@ SIGN_FILE = "${B}/scripts/sign-file"
...
-        echo "CONFIG_MODULE_SIG_KEY=\"${KERNEL_MODULE_SIG_CERT}\"" >> \
-               ${B}/.config
-        sed -i -e '/CONFIG_MODULE_SIG_ALL[ =]/d' ${B}/.config
-        echo "# CONFIG_MODULE_SIG_ALL is not set" >> \
-               ${B}/.config
+    if ! grep -q '^CONFIG_MODULE_SIG=y' ${B}/.config ; then
+        return
+    fi
+    if [ -z "${KERNEL_MODULE_SIG_CERT}" ]; then
+        bbfatal "Kernel module signing should only be used when setting \
```
jandryuk, line 14: Something to note here.  The bbfatal message is printed and then bitbake prints the contents of log.do_configure, which is quite long with all the `make oldconfig` output.  All you see at the end of it is: 

```
make[1]: Entering directory '.../build'
  GEN     Makefile
# 
# No change to .config
# 
make[1]: Leaving directory '.../build'
make: Leaving directory '/home/build/openxt/build-201218/tmp-glibc/work-shared/usbvm/kernel-source'
WARNING: .../temp/run.do_configure.7382:1 exit 1 from 'exit 1'

ERROR: Logfile of failure stored in: .../temp/log.do_configure.7382
```

If you look back far enough you see the message.  It's on line 2299 of the 4596 line log.do_configure.  Not a blocker, but something I noticed.

eric-ch: We could just `bberror` and `exit 1` I guess.

jandryuk: That didn't change the output :/

jandryuk:
Will merge soon.

PR #1402, opened 2021-04-13
[networkmanager] Clear user permissions

crogers1:
from connection profile before persisting it to disk. This allows
  networkmanager to automatically connect a wireless connection based
  on the SSID and credentials stored in a saved connection profile.
  The ndvm only has one user and is run headless so assigning user
  permissions to a connection profile is pointless, in addition to
  causing this regression.

  OXT-1817

### DIFF
--- a/recipes-connectivity/networkmanager/networkmanager-1.18.4/clear-permissions-on-connection-profiles.patch
+++ b/recipes-connectivity/networkmanager/networkmanager-1.18.4/clear-permissions-on-connection-profiles.patch
@@ -1,0 +1,72 @@
+################################################################################
+SHORT DESCRIPTION:
+################################################################################
+Remove any permission entries from a connection before persisting it to disk
+
+################################################################################
+LONG DESCRIPTION:
+################################################################################
+When a wireless connection has permissions associated with it, networkmanager
+will bail out of its autoconnection operation, even if autoconnect is TRUE and
+the wireless connection profile has access point credentials saved.
+
+The ndvm only has one user and is run headless, so there's no reason to define
+permissions for a given connection with the root user. When we persist the
+connection profile to disk in the keyfile plugin, prune out any permissions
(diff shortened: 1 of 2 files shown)


crogers1:
Sorry to add you to another networkmanager one @jandryuk.

Also either this one or https://github.com/OpenXT/xenclient-oe/pull/1397 will need to be rebased, whichever gets merged first.

jandryuk:
Nice investigation.  `When this is not empty, the connection can be active only when one of the specified users is logged into an active session.`  Why did NM write a permission when a user wasn't logged in to begin with :/

What do you think of adding this (between the two other match statements) to db_to_nm.awk:
```
/^  permissions = .+$/ {
    continue
}
```
It skips writing permissions lines when pulling out of dom-store.

crogers1:
As opposed to a NM patch? Not a bad idea if we wanted to avoid touching NM. Downside is the dom-store wouldn't match the config in system-connections/ and it would run every time we pull from the db. With my patch it would prune it once and then ```num_permissions``` for that connection would be 0 from then on, and it would just skip the block on subsequent boots.  

I'll go with whatever we think is best.

jandryuk (commented):
Patch code looks good.  Please add to the patch description this quote from the NM documentation "When [permissions] is not empty, the connection can be active only when one of the specified users is logged into an active session." since that is the issue this is solving.

jandryuk:
Yeah, as an alternative to patching NM.  In general, I prefer modifying OpenXT code to patching upstream code.  Your NM patch is more robust since you aren't just unilaterally removing a line.

crogers1:
I'm gonna try modifying both db_to_nm and nm_to_db instead of patching networkmanager.

crogers1:
Closing and replacing this PR with:
https://github.com/OpenXT/xenclient-oe/pull/1403

PR #1407, opened 2021-06-15
RFC: rpc-autogen: module for haskell bindings

eric-ch:
Provide generated Haskell bindings as a [cabal module](https://github.com/eric-ch/dbus-gen-bindings). The bindings were generated and versioned using [dbus-gen](https://github.com/eric-ch/dbus-gen/tree/haskell-bindings), then are provided by the rpc-autogen recipe.

This makes it easier to track changes across versions and avoid depending on generated sources in-tree for the haskell components.

The goal of this is to provide a reasonably easy to modify alternative to [rpcgen](https://github.com/OpenXT/idl/tree/master/rpcgen) that could still, with low efforts, target different languages.

To avoid changes in existing Haskell projects the module names have been preserved. Changes to the cabal configuration is still necessary though. The GHC 6 dependent projects being the main consumer of the generated bindings, these were targeted first. As, given the age of the runtime, they are the most likely to become a target for replacement.

In this current implementation, only haskell bindings are generated by dbus-gen, the other generated bindings (for OCaml, C and JavaScript) still rely on [rpcgen](https://github.com/OpenXT/idl/tree/master/rpcgen).

#### Dependencies:

- https://github.com/OpenXT/xclibs/pull/12
- https://github.com/OpenXT/manager/pull/184
- https://github.com/OpenXT/network/pull/21

#### External dependencies:

- rpc-autogen: https://github.com/eric-ch/dbus-gen-bindings
- dbus-gen: https://github.com/eric-ch/dbus-gen/tree/haskell-bindings

### DIFF
--- a/recipes-openxt/dbus-gen-bindings/rpc-autogen_0.1.0.0.bb
+++ b/recipes-openxt/dbus-gen-bindings/rpc-autogen_0.1.0.0.bb
@@ -1,0 +1,17 @@
+SUMMARY = "Generated Haskell DBus bindings used by OpenXT components."
+
+LICENSE = "GPLv2"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/GPL-2.0;md5=801f80980d171dd6425610833a22dbe6"
+
+SECTION = "devel"
+
+DEPENDS = " \
+    libxch-rpc \
+"
+
+SRCREV = "${AUTOREV}"
+SRC_URI = "git://github.com/eric-ch/dbus-gen-bindings.git;protocol=https;branch=master"
+
+S = "${WORKDIR}/git/haskell"
+
+inherit haskell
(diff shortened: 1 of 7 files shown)


dpsmith:
Big +1 on the approach. While the bindings generation may not be a significant contributor to full build time, every bit helps. This is also is very useful for development, in particular external development.

eric-ch:
Some clarifications:
- The content of [dbus-gen-bindings](https://github.com/eric-ch/dbus-gen-bindings) is the output of [openxt-haskell-stubs.sh](https://github.com/eric-ch/dbus-gen/blob/haskell-bindings/openxt-haskell-stubs.sh) against the currently OpenXT provided [IDLs](https://github.com/OpenXT/idl/tree/master/interfaces) and the ones provided with [NetworkManager 1.18.4](https://github.com/OpenXT/xenclient-oe/blob/master/recipes-connectivity/networkmanager/networkmanager_1.18.4.bbappend), with matching nm-applet.
- The bindings generated are intended to be semantically identical to the ones generated by rpcgen. Yet, the indentation and some expressions do not match, which will generate diff noise if compared as such.

PR #1410, opened 2021-07-07
initscripts: attend bootstrap races and standardization

eric-ch:
There is a possible race during bootstrap between dbd, xenmgr and vGlass. Should dbd or xenmgr not own their DBus services, vGlass will abort and not restart. Similarly this has been seen with other daemons: https://github.com/OpenXT/xenclient-oe/pull/1133. As a mitigation, wait for the daemon to expose its service on DBus before leaving the initscript.

Some initscripts were showing signs of bitrot, so this changeset also attempt at restoring some of them (xenmgr, xcpmd, dbd, tapback). Note that the loudest initscripts: xenclient-boot-sound and tapdisk are still very noisy.

This layer overwrites `/etc/init.d/functions`. Since some upstream recipes might provide initscripts that rely on it, the additional functions shared by OpenXT initscripts are instead put in separate files: `/etc/init.d/functions-selinux` and `/etc/init.d/functions-dbus`.

Dbd also did not manage a pidfile. This is added in https://github.com/OpenXT/manager/pull/185, so the refpolicy is amended to allow this behavior.

Additionally, xenmgr should support daemonizing on its own yet `start-stop-daemon` is used instead. This is because, if started as a daemon, xenmgr will fail to daemonize hanging forever on a lock (likely [`forkProcess` warning](https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-Process.html).

Dependencies:
- https://github.com/OpenXT/manager/pull/185
- (Optional) https://github.com/OpenXT/network/pull/22


recipes-core/initscripts/initscripts-1.0/udev-volatiles.sh
```diff
@@ -3,25 +3,25 @@
 # Populate early tmpf that may be required by udev.
 #
 
-. /etc/default/rcS
-. /etc/init.d/functions
+set -e
```
jandryuk, line 6: Is this `set -e` intentional?

recipes-openxt/manager/xenmgr/xenmgr.initscript
```diff
@@ -17,90 +17,84 @@
...
+# Default-Stop:         0 1 6
+# Short-Description:    Starts OpenXT Xen manager.
+### END INIT INFO
+
+DESC="XenMgr"
+EXEC="/usr/bin/xenmgr"
+PIDFILE="/var/run/xenmgr.pid"
+OPTS=("--writepid=$PIDFILE" "--no-daemonize")
+
+. /etc/init.d/functions-selinux
```
jandryuk, line 34: The commit message says `Use functions-dbus to get restore.` but I tyink you mean functions-selinux.

recipes-core/initscripts/initscripts-1.0/udev-volatiles.sh
eric-ch: It is, although this one would only stop the initscript if `/etc/init.d/functions-selinux` is missing really.

recipes-openxt/manager/xenmgr/xenmgr.initscript
eric-ch: Indeed, the commit message is wrong.

recipes-core/initscripts/initscripts-1.0/functions-dbus
```diff
@@ -0,0 +1,29 @@
...
+
+# Usage: dbus_name_has_owner NAME
+# Returns an error code if the service has not been claimed, else success.
+dbus_name_has_owner () {
+    local name="$1"
+
+    dbus-send --system --type=method_call --print-reply \
+        --dest=org.freedesktop.DBus /org/freedesktop/DBus \
+        org.freedesktop.DBus.NameHasOwner \
+        string:"com.citrix.xenclient.db" >/dev/null 2>&1
```
jandryuk, line 16: `com.citrix.xenclient.db` should be `$name`?

recipes-openxt/vusb/files/xenclient-vusb.initscript
```diff
@@ -17,43 +17,40 @@
...
+DESC="VUSBd"
+EXEC="/usr/bin/vusb-daemon"
+PIDFILE="/var/run/vusb.pid"
+OPTS=""
 
-# Make sure the progam exists
-[ -f "$PROG" ] || exit 0
-
-start() {
-	"$PROG" 2>&1 | logger -t 'vusb-daemon' &
```
jandryuk, line 30: vusb-daemon prints to stderr and not syslog, so you need this pipe to logger.  That may make it incompatible with start-stop-daemon.

recipes-core/initscripts/initscripts-1.0/functions-dbus
eric-ch: My apologies, I messed up something, I thought I had attended that already.
I will refresh the branch today.

recipes-openxt/vusb/files/xenclient-vusb.initscript
eric-ch: Right, lets move that to another PR then.

eric-ch:
#### v2:

- Removed the vusb initscript refactor, this should first have vusb daemon use syslog directly.
- Sanitize the PR (functions-dbus)
- Amend commit messages.

eric-ch:
v3:
- tapback shouldn't be started as --background
- Amend copy/pasted comments in the LSB headers (only informational, nothing parses them)
- Refresh language-sync as well to handle sending the TERM process to descendants of the daemonized script.
- Amend the stop sequence order to stay coherent with dependencies of each daemon.

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/22/builds/41): [$master-custom-41](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom/41/)

jandryuk (approved):
Looks good.

I checked and all the dbus_wait_for_service initscripts have `set -e`, so they will fail without printing "OK".

I'm slightly nervous that changing the shutdown sequence might regress something.  But that shutdown sequence is so clearly wrong that this is definitely an improvement.   I think we take it and fix things up going forward.

jandryuk:
Will merge soon

recipes-openxt/manager/xenmgr/xenmgr.initscript
```diff
@@ -17,90 +17,84 @@
...
-			rm -f /storage/uivm/uivm-suspend-image
-			;;
-	esac
-done
+# Handle options passed by the bootloader on the kernel cmdline.
+cmdline_options() {
+    read -r -a cmdline_opts < /proc/cmdline
+    for i in "${cmdline_opts[@]}"; do
+        case "$i" in
+            no-autostart)
```
jandryuk, line 44: You flipped the matched string from `autostart` to `no-autostart`, but the grub.cfg/openxt.cfg only add `autostart` or nothing.  Now VMs are always started for regular or console boots.

Defaulting to autostart and disabling on no-autostart makes sense, but that isn't backwards compatible.  I guess that is fine, but "everyone" will have to update any customized openxt.cfg/grub.cfg files.

eric-ch: That qualifies as a regression, I can open a new PR and change it back.

PR #1412, opened 2021-07-13
tpm-scripts: fix tpm_get_pcr grep format

eric-ch:
For TPM1.2, /sys/class/tpm/tpm0/pcrs is [formatted with](https://elixir.bootlin.com/linux/v5.4.131/source/drivers/char/tpm/tpm-sysfs.c#L118): 'PCR-%02d: %02X %02X...'
The use of grep does not account for the 0 padding for PCRs below 10, which breaks forward seal on UEFI/TPM1.2 systems. The bug is mostly silent (apart from failing to unseal after upgrading/forward-sealing):
- tpm_get_pcr returns an empty string for PCR 8
- `! contains_only "$pcr8" "0"` is used to detect a UEFI measured   installation. `contains_only` will return true for   an empty-string given for the PCR value, for any character value.
- In turn, `seal-system` calls `calculate_drtm_pcr` for legacy installations, using `/boot/system/grub/grub.cfg` and setting SRTM to 0.

This leads to PCR18 and PCR19 not being predicted correctly, but predicted nonetheless while no error is apparently raised when performing the upgrade/forward-seal.

### DIFF
--- a/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
+++ b/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
@@ -25,8 +25,9 @@ contains_only() {
     local s="$1"
     local c="$2"
 
-    # Remove all instances of c from s and check if the result is empty
-    [[ -z "${s//$c}" ]]
+    # Check if s is empty then remove all instances of c from s and check if
+    # the result is empty.
+    [[ -n "$s" ]] && [[ -z "${s//$c}" ]]
 }
 
 usage() {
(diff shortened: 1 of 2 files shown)


jandryuk:
Would you please expand the commit message with something like:
Currently, forward seal on a UEFI/TPM1.2 system is broken.  The check for PCR-8 returns an empty string, which matches in contains_only 0.  That in turn causes calculate_drtm_pcrs to use the legacy grub.cfg instead of openxt.cfg leading to an incorrect PCR-19.  With this change, forward seal is against grub.cfg.

You can use that verbatim if you like.

jandryuk (approved):
Code LGTM

eric-ch:
@jandryuk Yes of course, the commit message is too succinct. Also, I wanted to fix the false-positive in `contains_only`, but did not push it.

jandryuk (approved):
LGTM.  Thanks.

jandryuk:
Will merge soon.

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/22/builds/44): [$master-custom-44](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom/44/)

PR #1414, opened 2021-07-15
xenmgr: regression on cmdline argument autostart

eric-ch:
Xenmgr can be configured to not autostart VMs at startup, this used to be handled by the initscript by checking "autostart" on the kernel cmdline.
Commit: ae940b14 xenmgr: cleanup initscript introduced a regression on a knee-jerk change from "autostart" to "no-autostart" being checked on the command line, yet the configuration files used at bootstrap were not changed.

This is in any case a regression that would affect any customised openxt.cfg or grub.cfg, making it not worth it.

Restore the original behavior: normal operation boot entries add "autostart" to the kernel cmdline, "no-autostart" is the default.

Regression introduced by: https://github.com/OpenXT/xenclient-oe/pull/1410 (https://github.com/OpenXT/xenclient-oe/pull/1410#discussion_r670423609)

### DIFF
--- a/recipes-openxt/manager/xenmgr/xenmgr.initscript
+++ b/recipes-openxt/manager/xenmgr/xenmgr.initscript
@@ -29,7 +29,7 @@
 DESC="XenMgr"
 EXEC="/usr/bin/xenmgr"
 PIDFILE="/var/run/xenmgr.pid"
-OPTS=("--writepid=$PIDFILE" "--no-daemonize")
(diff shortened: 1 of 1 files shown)


jandryuk (approved):

eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/22/builds/47): [$master-custom-47](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom/47/)

PR #1415, opened 2021-07-16
initscripts: fix dbus wait helpers

eric-ch:
Commit:
2ef30c8c initscripts: dbus wait for service providers
Did not handle its purpose correctly.

dbus-send will not return an error code regardless of the method_call result as long as the method was called successfully. Since the signature returns a boolean this is easy to grep.

The timeout is expected to be roughly in seconds, but the loop does not wait, so add that.


eric-ch:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/22/builds/47): [$master-custom-47](https://openxt.ainfosec.com/OpenXT/auto/master/master-custom/47/)

jandryuk:
Will merge soon.

PR #1417, opened 2021-07-21
linux: syncvm: Enable hypervisor uuid

jandryuk:
This will expose the domain's uuid to userspace through
/sys/hypervisor/uuid.  This lets userspace like the sync-client just
read the sysfs file instead of having to read xenstore to get the uuid.


recipes-kernel/linux/5.4/defconfigs/xenclient-syncvm/defconfig
```diff
@@ -106,7 +106,6 @@ CONFIG_VIRT_DRIVERS=y
 # CONFIG_XEN_BALLOON is not set
 # CONFIG_XEN_DEV_EVTCHN is not set
 CONFIG_XEN_BACKEND=y
-# CONFIG_XEN_SYS_HYPERVISOR is not set
```
crogers1, line 109: This might need to explicitly set CONFIG_XEN_SYS_HYPERVISOR=y like the dom0 defconfig unless OE is doing something under the hood.

Also, I checked the value of /sys/hypervisor/uuid in a VM that had this compile option turned on and got the same uuid as in dom0 (all 0's).

jandryuk: Just dropping the line works.  The defconfig is saved from bitbake -c savedefconfig (which runs linux's make savedefconfig), which only stores non-default values.  CONFIG_XEN_SYS_HYPERVISOR is default y in the Kconfig, so it gets enabled when linux's `make olddefconfig` runs.

`/sys/hypervisor/uuid` works here in the syncvm and in a CentOS 7 guest.  The code in drivers/xen/sys-hypervisor.c first tries HYPERVISOR_xen_version(XENVER_guest_handle) and then falls back to effectively `xenstore-read $( xenstore-read vm )/uuid )`.  The OpenXT xsm flask policy only allows xen_guest_handle for domains defined with xen_hvm_type().  I'm using a customized policy, so my syncvm may not be exercising the fallback.  Running `xenstore-read $( xenstore-read vm )/uuid )` gave the correct uuid which should be equivalent to the fallback.  I'll try to test on a vanilla openxt build.

jandryuk: I tested a vanilla-ish build with this PR and the syncvm-pvh PRs and /sys/hypervisor/uuid showed the correct uuid.  @crogers1 Did your VM's xl.cfg have a uuid line defined?

crogers1: Yea it did.  Maybe there's something else wrong with how I have the guest configured.  Either way, if it's working on your latest test then that's fine.

crogers1 (approved):

crogers1:
Merging soon

PR #1418, opened 2021-07-22
refpolicy fixes

jandryuk:
Two fixes:

The first restores sysadm's ability to list /dev/xen/blktap-2.  A side effect of this is that sysadm can read/write the /dev/xen/blktap-2/tapdev* nodes.  This was not allowed in stable-9 because those nodes were labeled xen_device_t.  sysadm can already mount the tapdevs, so I think this is okay.  Since OpenXT was open sourced, there's existed: storage_rw_blktap(sysadm_t), so I think this was intended.  sysadm_t can modify the .vhds directly anyway.

The second fixes some avc denials with writing /config/asound{,.new} on shutdown.


crogers1:
LGTM.  Will have a custom build kicked off soon.

crogers1:
[Build](https://openxt.ainfosec.com/buildbot/#/builders/22/builds/48)

dpsmith (approved):

crogers1:
Merging soon.

PR #1420, opened 2021-07-29
linux: syncvm: enable PVH mode

jandryuk:
Enable XEN_PVH_GUEST so the syncvm can be run as virt-type PVH to avoid
spectre and meltdown issues.

To enable PVH, we also have to enable ACPI.  That brings in a bunch of
unnecessary ACPI defaults which are then turned off.

Dom0 support is also turned off since this isn't dom0.


recipes-kernel/linux/5.4/defconfigs/xenclient-syncvm/defconfig
```diff
@@ -33,8 +34,13 @@ CONFIG_X86_CPUID=y
 CONFIG_HZ_100=y
 # CONFIG_RANDOMIZE_BASE is not set
 CONFIG_LEGACY_VSYSCALL_EMULATE=y
-# CONFIG_ACPI is not set
```
crogers1, line 36: I assume this is another instance where CONFIG_ACPI is default =y in the Kconfig?

jandryuk: Yes, you are correct - `default y if X86`.

crogers1 (approved):

crogers1:
LGTM, merging soon

PR #1422, opened 2021-08-26
xen: Run qemu with -no-shutdown to avoid timeout

jandryuk:
libxl-RFC-5of7-Handle-Linux-stubdomain-specific-QEMUoptions.patch moved
-no-shutdown as "There is currently no way to access the QMP socket in
the stubdom".

libxl-openxt-qemu-args.patch introduces a way to access QMP, but it did
not re-add -no-shutdown.

Without -no-shutdown, QEMU exits when the VM stops.  libxl expects QEMU
to still be running to issues "stop" and "quit" commands.  When QEMU is
gone, we get 10 seconds (2 x 5) of delay while libxl waits for the
timeouts. By re-adding "-no-shutdown" we avoid this shutdown delay.

On other issue with QEMU exiting is that it closes up its vGlass
connection which results in a slew of "xen:grant_table: WARNING: g.e.
0x1200 still in use!" since those grants are still mapped by vGlass in
dom0.  By having QEMU wait for an explicit quit, vGlass starts unmapping
the grants before QEMU.  There are still some "still in use" messages,
but it's ~12 instead of thousands.

### DIFF
--- a/recipes-extended/xen/files/libxl-openxt-qemu-args.patch
+++ b/recipes-extended/xen/files/libxl-openxt-qemu-args.patch
@@ -36,11 +36,12 @@ PATCHES
 ################################################################################
 --- a/tools/libxl/libxl_dm.c
 +++ b/tools/libxl/libxl_dm.c
-@@ -1168,17 +1168,15 @@ static int libxl__build_device_model_arg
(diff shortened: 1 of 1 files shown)


crogers1 (approved):
Good find

crogers1:
Also want to merge this today,  if no objections

PR #1425, opened 2021-09-01
xenmgr: Remove depends on xenmgr-data

jandryuk:
xenmgr-data is the HTML+JS UI from toolstack-data.git.  xenmgr just
doesn't depend on it.  It's unclear why it was added in the first place.
Remove it from depends.

### DIFF
--- a/recipes-openxt/manager/xenmgr_git.bb
+++ b/recipes-openxt/manager/xenmgr_git.bb
@@ -21,7 +21,6 @@ DEPENDS = " \
     hkg-text \
     hkg-mtl \
     hkg-split \
-    xenmgr-data \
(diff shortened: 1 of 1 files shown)


crogers1:
I wonder if it just was a holdover from when the dependencies were a little more intertwined.  If anything xenmgr-data should 
"depend" on xenmgr, since xenmgr would be considered the 'backend'.

But yea, A+ cleanup, I'm fine with merging

crogers1 (approved):

jandryuk:
Thanks.  I had just waited through a xenmgr-data build, so rebuilding xenmgr was adding insult to injury.

PR #1426, opened 2021-09-03
uivm-xsession: Remove resize-agent

jandryuk:
resize-agent monitored the surfman controlled xenstore path
/xc_tools/switcher/current_display_size to resize the uivm's display.
With surfman gone, it is no longer needed.

vGlass resizes through kernel drm uevents.  udev calls
/etc/udev/rules.d/10-openxtfb.rules, which calls
/usr/share/openxt/display_hotplug.sh, which calls xrandr.


crogers1 (approved):
LGTM, uivm works fine without resize-agent.  Merging later.

PR #1427, opened 2021-09-13
[tboot] Add RPROVIDES to tboot

Tmcg2:
This does not fix any problems or issues. It just adds RPROVIDES for each
 of the tboot packages. This helps anyone using opkg find which package
 provides which binary.

For example, running ```opkg whatprovides acmmatch``` currently doesn't
 return any useful information. With this change, it should.

Signed-off-by: Tyler McGavran <mcgavrant@ainfosec.com>

### DIFF
--- a/recipes-security/tboot/tboot.inc
+++ b/recipes-security/tboot/tboot.inc
@@ -77,3 +77,32 @@ FILES_${PN}-pcr-calc = " \
     ${sbindir}/pcr-calc \
     ${sbindir}/acmmatch \
 "
+
+RPROVIDES_${PN}-lcptools += " \
+    lcp_readpol \
+    lcp_writepol \
+    tb_polgen \
+    tpmnv_defindex \
+    tpmnv_getcap \
+    tpmnv_lock \
+    tpmnv_relindex \
+"
+
+RPROVIDES_${PN}-lcptools-v2 += " \
+    lcp2_mlehash \
+    lcp2_crtpol \
+    lcp2_crtpollist \
+    lcp2_crtpolelt \
+"
+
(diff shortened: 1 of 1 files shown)


jandryuk:
Hi, @Tmcg2 

I think "whatprovides" operates in the "package namespace", but "acmmatch" is in the "file namespace".  "search" operates in the "file namespace", but it doesn't match on just the basename:
```
root@xenclient-dom0:~# opkg search acmmatch
root@xenclient-dom0:~# opkg search $( which acmmatch )
tboot-pcr-calc - 1.9.12-r0.3
root@xenclient-dom0:~# opkg search *acmmatch
tboot-pcr-calc - 1.9.12-r0.3
```

So this is conflating the two namespaces together.  Is this really what you want?  You could package each utility by name and achieve a simliar result that way.  What was the motivation for this change?

dpsmith:
This is what @jandryuk was alluding to, but here is what the Yocto/OE definition for RPROVIDES:

_A list of package name aliases that a package also provides. These aliases are useful for satisfying runtime dependencies of other packages both during the build and on the target._

I don't think this is what you are intending to do. As @jandryuk suggest, I think `opkg search` is more than likely what you probably require.

Tmcg2:
I think @jandryuk hit the nail on the head here. I was ill informed on what RPROVIDES, whatprovides, and search did, leading to me thinking that RPROVIDES was related to information stored in an IPK. My motivation was trying to find what IPK provided what binary and ```opkg search``` appears to be what I was actually looking for. Got a bit lazy after whatprovides didn't get the job done and did some assuming.

Thank you for the corrections, will close as this does appear to be an unnecessary change.

PR #1428, opened 2021-09-16
xen: EFI fixes for Dell UEFI

jandryuk:
Some Dell systems have issues booting UEFI.  Marek's 4 EFI
SetVirtualAddressMap patches and corresponding update to defconfig are
needed to boot with efi=no-rs.  Disabling GetTime allows booting with
efi=rs.

I needed these for booting a newer Dell.  Qubes & XenServer remove the GetTime call.  I'm curious if others have needed similar.


crogers1:
I had a hackier solution that just dropped the efi_get_time block altogether, but I didn't need Marek's patchset.  Could have been the hardware I was testing though.

jandryuk:
I needed the Marek ones for a 10th gen Intel, but it does seem firmware dependent.  Disabling GetTime was needed for an 8th and 10th gen.  But it too is probably firmware dependent.

crogers1:
[build](https://openxt.ainfosec.com/buildbot/#/builders/22/builds/56)

crogers1:
Tested master-custom 56. Installed and booted fine on 8th gen Dell 7060. Installed and booted fine on 10th gen Dell Precision 7550.  Merging later.

PR #1432, opened 2021-09-20
qemu-dm: Remove acpi-wakeup.patch

jandryuk:
The patch is no longer used.  xenvm wrote to the wakeup-req node, but it
is no longer used and nothing else writes it.  As the patch itself says,
QMP can be used instead, and that is what xl uses.

Drop the patch.

I don't use sleep and haven't tested much.  My Win10 image doesn't show a sleep option.  An AlmaLinux image went to sleep and resumed.  /var/logs/messages shows qmp_helper being used.  However, the display & cursor seem frozen on resume.  This AlmaLinux image sometimes freezes the UI on login.  hvc0 is showing /etc/issue and accepting a username, but it doesn't prompt for a password.  Something is weird but it did resume.

I am confident that xenvm no longer exists and nothing writes wakeup-req.


crogers1 (approved):
I'm in agreement here.  Merging later.

PR #1433, opened 2021-09-21
Linux 5.10

tsirakisn:


recipes-kernel/linux/5.10/patches/xsa-155-qsb-023-xen-netfront-copy-response-out-of-shared-buffer-befo.patch
```diff
@@ -108,14 +110,14 @@ PATCHES
 +		if (unlikely(rx.status < 0 ||
 +			     rx.offset + rx.status > XEN_PAGE_SIZE)) {
  			if (net_ratelimit())
--				dev_warn(dev, "rx->offset: %u, size: %d\n",
--					 rx->offset, rx->status);
+ 				dev_warn(dev, "rx->offset: %u, size: %d\n",
+ 					 rx->offset, rx->status);
```
jandryuk, line 114: Similiarly this still should remove the `rx->` lines.  I think this patch doesn't compile.

recipes-kernel/linux/5.10/patches/xsa-155-qsb-023-xen-netfront-do-not-use-data-already-exposed-to-back.patch
```diff
@@ -57,3 +57,54 @@ PATCHES
...
  	if (skb->ip_summed == CHECKSUM_PARTIAL)
  		/* local packet? */
+@@ -907,9 +908,8 @@ static int xennet_get_responses(struct n
+ 				struct sk_buff_head *list,
+ 				bool *need_xdp_flush)
+ {
+-	struct xen_netif_rx_response *rx = &rinfo->rx;
+ 	struct xen_netif_rx_response rx = rinfo->rx;
+-	int max = XEN_NETIF_NR_SLOTS_MIN + (rx->status <= RX_COPY_THRESHOLD);
++	int max = XEN_NETIF_NR_SLOTS_MIN + (rx.status <= RX_COPY_THRESHOLD);
```
jandryuk, line 67: This patch has a fixup of the previous patch, xsa-155-qsb-023-xen-netfront-copy-response-out-of-shared-buffer-befo.patch .  It should be moved there.

tsirakisn:
Addressed the patch concerns and added strace diff. OE-tracked version of strace is not yet compatible with 5.10 kernel.

recipes-kernel/linux/5.10/patches/xsa-155-qsb-023-xen-netfront-copy-response-out-of-shared-buffer-befo.patch
```diff
@@ -18,9 +18,11 @@ use issue.
 ################################################################################
 PATCHES 
 ################################################################################
---- a/drivers/net/xen-netfront.c
-+++ b/drivers/net/xen-netfront.c
-@@ -387,13 +387,13 @@ static void xennet_tx_buf_gc(struct netf
+Index: linux-5.10.49/drivers/net/xen-netfront.c
+===================================================================
+--- linux-5.10.49.orig/drivers/net/xen-netfront.c
++++ linux-5.10.49/drivers/net/xen-netfront.c
```
jandryuk, line 24: Please refresh this patch with a config matching the one found here: https://openxt.atlassian.net/wiki/spaces/DC/pages/20905986/Dealing+with+quilt

That cuts down on the version specific churn when refreshing patches.

jandryuk (changes requested):
Good find on strace.  Is it a build time or run time incompatibility?

recipes-kernel/linux/5.10/patches/xsa-155-qsb-023-xen-netfront-copy-response-out-of-shared-buffer-befo.patch
tsirakisn: > Please refresh this patch with a config matching the one found here

Will do.

> Good find on strace. Is it a build time or run time incompatibility?

Likely both, build time in this case. strace seems to align itself with the same versioning as linux does (eg v5.10 strace is the first strace version to be compatible with v5.10 linux, and so on). v5.12 was the latest version at the time of our downstream work, so we went with that.

jandryuk: Yes, adding the .quiltrc to the bordel docker image is a good idea.

jandryuk:
I think this looks good.  I'm building locally... and will need to find a free machine to test on.

tsirakisn:
@jandryuk automated build here: https://openxt.ainfosec.com/OpenXT/auto/master/master-custom/59/

I need to do some local testing as well. I also need to force push one more time to add OXT-1825 to the commits

crogers1 (approved):
LGTM

jandryuk (commented):
I was about to click approve, but I noticed these two changes are missing a Signed-off-by:

https://github.com/OpenXT/xenclient-oe/pull/1433/commits/465bfb4f8f351777bdd09a48137979e787f75e63
https://github.com/OpenXT/xenclient-oe/pull/1433/commits/ccc690a6499422d82957ad3b78038822074e23c0

tsirakisn:
> I was about to click approve, but I noticed these two changes are missing a Signed-off-by:

Fixed

jandryuk (approved):
Thanks.

I will test a little more locally.  If everything goes well, I'll merge later this afternoon.

PR #1434, opened 2021-09-27
[guest-efi] Do not merge yet; read description

tsirakisn:
This PR is the first of a series of PRs geared towards supporting Guest EFI & Guest Secure Boot on OpenXT. Before committing the rest of the changes, I would like to give a high-level overview of the solution to make sure it is still of interest:

Background:
- EFI makes use of NVRAM to store variables that need to persist between guest boots
- This data includes boot configs, driver info, and cert information used to verify signed modules for SB
- There is currently no method of accessing this database in Xen (or creating it in the first place)
- Solution: use a backend/frontend pair in dom0/OVMF, respectively, to allow the guest to utilize this capability
	- Note that this solution was taken from XAPI and adapted to OpenXT

OVMF:
- An OVMF module called XenVariable acts as the frontend
- This module proxies EFI variable calls to the backend daemon that runs in dom0
- IOREQ server used for communication to backend
- Changes needed: uprev OVMF to 2020, add patch for XenVariable

varstored:
- varstored is the backend daemon that receives requests from OVMF
- One instance of varstored runs for each running guest and is killed on guest shutdown
- EFI variables are ultimately persisted in the OpenXT db
- Helper binaries are provided for debug (eg varstore-ls, varstore-get, varstore-set, etc)
- Insecure (test) certs are provided for SB in /usr/share/varstored. User should replace these with real certs.
- Changes needed: add varstored recipe, add patch to use OpenXT db instead of XAPI db

varstored-watch:
- Helper binary that starts and watches a single varstored instance and restarts it if it dies
- Added as an additional protection against unexpected varstored crashes
- Guest might be unrecoverable if varstored dies and doesn't restart before the guest shuts down, especially if using SB
- Changes needed: add varstored-watch binary to xctools

toolstack changes:
- Add "secureboot" and "authenforce" boolean VM properties
- "secureboot" key is self-explanatory. When secureboot is enabled, SB certs are loaded from /usr/share/varstored
- "authenforce" will determine if SB is running in permissive mode or enforcing mode
- Keys are read by varstored

libxl changes:
- Start varstored-watch (which starts varstored) on guest startup and kill on shutdown
- Write secureboot/authenforce keys to xenstore on guest startup for varstored to read

Miscellaneous changes:
- SELinux rules
- UI support for changing guest boot type & enabling SB


jandryuk:
XenServer and XCP-ng are using this approach, so I think it's sensible to follow.  Your general outline seems fine.

> Insecure (test) certs are provided for SB in /usr/share/varstored. User should replace these with real certs.

These are the certs populated into the initial keystore?  Can these just be the Microsoft ones typically shipped on hardware?

Do you have an estimate of how much space is used storing certs in OpenXT db?

Out of curiosity, does XenServer/XCP-ng use something equivalent to varstored-watch or do they just run varstored directly?  How often is varstored crashing?

> Solution: use a backend/frontend pair in dom0/OVMF, respectively, to allow the guest to utilize this capability

I looked at varstored a while ago and it relied on mapping foreign memory and not grant tables.  That is unfortunate since you don't have fine grain restrictions.

tsirakisn:
> These are the certs populated into the initial keystore? Can these just be the Microsoft ones typically shipped on hardware?

Correct, they're generated at build time. I need to look into the Microsoft certs again cause afaik they aren't fair-use, at least for distribution purposes. I think the general expectation is that if you need Secure Boot enabled, you're responsible for finding your certs and installing them. 

> Do you have an estimate of how much space is used storing certs in OpenXT db?

The default certs are about 6kb
 
> Out of curiosity, does XenServer/XCP-ng use something equivalent to varstored-watch or do they just run varstored directly? How often is varstored crashing?

I've never personally seen it crash, but the effects of it crashing can be fatal to the guest, so we added an extra layer of protection. xcp-ng has a varstored-guard, which I believe has some pid monitoring but also does other stuff. The main problem with it is that it's written in ocaml, which afaik is a no-go. Looping @deepin125 (Joel) into this discussion as I don't remember all the details of the guard. IIRC the usecase was different than ours.

> I looked at varstored a while ago and it relied on mapping foreign memory and not grant tables. That is unfortunate since you don't have fine grain restrictions.

Agreed.

deepin125:
> > These are the certs populated into the initial keystore? Can these just be the Microsoft ones typically shipped on hardware?
> 
> Correct, they're generated at build time. I need to look into the Microsoft certs again cause afaik they aren't fair-use, at least for distribution purposes. I think the general expectation is that if you need Secure Boot enabled, you're responsible for finding your certs and installing them.
> 

Not sure how xcpng is dealing with the certs.  It looks as if Xenserver is creating certs that are designed to allow the SecureBoot process for Citrix and Microsoft GUIDS to allow for Windows to have SB enabled, and because they are being built into the Db.auth, you don't need any permission from MS:
https://github.com/xapi-project/varstored/blob/master/create-auth.c

> > Do you have an estimate of how much space is used storing certs in OpenXT db?
> 
> The default certs are about 6kb
>
 
I have about 29K file if I do a db-dump of just the varstored-efivar-blob/0 node.

> > Out of curiosity, does XenServer/XCP-ng use something equivalent to varstored-watch or do they just run varstored directly? How often is varstored crashing?
> 
> I've never personally seen it crash, but the effects of it crashing can be fatal to the guest, so we added an extra layer of protection. xcp-ng has a varstored-guard, which I believe has some pid monitoring but also does other stuff. The main problem with it is that it's written in ocaml, which afaik is a no-go. Looping @deepin125 (Joel) into this discussion as I don't remember all the details of the guard. IIRC the usecase was different than ours.
> 

I have never seen varstored crash in the year that I have used it everyday. However I could simulate a crash by killing it, and found that you could have issues with guests shutting down with PCI passthrough, so we decided to create a varstored-watch program to bring it back up.  Xenserver has varstored-guard which is a program that runs in dom0 to write their XAPI database.

> > I looked at varstored a while ago and it relied on mapping foreign memory and not grant tables. That is unfortunate since you don't have fine grain restrictions.
> 
> Agreed.

I'm a little confused as to what you mean, since the program does restrict which domU's it can access in the early setup of the varstored setup.  It uses xentoolcore_restrict_all and honestly it seemed to lockdown the process to only dealing with that domU that has access to for the foreign memory.  I believe it was used because QEMU, Xen, and the HVM guest all need access to the same memory (efivars is used for the hardware emulated devices at times), so this was the chosen method, and there were improvements needed in order to run this so uprevving to 4.1 was needed for the correct foreign memory libraries to be used for QEMU (+ XEN 4.12).

tsirakisn:
Merge with:
https://github.com/OpenXT/manager/pull/187
https://github.com/OpenXT/idl/pull/44
https://github.com/OpenXT/xctools/pull/69

recipes-core/varstored/varstored_git.bb
```diff
@@ -0,0 +1,76 @@
...
+
+SRC_URI = " \
+    git://github.com/xapi-project/varstored.git;protocol=https;branch=master \
+    file://0001-add-oxtdb-varstore.patch \
+"
+
+# download the ever-growing community dbx.auth file, which contains
+# a list of known malicious guids that we should never boot with.
+SRC_URI += " \
+    https://uefi.org/sites/default/files/resources/dbxupdate_x64.bin;downloadfilename=dbx.auth \
```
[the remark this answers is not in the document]
jandryuk: I think it should be included.  People can turn off secureboot like they would on a real machine to bypass it.

recipes-core/ovmf/ovmf_git.bbappend
```diff
@@ -1,6 +1,16 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+
+# xcp-ng patches taken from:
```
jandryuk, line 3: It's a little weird the commit message says "Taken from XAPI", but the patches are prefixed xcp-ng.  Yes, I realized xapi is the upstream for xcp-ng.

recipes-security/libseccomp/libseccomp_2.4.3.bb
```diff
@@ -0,0 +1,43 @@
+SUMMARY = "interface to seccomp filtering mechanism"
```
jandryuk, line 1: Why not include meta-security instead of importing a recipe?

commit messages has a typo "libbseccomp" double b.

recipes-core/varstored/varstored/0001-add-oxtdb-varstore.patch
```diff
@@ -0,0 +1,864 @@
...
+################################################################################
+varstored can store EFI variables in an arbitrary format. This patch replaces
+the current storage backend (XAPI db) with the OpenXT db. The OpenXT db was
+the preferred choice as it already propagates across host upgrades, which means
+we don't need to worry about carrying EFI guest information over to upgraded
+installs.
+
+Also, remove `chroot` and `unshare` operations in order to avoid granting
+varstored SYS_ADMIN capabilities. We don't need these features and we want to
+keep varstored as limited as possible (SYS_ADMIN would be too permissive).
```
jandryuk, line 17: I think there is a way to drop SYS_ADMIN, so you could.  Before this patch header, I had wondered if namespaces caused a problem with connecting to DBus IPC.

recipes-core/varstored/varstored_git.bb
```diff
@@ -0,0 +1,76 @@
...
+    dbus-glib \
+    libseccomp \
+    libxml2 \
+    libxcdbus \
+    openssl \
+    xen-tools \
+"
+
+CFLAGS_prepend += "-I${STAGING_INCDIR}/dbus-1.0 -I${STAGING_LIBDIR}/dbus-1.0/include -I${STAGING_INCDIR}/glib-2.0 -I${STAGING_LIBDIR}/glib-2.0/include "
+LDFLAGS_prepend += " -levent -lgobject-2.0 -lglib-2.0 -ldbus-glib-1 -lxcdbus "
```
jandryuk, line 18: These lines look wrong.  Seems like the Makefile should be patched to add them.

recipes-core/varstored/varstored_git.bb
```diff
@@ -0,0 +1,76 @@
...
+SRC_URI[sha256sum] = "46ba1f2a0a2ed7aabe20f9b7b2a8d717cb0b514cea83c7a1a24fe25f6b208784"
+
+USERADD_PACKAGES = "${PN}"
+USERADD_PARAM_${PN} = "--system --no-create-home \
+                       --shell /bin/false \
+                       --groups varstored \
+                       --gid 415 \
+                       --uid 416 \
+                       varstored"
+GROUPADD_PARAM_${PN} = "--system --gid 415 varstored"
```
jandryuk, line 48: I was going to comment on not having hardcoded uid & gid.  Looking ahead, varstored only takes numeric values, and varstored-watch hardcodes these numbers.  Which is a little unfortunate, but understandable.

recipes-core/varstored/varstored_git.bb
```diff
@@ -0,0 +1,76 @@
...
+    oe_runmake db.pem
+
+    echo ${S}/KEK.pem > ${S}/KEK.list
+    echo ${S}/db.pem > ${S}/db.list
+
+    oe_runmake auth
+}
+
+do_install() {
+    install -m 0700 -o root -g root -d ${D}/usr/sbin
```
jandryuk, line 67: You don't need to specify all the -o/-g root - that's the default.  mode 0700 is too strict for this public directory - you can just do `install -d ${D}/usr/sbin`.

recipes-openxt/xctools/varstored-watch_git.bb
```diff
@@ -0,0 +1,26 @@
+SUMMARY = "Provides binary that restarts varstored on failure"
+LICENSE = "GPLv2"
+LIC_FILES_CHKSUM = "file://COPYING;md5=4641e94ec96f98fabc56ff9cc48be14b"
+
+DEPENDS = " \
+    xen-tools \
+"
+
```
jandryuk, line 8: RDEPENDS_${PN} += " \
    varstored \
"

recipes-openxt/xctools/varstored-watch_git.bb
```diff
@@ -0,0 +1,26 @@
...
+export INCDIR = "${STAGING_INCDIR}"
+export LIBDIR = "${STAGING_LIBDIR}"
+
+SRCREV = "${AUTOREV}"
+PV = "0+git${SRCPV}"
+
+S = "${WORKDIR}/git/varstored-watch"
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
```
jandryuk, line 17: This is unnecessary - no local files in SRC_URI (and I think it's the default for a .bb)

recipes-openxt/xctools/varstored-watch_git.bb
```diff
@@ -0,0 +1,26 @@
+SUMMARY = "Provides binary that restarts varstored on failure"
+LICENSE = "GPLv2"
+LIC_FILES_CHKSUM = "file://COPYING;md5=4641e94ec96f98fabc56ff9cc48be14b"
+
+DEPENDS = " \
+    xen-tools \
+"
+
+export INCDIR = "${STAGING_INCDIR}"
+export LIBDIR = "${STAGING_LIBDIR}"
```
jandryuk, line 10: I think these are not necessary with a Makefile change in xctools.

recipes-extended/xen/xen-common.inc
```diff
@@ -122,6 +122,7 @@ SRC_URI_append = " \
     file://0003-efi-use-directmap-to-access-runtime-services-table.patch \
     file://0004-efi-do-not-use-runtime-services-table-with-efi-no-rs.patch \
     file://patch-xen-disable-efi-gettime.patch \
+    file://libxl-add-ovmf-varstored.patch \
```
jandryuk, line 125: Wherever it is appropriate, you should add
```
RDEPEND_xen-tools-libxenlight += \
    varstored-watch \
"
```

recipes-extended/xen/files/libxl-add-ovmf-varstored.patch
```diff
@@ -0,0 +1,202 @@
...
+ {
+     STATE_AO_GC(sdss->dm.spawn.ao);
+@@ -2349,6 +2389,16 @@ void libxl__spawn_stub_dm(libxl__egc *eg
+         fork_helper(gc, AUDIO_HELPER_PATH, XS_AUDIO_PID, guest_domid, dm_domid);
+     }
+ 
++    if (guest_config->b_info.u.hvm.bios == LIBXL_BIOS_TYPE_OVMF) {
++        const char *dom_path = libxl__xs_get_dompath(gc, guest_domid);
++        /* Find uuid. */
++        t = xs_transaction_start(ctx->xsh);
```
jandryuk, line 64: This transaction is never finished?

recipes-extended/xen/files/libxl-add-ovmf-varstored.patch
```diff
@@ -0,0 +1,202 @@
...
++}
++
+ void libxl__spawn_stub_dm(libxl__egc *egc, libxl__stub_dm_spawn_state *sdss)
+ {
+     STATE_AO_GC(sdss->dm.spawn.ao);
+@@ -2349,6 +2389,16 @@ void libxl__spawn_stub_dm(libxl__egc *eg
+         fork_helper(gc, AUDIO_HELPER_PATH, XS_AUDIO_PID, guest_domid, dm_domid);
+     }
+ 
++    if (guest_config->b_info.u.hvm.bios == LIBXL_BIOS_TYPE_OVMF) {
```
jandryuk, line 61: I thought this would check libxl_defbool_val(info->u.hvm.secureboot) to skip starting the watcher/varstored.  Does that still work?

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.te
```diff
@@ -119,3 +126,48 @@ allow audio_helper_t self:shm create_shm_perms;
...
+#
+# Local Policy: varstored
+#
+gen_require(`
+	type xen_device_t, xenstored_t, xenstored_var_run_t;
+')
+
+# varstored-watch executing & monitoring varstored
+allow varstored_t bin_t:dir search;
+allow varstored_t initrc_t:fd use;
```
jandryuk, line 140: This should probably be a dontaudit.

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.te
```diff
@@ -119,3 +126,48 @@ allow audio_helper_t self:shm create_shm_perms;
...
+allow varstored_t initrc_t:unix_stream_socket connectto;
```
jandryuk, line 141: This looks like the other end isn't labeled properly.  Or it should not be allowed.

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.te
```diff
@@ -119,3 +126,48 @@ allow audio_helper_t self:shm create_shm_perms;
...
+allow varstored_t varstored_exec_t:file execute_no_trans;
+allow varstored_t self:unix_dgram_socket { connect create write };
+allow varstored_t self:unix_stream_socket { connect create getattr read write };
```
jandryuk, line 144: These should be something like:
allow varstored_t self:unix_stream_socket create_socket_perms;

Similar for below, there should be more use of interfaces, like `xen_stream_connect_xenstore(varstored_t)`

I do appreciated the comments explaining the groupings of rules.

recipes-core/packagegroups/packagegroup-xenclient-dom0.bb
```diff
@@ -110,6 +110,9 @@ RDEPENDS_${PN} = " \
     udev-extraconf-dom0 \
     vglass \
     disman \
+    libseccomp \
```
jandryuk, line 113: libseccomp should be added automatically as a varstored rdepends (there aren't any command line tools packaged inside that are needed, right?).  With proper RDEPENDS in place, this patch can go away.

jandryuk (commented):
This was non-exhaustive.  I only skimmed the varstored patch (but it was generally okay when I looked at the RFC)

recipes-security/libseccomp/libseccomp_2.4.3.bb
jandryuk: meta-security, the top level layer, doesn't have many bbappends.  There is a busybox one, but that only matters for ptest.  The linux ones only apply to linux-yocto, so those wouldn't affect use.  So at first glance, it seems like it would be okay to include.

dunfell meta-security has libseccomp, but it is removed in the master branch as it moved to core: https://git.yoctoproject.org/cgit/cgit.cgi/meta-security/commit/?id=baca6133f9c7f77a12dc137fe5b90723fbb4c15b

recipes-core/varstored/varstored_git.bb
tsirakisn: I'm not too big a fan of this either. We were having issues with the IDs in OE and hardcoding seemed to solve it. I noticed that other recipes have the IDs hardcoded as well (trousers maybe?), so I stuck with it.

recipes-extended/xen/files/libxl-add-ovmf-varstored.patch
tsirakisn: The varstore is not dependent on secureboot, there are a bunch of other EFI vars that it tracks (eg BootOrder, PlatformLang, MemoryTypeInformation, etc). It should be started for all EFI guests.

recipes-openxt/xctools/varstored-watch_git.bb
tsirakisn: Good catch, this might have been a holdover.

recipes-core/varstored/varstored_git.bb
tsirakisn: Hm can we dynamically find the headers in the staging dir if we move to the Makefile? I thought autotools were needed for that kind of thing in OE. That's why I went with this approach.

jandryuk: For CFLAGS, I'm not sure how you could add a sysroot-prefixed directory to the includes.  But you can sidestep that issue by expanding relevant include lines to add the prefix, like dbus-1.0/dbus/dbus.h

LDFLAGS additions should be in the Makefile

recipes-core/ovmf/ovmf_git.bbappend
tsirakisn: Joel and I are honestly a little confused ourselves. xcp-ng was the original source of the patches as XAPI hadn't upstreamed their solution yet when we started this work. Now that varstored is in XAPI proper, there should theoretically be a corresponding set of changes in ovmf, but we couldn't find any, neither in upstream OVMF nor in XAPI's repos. Their README points to a fork of edk2 that isn't accessible anymore. Not sure what to make of it.

I can rename these to `xapi-`, but the patches will still technically be from xcp-ng.

jandryuk: I'd leave the file names and change the commit message.

recipes-core/varstored/varstored/0001-add-oxtdb-varstore.patch
tsirakisn: If I dropped SYS_ADMIN elsewhere, wouldn't unshare() and chroot() still fail? Meaning I'd have to get rid of them anyway.

jandryuk: I was thinking you do unshare() and chroot() first and then drop SYS_ADMIN afterwards.

tsirakisn: Per our conversation on Wire, opening a ticket for future evaluation. chroot() had a number of problems, most notably that we couldn't access the db from inside the chroot. unshare() had to be patched to remove a few of the namespaces anyway (as they didn't apply), and the others required varstored to have SYS_ADMIN capabilities, so it was removed altogether.

See ticket: https://openxt.atlassian.net/browse/OXT-1828

recipes-core/varstored/varstored/0001-add-oxtdb-varstore.patch
```diff
@@ -0,0 +1,891 @@
...
++{
++    return xapidb_serialize_variables(out, len, only_nv);
++}
++
++static bool
++write_all(const char *buf)
++{
++    bool check = false;
++    int var_num = 0;
++    check = oxtdb_exists("/vm/%s/varstored-efivar-blob/%d", oxtdb_arg_uuid, var_num);
```
jandryuk, line 466: var_num is always 0.  So you could just use "/vm/%s/varstored-efivar-blob/0", but there is no need for sub-keys, so "/vm/%s/varstored-efivar-blob" should suffice.

recipes-core/varstored/varstored/0001-add-oxtdb-varstore.patch
```diff
@@ -0,0 +1,891 @@
...
++        INFO("Backend init firstboot\n");
++        oxtdb_write("true","/vm/%s/varstored", oxtdb_arg_uuid);
++        free(check);
++        return BACKEND_INIT_FIRSTBOOT;
++    } else {
++        free(check);
++        return BACKEND_INIT_FAILURE;
++    }
++    free(check);
++    read_all(&encoded);
```
jandryuk, line 563: This seems like it could just be:
```
if (!oxtdb_exists("/vm/%s/varstored-efivar-blob/%d", oxtdb_arg_uuid, var_num)) {
    return BACKEND_INIT_FIRSTBOOT;
}

read_all(&encoded);
```
With the use of "/vm/%s/varstored" removed.  Unless I'm missing something.

recipes-extended/xen/files/libxl-add-ovmf-varstored.patch
```diff
@@ -0,0 +1,214 @@
...
++
++
+     libxl__store_libxl_entry(gc, guest_domid, "dm-version",
+         libxl_device_model_version_to_string(dm_config->b_info.device_model_version));
+     libxl__store_libxl_entry(gc, dm_domid, "stubdom-version",
+@@ -2736,6 +2791,7 @@ void libxl__spawn_local_dm(libxl__egc *e
+         libxl__xs_printf(gc, XBT_NULL,
+                          GCSPRINTF("%s/hvmloader/bios", path),
+                          "%s", libxl_bios_type_to_string(b_info->u.hvm.bios));
++
```
jandryuk, line 83: stray newline

recipes-core/packagegroups/packagegroup-xenclient-dom0.bb
```diff
@@ -110,6 +110,8 @@ RDEPENDS_${PN} = " \
     udev-extraconf-dom0 \
     vglass \
     disman \
+    varstored \
+    varstored-watch \
```
jandryuk, line 114: With proper RDEPENDS, which I think you have now, this patch https://github.com/OpenXT/xenclient-oe/pull/1434/commits/489d60df208661e91ac2198cad878240881e9b4e can be removed.

recipes-core/varstored/varstored/0001-add-oxtdb-varstore.patch
```diff
@@ -0,0 +1,891 @@
...
++        return false;
++    }
++
++    return true;
++}
++
++static bool
++oxtdb_save(void)
++{
++    return xapidb_save();
```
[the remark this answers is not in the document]
deepin125: I believe this is for migration, and it isn't used for sleep or anything that we really use in OXT.

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.te
tsirakisn: Can you elaborate on this one? Is this denial typically not blocking?

recipes-core/varstored/varstored/0001-add-oxtdb-varstore.patch
```diff
@@ -0,0 +1,891 @@
...
++// Write a value to the specified DB path.
++static void oxtdb_write(char * value, const char * path_format, ...) {
++
++    va_list args;
++    char * path = malloc(MAX_OXTDB_PATH_SIZE);
++
++    oxtdb_dbus_init();
++
++    va_start(args, path_format);
++    vsnprintf(path, MAX_OXTDB_PATH_SIZE, path_format, args);
```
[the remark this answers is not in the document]
tsirakisn: We really only added PATH_SIZE for protection against strings that were maliciously large. 512 is plenty for path sizes and should never be exceeded in the normal case.

recipes-core/varstored/varstored/0001-add-oxtdb-varstore.patch
```diff
@@ -0,0 +1,891 @@
...
++    }
++
++    return 0;
++}
++
++// Write a value to the specified DB path.
++static void oxtdb_write(char * value, const char * path_format, ...) {
++
++    va_list args;
++    char * path = malloc(MAX_OXTDB_PATH_SIZE);
```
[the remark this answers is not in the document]
tsirakisn: I'm gonna add a bunch of protections on these functions

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.te
jandryuk: tl:dr: I think you want to add `init_dontaudit_use_script_fds(varstored_t)`

I think an FD from the initscript leaks into xenmgr which then leaks to xl and to varstored/watch.  Maybe varstored/watch's use of stdio is part of the reason?  Anyway, we know varstored/watch ware self contained and don't rely on using inherited FDs, so we can just dontaudit them.  That way they are disallowed but don't log avc denials.

Cleaning up xenmgr leaking FDs to xl and xl leaking FDs to helpers is another option... but dontaudit is fine for now.

recipes-security/refpolicy/refpolicy-mcs-2.%/patches/policy.modules.system.xen.diff
```diff
@@ -420,6 +420,7 @@
 +        atapi_helper_signal(xend_t)
 +        audio_helper_domtrans(xend_t)
 +        audio_helper_signal(xend_t)
++        varstored_domtrans(xend_t)
```
jandryuk, line 423: Shouldn't this be varstored_watch_domtrans(xend_t) now?  Without this, you still were running varstored-watch as varstored_t, so you probably missed some new rules.

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.te
```diff
@@ -119,3 +134,41 @@ allow audio_helper_t self:shm create_shm_perms;
...
+########################################
+#
+# Local Policy: varstored/varstored-watch
+#
+
+# varstored-watch executing & monitoring varstored
+allow varstored_t initrc_t:fd use;
+allow varstored_watch_t bin_t:dir search;
+allow varstored_watch_t initrc_t:unix_stream_socket connectto;
+allow varstored_watch_t varstored_exec_t:file execute_no_trans;
```
jandryuk, line 147: Don't you want a trans here so that varstored runs as varstored_t?

I think you want `varstored_domtrans(varstored_watch_t)` here.

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.te
```diff
@@ -119,3 +134,41 @@ allow audio_helper_t self:shm create_shm_perms;
...
+dbus_system_bus_client(varstored_t)
+
+# pid creation
+allow varstored_t var_run_t:dir add_entry_dir_perms;
+allow varstored_t var_run_t:file manage_file_perms;
+allow varstored_t var_run_t:lnk_file read;
+
+# privcmd & xenstore access for checking VM configs
+dev_rw_xen(varstored_t)
+xen_stream_connect_xenstore(varstored_t)
```
jandryuk, line 163: You need `xen_stream_connect_xenstore(varstored_watch_t)` as well?

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.fc
```diff
@@ -21,3 +21,7 @@
 /usr/lib/xen/bin/atapi_pt_helper        --      gen_context(system_u:object_r:atapi_helper_exec_t,s0)
 /usr/lib/xen/bin/audio_helper(_start)?  --      gen_context(system_u:object_r:audio_helper_exec_t,s0)
 /var/lock/xen-atapi-pt-lock-[0-9]+_[0-9]+_[0-9]+_[0-9]+ -- gen_context(system_u:object_r:atapi_helper_lock_t,s0)
+
+/usr/sbin/varstored            --      gen_context(system_u:object_r:varstored_exec_t,s0)
+/usr/sbin/varstored-watch      --      gen_context(system_u:object_r:varstored_watch_exec_t,s0)
+/usr/share/varstored(/.*)?     --      gen_context(system_u:object_r:varstored_auth_t,s0)
```
jandryuk, line 27: `--` means it only applies to files, so directory /usr/share/varstored will remain usr_t.  And that is why below you only have `allow varstored_t varstored_auth_t:file read_file_perms;` without `:dir` .  So you probably want to drop `--` and then add `allow varstored_t varstored_auth_t:dir search_dir_perms` (though it might have to be list_dir_perms depending on what varstored_t does).

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.te
```diff
@@ -61,6 +61,21 @@ type audio_helper_tmpfs_t;
...
+domain_entry_file(varstored_t, varstored_exec_t);
+role system_r types varstored_t;
+
+type varstored_watch_t;
+type varstored_watch_exec_t;
+domain_type(varstored_watch_t);
+domain_entry_file(varstored_watch_t, varstored_watch_exec_t);
+role system_r types varstored_watch_t;
+
+type varstored_auth_t;
```
jandryuk, line 77: Add `files_type(varstored_auth_t)` here.

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.te
```diff
@@ -119,3 +134,41 @@ allow audio_helper_t self:shm create_shm_perms;
...
+
+########################################
+#
+# Local Policy: varstored/varstored-watch
+#
+
+# varstored-watch executing & monitoring varstored
+allow varstored_t initrc_t:fd use;
+allow varstored_watch_t bin_t:dir search;
+allow varstored_watch_t initrc_t:unix_stream_socket connectto;
```
jandryuk, line 146: Like `initrc_t:fd`, my gut tells me you can just do `dontaudit varstored_watch_t initrc_t:unix_stream_socket connectto;`

recipes-security/refpolicy/refpolicy-mcs-2.%/policy/modules/system/stubdom-helpers.fc
tsirakisn: Don't think I ever realized that. Always thought it was a separator for the refpolicy parser.

recipes-openxt/xctools/varstored-watch_git.bb
```diff
@@ -0,0 +1,24 @@
...
+RDEPENDS_${PN} = " \
+    varstored \
+"
+
+SRCREV = "${AUTOREV}"
+PV = "0+git${SRCPV}"
+
+S = "${WORKDIR}/git/varstored-watch"
+
+SRC_URI = " \
```
jandryuk, line 18: and SRC_URI should be replaced by `require xctools.inc` for the final submission.  meta-openxt-externalsrc needs an update as well, which I can take care of.

tsirakisn: Roger. I'll wait for your testing first in case there are other changes.

crogers1 (approved):
LGTM. Leaving up for a bit, if no more comments/objections will merge.

jandryuk:
I was expecting SB working from this PR :/

The problem with merging now, as I see it, is SB says it is enabled and working when it isn't.

How about we add `-DSECURE_BOOT_ENABLED=TRUE` to the ovmf build?  For a quick test then nothing can boot with SB enabled.  It's merged, but it's broken in a way that doesn't confuse users.  That ovmf change also seems to be required for any solution anyway.

deepin125:
While that will get Secure Boot working, using with the current certs will leave no guest able to actually boot.  The certs that appear to be created don't seem to allow even the Microsoft Boot loader to work.  I did get OVMF to build with the option, and can confirm that ISOs that aren't signed correctly fail to boot, however until I have some good working certs in place, I believe we need a little more time.  I am following an outline of how to do it, put forth by the xcp-ng community to get MS db certs setup and signed with the private KEK, and will report back when I'm done.

jandryuk:
Yes.  I'd rather nothing boots than everything boots and returns true for
secure boot.

jandryuk:
Ok, I think this works.

ovmf build with addition of `-DSECURE_BOOT_ENABLED=TRUE`

db.auth and KEK.auth generated by:
```
make create-auth
openssl req -newkey rsa:4096 -nodes -new -x509 -sha256 -days 3650         -subj "/CN=KEK Owner/"         -keyout KEK.key         -out KEK.crt
./create-auth -k KEK.key -c KEK.crt KEK KEK.auth KEK.crt
curl 'https://www.microsoft.com/pkiops/certs/MicCorUEFCA2011_2011-06-27.crt' --output MicCorUEFCA2011_2011-06-27.crt
curl 'https://www.microsoft.com/pkiops/certs/MicWinProPCA2011_2011-10-19.crt' --output MicWinProPCA2011_2011-10-19.crt
openssl x509 -in MicCorUEFCA2011_2011-06-27.crt   -text -inform DER -outform PEM -out MicCorUEFCA2011_2011-06-27.pem
openssl x509 -in MicWinProPCA2011_2011-10-19.crt   -text -inform DER -outform PEM -out MicWinProPCA2011_2011-10-19.pem
./create-auth -k KEK.key -c KEK.crt db db.auth MicWinProPCA2011_2011-10-19.pem MicCorUEFCA2011_2011-06-27.pem

# scp KEK.auth db.auth ...
```

Windows boots with Secure Boot and unsigned ISO doesn't boot.

jandryuk:
Ok, so we should include the Windows KEK then.  Thanks for finding that.

I don't know what the UIApp allows configuring.  Would it allow the person sitting at the keyboard to replace a VM's SB certs and potentially backdoor it?

PR #1435, opened 2021-09-29
rpc-proxy: Fixup rules for syncvm

jandryuk:
syncvm (and any users of domstore) need to be able to call
org.freedesktop.DBus.Hello.  Without it, the db-*-dom0 calls get blocked
by rpc-proxy.  Allow that when domstore-read access is given so things
just work.

syncvm's sync-client also calls org.freedesktop.DBus.GetOwnerName in
Domstore.__init__()
db_obj = bus.get_object(DOMSTORE_SERVICE, DOMSTORE_OBJECT)
allow that as well.


crogers1 (approved):
I've tested with these two in the sync template's rpc-firewall-rules, so this change is good to go.  Merging later.

PR #1437, opened 2021-09-29
tboot: Fix acmmatch segfault on microcode

jandryuk:
acmmatch is segfaulting on microcode_intel.bin and its date symlink.
They are not acms, but sanity check error handling has problems.

If get_acm_header() fails, acm->header is NULL when we jump to fail_map.
Change munmap() to use p, which is the valid pointer from mmap().  This
is the part that avoids the segfault.

Also if mmap() fails, go to fail_sanity.  There is nothing to munmap in
that case.


crogers1 (approved):
LGTM, merging soon.

PR #1438, opened 2021-09-30
blktap2: cancel_delayed_work

jandryuk:
I saw a dom0 kernel backtrace in console mode, I think while shutting down.  I put some hand copied info from that BUG into the commit message here: https://github.com/OpenXT/xenclient-oe/commit/c122ed5ddcab7ea2ef0b7800e67d16c5f13a5e15

I'm not sure what caused it, but the call stack seems to indicated that an already un-linked delayed_work struct might have been called a second time.  Since I didn't find any existing cancel_delayed_work call, this PR reorganizes code and adds one.

It was developed against 5.4, so it "Can't automatically merge" at this time.  At this time, I want to see if there is any feedback.  I'll put it in 5.10 locally for a little bit of testing.

### DIFF
--- a/recipes-kernel/linux/5.10/patches/blktap2.patch
+++ b/recipes-kernel/linux/5.10/patches/blktap2.patch
@@ -625,7 +625,7 @@ PATCHES
 +MODULE_LICENSE("Dual BSD/GPL");
 --- /dev/null
 +++ b/drivers/block/blktap/device.c
-@@ -0,0 +1,710 @@
+@@ -0,0 +1,715 @@
 +#include <linux/fs.h>
 +#include <linux/blkdev.h>
 +#include <linux/blk-mq.h>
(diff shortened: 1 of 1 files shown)


crogers1:
Took some time to walk through this in my head. I think it's good.  Since both threads have to drop into ```blktap_device_try_destroy```, one of them will be grabbing the mutex and calling ```clear_bit(BLKTAP_DEVICE, &tap->dev_inuse)```, which would make it impossible to schedule another delayed work.

jandryuk:
Thanks for reviewing @crogers1 

The commit message already has this:

> Have blktap_destroy_device cancel destroy_work, but re-scheduling from
another thread needs to be prevented.  Use BLKTAP_DEVICE bit to
determine whether or not destroy_work should be scheduled.  To avoid a
race, grab tapdev->lock in blktap_destroy_device and
blktap_destroy_work_common to ensure it is atomic.

Maybe I should change it to something like:

> We need blktap_destroy_device() to cancel destroy_work, but it has to guard against a racing thread scheduling destroy_work in blktap_destroy_work_common().  Use the BLKTAP_DEVICE bit to determine whether or not to schedule destroy_work.  Have blktap_destroy_device() hold tapdev->lock around clearing the bit and canceling destroy_work.  Have blktap_destroy_work_common() also grab the lock before checking BLKTAP_DEVICE and possibly scheduling.  The two operations occur atomically and avoid a race like:

```
thread A        thread B
test_bit
                clear_bit
                cancel_work
schedule_work
```

crogers1:
Yea the revised explanation illustrates the problem more clearly, though your first one I think was fine.  I was just trying to think about if there were any other cases that could possibly happen.

jandryuk:
Expanded commit message and rebased on 5.10.

crogers1:
Nothing else needed, slipped off my radar. Will merge today.

PR #1439, opened 2021-10-12
linux: txt.ko fix seq_file .next tpm_evtlog_next

jandryuk:
With the 5.10 uprev, Linux is warning:

kernel:[  130.724434] seq_file: buggy .next function tpm_evtlog_next
[txt] did not update position index

commit 3bfa7e141b0b "fs/seq_file.c: seq_read(): add info message about
buggy .next functions" added the warning, but it points a real the
problem: "Some ->next functions do not increment *pos when they return
NULL...  Note that such ->next functions are buggy and should be fixed"

That's whats happening in txt.ko.  tpm_evtlog_next needs to update pos
before returning NULL, so move the advancement forward to handle all
cases.


crogers1:
Nothing else needed, slipped off my radar. Will merge today.

PR #1441, opened 2021-10-12
Add monit to monitor vglass

jandryuk:
This adds monit to dom0.  It will monitor vglass, disman & ivcdaemon and restart them as necessary.

We want to run more frequently than the default monit interval of 30 seconds, so we use every 2 seconds.  There is also a 12 second delay at startup to give things a change to start.  I did this via dom0-cfg to avoid overriding monitrc which would require keeping in sync... Maybe this was a bad choice and monitrc should just be overridden.

refpolicy needs tweaks to work with volatiles and actually start the daemons.

Goes with https://gitlab.com/vglass/meta-vglass/-/merge_requests/20

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -1,12 +1,7 @@
 # We have a conf and classes directory, append to BBPATH
 BBPATH .= ":${LAYERDIR}"
 
-# meta-selinux appends populate-volatiles.sh with commands that will not work
-# under a read-only rootfs:
-# - liberaly restorecon /var/lib
-# - sed checkroot.sh
 BBMASK = " \
-    meta-selinux/recipes-core/initscripts/ \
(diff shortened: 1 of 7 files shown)


crogers1:
Will merge when corresponding PR is merged on vglass

jandryuk:
Re-worked the dom0 configuration into monitrc via an override and moved the volatiles portion into volatiles-cfg.

crogers1:
@jandryuk can you resolve the conflicts and then I'll merge?

jandryuk:
I rebased and force-pushed here.  The conflicts were minor from the refpolicy uprev.  This PR and https://github.com/OpenXT/xenclient-oe/pull/1442 will conflict since they both add patches to refpolicy.  Let's just do this one first and then I'll re-rebase the usbvm one.

crogers1:
Merged

PR #1444, opened 2021-11-17
varstored: Create fresh {KEK,db}.list

jandryuk:
The append redirects may get called repeatedly if re-builds are
triggered without cleaning.  This will created larger and larger db.auth
files.  At runtime, varstored will fail to load them with:
"Auth file '/usr/share/varstored/db.auth' is too large: 57829"

The guest sits spinning in OVMF, and the display doesn't initialize,
showing "Guest has not initialized the display (yet)."

Clear the files first with '>' before appending more.


jandryuk:
I think this one is straight forward and fixes a bug.  Anything else needed?

crogers1:
Nope nothing else needed.  Will merge today.

crogers1 (approved):

PR #1445, opened 2021-11-18
ifplugd and refpolicy fixes

jandryuk:
commit 0de2f2e "[OXT-1643] Removing legacy ifplugd recipe" removed
the ifplugd package because busybox can provide it.  However, busybox
needs to be built with support for it, which we do not do.  This turns it back on and makes it usable for syncvm.

It also uses ifplugd for dom0 networking and removes network-daemon launching udhcpc itself.

There are some refpolicy fixes.  I started off fixing denials and then used ifplugd to avoid udhcpc denials from inherited FDs.

[OXT-1643]: https://openxt.atlassian.net/browse/OXT-1643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jandryuk:
Goes with https://github.com/OpenXT/network/pull/25

jandryuk:
This has gone stale.  I can rebase in new PRs.  I think ifplugd getting restored for syncvm should be fine and uncontroversial.

I think using ifplugd for dom0 eth0 if better than network-daemon launching udhcpc directly.  Maybe ifplugd running should be tied into xenmgr enable-dom0-networking kinda like enable-ssh?  i.e. ifplugd only runs when /config/system/dom0-networking-disabled isn't present.  Are others okay with that?

https://github.com/OpenXT/xenclient-oe/pull/1480 has some of the other pieces.

PR #1446, opened 2022-03-20
Uprev to Linux 5.15.29

crogers1:
This PR uprevs OpenXT to Linux 5.15.29.  It also modifies the installer defconfig to enable extra framebuffer support.  During testing on an HP Z2 G5 EFI boot, the tty never loaded to step through the installer.  After adding the defconfig options, it became visible on that platform.

With 5.15.29 we will have (once all the options are turned on, for each specific platforms) support for 11th gen hardware.  5.15.x is also the current LTS Linux version for the foreseeable future.

Finally, we get the latest XSA and CVEs that have been backported to the LTS release.


recipes-kernel/linux/5.15/patches/blktap2.patch
```diff
@@ -642,6 +643,19 @@ PATCHES
 +
 +#define dev_to_blktap(_dev) container_of(_dev, struct blktap, device)
 +
++static struct block_device *bdget_disk(struct gendisk *disk, int partno)
++{
```
jandryuk, line 647: bdget_disk was removed in 5.15, so you had to re-add it?

recipes-kernel/linux/5.15/patches/blktap2.patch
```diff
@@ -1057,26 +1054,20 @@ PATCHES
 +	del_gendisk(gd);
 +	gd->private_data = NULL;
 +
-+	cleanup_queue(gd->queue);
++	blk_cleanup_disk(gd);
```
jandryuk, line 1057: With the cleanup_queue removal, blk_cleanup_disk takes the place of blk_cleanup_queue?  Another 5.15 change?

recipes-kernel/linux/5.15/patches/blktap2.patch
```diff
@@ -1057,26 +1054,20 @@ PATCHES
...
++
++	blk_mq_free_tag_set(&tapdev->tag_set);
 +
-+	put_disk(gd);
 +	tapdev->gd = NULL;
 +
-+	mutex_lock(&tapdev->lock);
```
jandryuk, line 1065: Looks like here and elsewhere, the changes from https://github.com/OpenXT/xenclient-oe/pull/1438 were dropped.

(I have since repro-ed the crash https://github.com/OpenXT/xenclient-oe/pull/1438 tried to fix.  I think https://github.com/OpenXT/xenclient-oe/pull/1438 is still valid, but the trigger is different.)

crogers1: Yea that's a mistake on my part, I will amend the patch to keep #1438 included.

recipes-kernel/linux/5.15/patches/blktap2.patch
crogers1: Good call.

recipes-kernel/linux/5.15/patches/blktap2.patch
crogers1: Yea, ```blk_cleanup_disk``` does the ```blk_cleanup_queue``` and ```put_disk```(also removed from line 1062) in the same function.

recipes-kernel/linux/5.15/patches/blktap2.patch
```diff
@@ -1223,7 +1184,19 @@ PATCHES
 +	if (blktap_device_validate_info(tap, info))
 +		return -EINVAL;
 +
-+	gd = alloc_disk(1);
++	memset(&tapdev->tag_set, 0, sizeof(tapdev->tag_set));
++        tapdev->tag_set.ops = &blktap_mq_ops;
```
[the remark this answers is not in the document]
crogers1: Will fix indentation.

recipes-kernel/linux/5.15/patches/blktap2.patch
crogers1: The latest push should have your #1438 content included.

recipes-kernel/linux/5.15/patches/blktap2.patch
```diff
@@ -290,7 +291,7 @@ PATCHES
 +
 +	mutex_lock(&blktap_lock);
 +
-+	for (minor = 0; minor < blktap_max_minor; minor++)
++	for (minor = 1; minor < blktap_max_minor; minor++)
```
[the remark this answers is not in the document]
crogers1: Seems this change isn't needed, was a hold over from earlier testing.  We're changing it back from 1 to 0 downstream as well.

recipes-kernel/linux/5.15/patches/blktap2.patch
```diff
@@ -2917,7 +2863,7 @@ PATCHES
 +	mutex_lock(&blktap_lock);
 +
 +	ret = 0;
-+	for (i = 0; i < blktap_max_minor; i++) {
++	for (i = 1; i < blktap_max_minor; i++) {
```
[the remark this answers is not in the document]
crogers1: Same as above. not including this change, it's fine from 0.

crogers1:
Latest push should address all the comments.  I'm going to kick off another custom build to test everything out.

recipes-kernel/linux/5.15/patches/blktap2.patch
```diff
@@ -1223,7 +1189,19 @@ PATCHES
...
++	tapdev->tag_set.ops = &blktap_mq_ops;
++	tapdev->tag_set.nr_hw_queues = 1;
++	tapdev->tag_set.queue_depth = BLKTAP_RING_SIZE / 2;
++	tapdev->tag_set.numa_node = NUMA_NO_NODE;
++	tapdev->tag_set.flags = BLK_MQ_F_SHOULD_MERGE | BLK_MQ_F_BLOCKING;
++	tapdev->tag_set.cmd_size = 0;
++	tapdev->tag_set.driver_data = tap;
++
++	if (blk_mq_alloc_tag_set(&tapdev->tag_set))
++		return NULL;
```
jandryuk, line 1202: This should return -ENOMEM

jandryuk:
I also see a compiler error about this format string and t.tv_sec:
```
-+                            "op:%x nr_pages:%02d time:%lu.%09lu\n",
++                            "op:%x nr_pages:%02d time:%llu.%09lu\n",
```

crogers1:
The custom build here seemed to finish fine without any build errors: https://openxt.ainfosec.com/buildbot/#/builders/22/builds/64

crogers1:
Addressed the format string by using ```lld``` and changed NULL return to -ENOMEM. Checked log.do_compile and didn't see any warnings for blktap.

recipes-kernel/linux/5.15/patches/blktap2.patch
```diff
@@ -1039,14 +1044,6 @@ PATCHES
 +
 +	bdev = bdget_disk(gd, 0);
 +
-+	err = !mutex_trylock(&bdev->bd_mutex);
-+	if (err) {
-+		/* NB. avoid a deadlock. the last opener syncs the
-+		 * bdev holding bd_mutex. */
-+		err = -EBUSY;
-+		goto out_nolock;
-+	}
```
jandryuk, line 1048: Sorry, I didn't notice this earlier - what is the rationale that it is safe to remove this locking?

Looks like linux commit a8698707a1835be3abd12a3b28079a80999f8dee
"block: move bd_mutex to struct gendisk" moved it up a level.  Looking at that commit, del_gendisk tries to take the lock, and we call that below, so maybe that is part of why is needs to be dropped?

I think the point of the lock is to serialize a race between the block device cleaning up and a second process opening it.  I also thought it should be held to look at the bdev data like bd_openers.

Maybe not for now, but commit 05d69d950d9d84218fc9beafd02dea1f6a70e09e
"xen-blkfront: sanitize the removal state machine" cleaned up blkfront for its removal case.  Following that may be good idea for blktap2 as well.

recipes-kernel/linux/5.15/patches/blktap2.patch
```diff
@@ -1039,14 +1044,6 @@ PATCHES
 +
 +	bdev = bdget_disk(gd, 0);
```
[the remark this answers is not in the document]
crogers1: I'm adding some defensive checks around this and other bdget_disk() calls.

recipes-kernel/linux/5.15/patches/blktap2.patch
crogers1: I think you're right with the bd_mutex, though perhaps the original code here was piggybacking on it and it was really needing a completely separate lock.  We don't want to double-lock, so replacing this call with something like ```bdev->bd_disk->open_mutex``` isn't going to work.  I have to imagine this is a pointer to the same lock that's in the ```gendisk``` object ```gd->open_mutex``` that ```del_gendisk``` grabs, like you mentioned.

If we still want to serialize access to the ```blktap_device_destroy``` function, which is effectively what ```bd_mutex``` was being used for originally, I agree, we need to replace it with something else.  Let me see what I can do

crogers1: Actually, can we just move your ```mutex_lock(&tapdev->lock)``` up?

crogers1: I'd rather not touch 05d69d950d9d84218fc9beafd02dea1f6a70e09e at the moment.  I will try the trylock for open_mutex and see what happens.

jandryuk: That sounds reasonable.

crogers1: using the trylock on open_mutex led to deadlock.  moving the tapdev->lock further up seems to work much better.

crogers1:
updated to address latest comments.

recipes-kernel/linux/5.15/patches/blktap2.patch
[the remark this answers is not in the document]
crogers1: This should be good now, I added a ```if  (!bdev)``` case down on line 1049

recipes-kernel/linux/5.15/patches/blktap2.patch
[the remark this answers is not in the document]
crogers1: Dropped this back down to zero, resolving this thread.

recipes-kernel/linux/5.15/patches/blktap2.patch
jandryuk: Locally, I tried `mutex_trylock(&bdev->bd_disk->open_mutex);` and can confirm it deadlocks :(

crogers1: :( indeed

jandryuk:
I saw this.  It's only a warning, but it is definitely a bug:
```
drivers/usb/xen-usbback/interface.c: In function 'usbif_map':
drivers/usb/xen-usbback/interface.c:112:8: warning: passing argument 1 of 'bind_interdomain_evtchn_to_irqhandler_lateeoi' makes pointer from integer without a cast [-Wint-conversion]
  112 |   usbif->domid, evtchn, usbif_be_int, 0, "usbif-backend", usbif);
      |   ~~~~~^~~~~~~ 
      |        |
      |        domid_t {aka short unsigned int}
drivers/usb/xen-usbback/interface.c:36:
include/xen/events.h:42:73: note: expected 'struct xenbus_device *' but argument is of type 'domid_t' {aka 'short unsigned int'}
   42 | int bind_interdomain_evtchn_to_irqhandler_lateeoi(struct xenbus_device *dev,

      |                                                   ~~~~~~~~~~~~~~~~~~~~~~^~~
```
The solution seems to be to just replace usbif->domid with usbif->be->dev.  I've compiled this, but not tried the usb functionality.

crogers1:
I'll take a look.

crogers1:
Updated usbback patch for the bind_interdomain_evtchn call. Tested vusb using a linux guest, was able to mount and access contents on a passed through USB drive.

jandryuk:
I'm building locally to give it a spin.

The use of `tapdev->lock` instead of `open_mutex` is slightly concerning since it seems like `open_mutex` should be held to access bd_openers.  But `tapdev->lock` will serialize access checks, so there is some protection that way.  Since `open_mutex` resulted in a deadlock and hung boot, that isn't a viable path anyway.

Hopefully I can find some time to remove its use like in Linux commit 05d69d950d9d84218fc9beafd02dea1f6a70e09e
"xen-blkfront: sanitize the removal state machine".

crogers1:
Buildbot build with the latest changes: https://openxt.ainfosec.com/OpenXT/auto/master/master-custom/65/

jandryuk:
Will merge soon.  Thanks, @crogers1 !

PR #1447, opened 2022-03-21
Blktap2 fixes

jandryuk:
https://github.com/OpenXT/xenclient-oe/commit/072a39ded084eea085ae3658312828af715ac3c9 didn't full solve the crash it was trying to.  Turns out you can repro by SIGKILL-ing a tapdisk process.  The destroy work calls cleanup_queue() -> blk_cleanup_queues(), and it crashes when blk_cleanup_queues() runs a second time.  

(blktap_device_destroy seems wonky.  It leaves the device open if there are still any openers, or if it is locked.  Locked implying that it is the final opener.  This means closing the final opener doesn't close the device and relies on destroy_work or an IOCTL to remove.  I feel like this should be done better, but I'm not sure.  Maybe destroy_work should just go away and leave cleanup tied to ioctl?)

There are also two small cleanups to remove a duplicate queue pointer and correct the units for the destroy_work timeout.

### DIFF
--- a/recipes-kernel/linux/5.15/patches/blktap2.patch
+++ b/recipes-kernel/linux/5.15/patches/blktap2.patch
@@ -1089,12 +1089,12 @@ PATCHES
 +blktap_device_fail_queue(struct blktap *tap)
 +{
 +	struct blktap_device *tapdev = &tap->device;
-+	struct request_queue *q = tapdev->gd->queue;
 +
 +	mutex_lock(&tapdev->lock);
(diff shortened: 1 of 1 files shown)


crogers1:
Looks like there's a merge conflict here @jandryuk

jandryuk:
Thanks, @crogers1 .  I force pushed the 5.15 version which I've been running for a while.

crogers1 (approved):
LGTM

PR #1448, opened 2022-03-22
RFC: Display PCRs when unseal fails

jandryuk:
This may be a bad idea.

When unseal fails for a PCR mismatch, it takes some work to figure out what failed.  And you can only do that after either unlocking or booting from alternate media.  We don't store the non-calculated (current) PCRs at sealing time, so we can't know how they compare.

This has the machine print the mis-matched PCRs... but you can't trust the output since it may be compromised!  In that light, this could be a bad idea in production.  For development, it is handy to see what is unexpected for a failed unseal.

The output looks like:
PXL_20220322_120859327a

Ideally we have a TPM Quote displayed as a QR code so we can verify the current PCRs...

Even if the unseal message isn't changed, I think the recording of all PCRs is worthwhile.  Another option is to add a third option to `<Continue>` and `<Shutdown>` and only display the diff when `<Untrustworthy Info>` is selected.

Thoughts?

### DIFF
--- a/recipes-core/images/xenclient-dom0-image.bb
+++ b/recipes-core/images/xenclient-dom0-image.bb
@@ -118,3 +118,9 @@ rw_config_partition() {
 }
 ROOTFS_POSTPROCESS_COMMAND += "rw_config_partition; "
 ROOTFS_POSTPROCESS_COMMAND += "start_tty_on_hvc0;"
+
+# Toggle flag to show PCRs during measured launch failure
+debug_pcrs() {
+	sed -i '/^DEBUG_DUMP_PCRS=/s/0/1/' "${IMAGE_ROOTFS}/sbin/init.root-ro"
+}
+ROOTFS_POSTPROCESS_COMMAND += '${@bb.utils.contains("IMAGE_FEATURES", "debug-tweaks", "debug_pcrs; ", "",d)}'
--- a/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
+++ b/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/seal-system
@@ -78,6 +78,7 @@ tpm_ver="$(get_tpm_version)"
 operation="seal"
 config_key="/config/keys/config.key"
 sealed_key="/boot/system/tpm/config.tss"
+forward_seal_list=/boot/system/tpm/forward_pcr.lst
 config_pcrs="/config/config.pcrs"
 good_pcrs="/config/good.pcrs"
 bad_pcrs="/boot/system/tpm/bad.pcrs"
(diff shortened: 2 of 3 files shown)


crogers1:
I think we made some quality of life changes to good.pcrs and bad.pcrs. Garrett found the behavior you're currently describing to also be extremely annoying for getting good data during dev and testing.  Let me make sure I can post the relevant diffs.

crogers1:
Yea, so basically we added some code in init-root.ro to do two things:
-  during the /config unlock, always write to the pcrs.good file when ```encrypted_unlock``` succeeds.
- if ```encrypted_unlock``` ever fails, get the current pcr values and write them to pcrs.bad.

At this point whenever you do fail measurement, you have the last known good pcr values (from the last time it unlocked succesfully, not just a seal/forward seal operation), and you also have the current boot pcr values that failed (prior to the PCR 15 cap).

jandryuk:
The encrypted_unlock success case sounds good.

What do you think about storing the current PCRs during forward seal?  It helps if the non-calculated ones change.  They end up on the unencrypted boot partition, because updatemgr_t can't write to /config.

crogers1:
> What do you think about storing the current PCRs during forward seal?
I think you should have these already because ```encrypted_unlock``` has to succeed before you can forward seal, so you'll have the current PCRs if we if implement the first bullet in my comment above.  I guess they would be in /config though as opposed to /boot.

jandryuk:
Yes, it would mostly be there.  The corner case which I am seeing with an finicky HP is PCR-1 likes to change.  It seems like you can forward seal to PCR-1, but then it will be different after reboot.  Or maybe the difference is between a reboot and cold boot.  If the PCR isn't captured at seal-time, you don't have a record of the value.

crogers1:
PR for writing PCRS_GOOD on encrypted unlock: https://github.com/OpenXT/xenclient-oe/pull/1454

Even with the finicky PCR-1s, it is still much easier to diagnose measurement problems with the above change.  I propose we close this PR in favor of the other one.

jandryuk:
Agreed that https://github.com/OpenXT/xenclient-oe/pull/1448/commits/8f869b38c3adc651623e8874f7ff2b84d52a3547 showing the PCR mismatch can be dropped.

I think https://github.com/OpenXT/xenclient-oe/pull/1454 is good to go in.  That will handle good.pcrs, but what do you think about https://github.com/OpenXT/xenclient-oe/pull/1448/commits/f819460dd03a355def2d0348fd5a61fdd8b6a1af (which depends on https://github.com/OpenXT/xenclient-oe/pull/1448/commits/c48e128571feb1c2852e1b701bcf4c766df471d7 IIRC)?  That records at the time of forward seal, so it can be used to diagnose PCR mismatches.

jandryuk:
I had this extra commit sitting around, so I pushed it out.   It makes the PCR debugging introduced here controlled by debug-tweaks at build time.

It still requires the PCR values written to /boot/system.

If this isn't going anywhere, I'll close the PR.  I think there is value in storing the PCRs at time of sealing.  Maybe the should be written to /config?

crogers1:
I'm fine with enabling it via IMAGE_FEATURES at build time.  Storing PCRs at the time of sealing I would feel more comfortable taking if that lived in /config alongside good.pcrs.  That limits the ability to debug the PCRs at the time of unseal failure, but at least the seal-time PCR values would be recorded, encrypted, and accessible to an admin with the recovery key for debugging.

jandryuk:
Closing.  Maybe I'll revisit if I come up with a better approach.

PR #1450, opened 2022-05-17
linux: usbback: Remove stdarg.h include

jandryuk:
We have warnings like this for va_start, va_arg and va_copy:
include/linux/stdarg.h:6: warning: "va_start" redefined
    6 | #define va_start(v, l) __builtin_va_start(v, l)

gcc/x86_64-oe-linux/9.3.0/include/stdarg.h:47: note: this is the location of the previous definition
   47 | #define va_start(v,l) __builtin_va_start(v,l)

usbback should not be including the compiler's stdarg, since it should
be using Linux's definitions.  Drop the compiler inclusion.


crogers1 (approved):
LGTM

PR #1451, opened 2022-05-17
tboot: pcr-calc: Error without evtlog

jandryuk:
It is an error to be unable to find the tpm 2.0 sysfs event log, so
return failure.  Without this, seal-system keeps running with empty pcr
17, 18 & 19, which are invalid forward seal values.

The sysfs event log can be missing when the txt_op hypercall number
mismatches between Xen and Linux.


crogers1 (approved):
LGTM

PR #1452, opened 2022-06-07
Drop many linux patches

jandryuk:
The first commit explains how xenbus-move-otherend-watches-onrelocate.patch can crash a host, which I have observed twice.  https://github.com/OpenXT/xenclient-oe/commit/ee17b7e3882a42c1e9f22345b5060f43b338de7f

The patch is unsafe.  While maybe it can be fixed, I think it is no longer relevant for the current OpenXT with xl and we should just drop it.

I then continued dropping many patches which I think are no longer needed.  It's hard to know since so few are actually documented.

Two that give me a little pause are:
[acpi-video-delay-init.patch](https://github.com/OpenXT/xenclient-oe/commit/52dd19913089a70d1594b51bf6af1a2f9d13dd0e) I think dropping this is fine for regular vglass use.  The part I wonder about is maybe GPU passthrough benefits from this?  However VFIO doesn't do anything similar, so it seems unlikely to be needed.

and:
[disable-csum-xennet.patch](https://github.com/OpenXT/xenclient-oe/commit/5b83307d88c1c74d9329a5ba4f17aa12c988afcb) vif checksum offload should work since Xen has been deployed a lot.  I did find this bug which mentions an incompatibility between iptables and the xen checksum offloading: https://bugzilla.redhat.com/show_bug.cgi?id=474191 .  But again it's not clean why it was really added.  Since NDVM does run NAT, maybe it would be worth keeping it in there.  But then it could be replaced by just disabling tx offload on all the interfaces.

Anyway, I think we can and should drop most of these to just ease our maintenance going forward.


crogers1:
- xenbus-move-otherend-watches-on-relocate.patch :heavy_check_mark:
   Agreed here, good find. I can't recall if I've seen that BUG specifically but I definitely agree with your logic.
- netfront-support-backend-relocate.patch :heavy_check_mark:
   Agreed here, drop it
- netback-skip-frontend-wait-during-shutdown.patch :heavy_check_mark: 
   Mostly agree here.  I suspect this case may have been for a PV linux vm with an OE built kernel that needed networking. If this guest ever hung, the backend NDVM might never shutdown and thus dom0 would never shutdown.  The only usecase today where I see this would be the SyncVM, though a hung guest might be better handled through the toolstack with a timeout rather than xen-netfront forcibly closing the xen-netback backend device (of which other guests may share it).
- xenkbd-tablet-resolution.patch :heavy_check_mark: 
   Agreed here, drop it
- acpi-video-delay-init.patch :heavy_check_mark:
   Agreed here, I don't think there's anything GPU PT related with this one.  It looks more related to onboard Intel GFX during boot time. I'd be hard pressed to justify supporting Sandybridge (released 2011) and earlier on dunfell+ OpenXT. 
- realmem-mmap.patch :heavy_check_mark:
   Agreed here, this definitely seems surfman related, other surfman specific patches Eric dropped modified vma mapping attributes so it stands to reason this is similar.
- intel-amt-support.patch :heavy_check_mark:
   Agreed here, no need to support intel AMT at all
- tpm-log-divid.patch :heavy_check_mark: 
   Agreed here, just for debugging and not really needed
- tpm-tis-force-ioremap.patch :heavy_check_mark: 
   Agreed here, no longer needed with 64-bit dom0
- skb-forward-copy-bridge-param.patch :heavy_check_mark:
   Agreed here, unused
- disable-csum-xennet.patch :heavy_check_mark: 
   Agreed here, removing this should boost in-guest download speeds for windows.  This was something we were testing downstream.

jandryuk:
Thanks for taking a look.

> linux: Drop netback-skip-frontend-wait-during-shutdown.patch 

I just double checked, and xenbus_frontend_dev_shutdown(), which is used without the patch, has a 5 second timeout.  So the wait is reasonably timebound.

Linux circa 5.10 re-worked xenbus code.  If you look at openxt commit 966aa96c894a22b83a3a655795b31c515746b738,  can see that the patch moved from applying to drivers/xen/xenbus/xenbus_probe_backend.c to drivers/xen/xenbus/xenbus_probe_frontend.c.  That is because linux commit https://github.com/torvalds/linux/commit/c534374ecf044384ddd24474b91fd0d31e720464 moved the function.  So it originally applied to the backend.  However that linux commit messages also says it didn't work for backends:
> xenbus_dev_shutdown() is seemingly intended to cause clean shutdown of PV frontends when a guest is rebooted. Indeed the function waits for a conpletion which is only set by a call to xenbus_frontend_closed().

Maybe at some point it used to work, but it doesn't seem to work now.

Hmmm, if the backend doesn't have a shutdown method, then it doesn't have a wait at all?

Anyway, lets drop it.

crogers1:
Yep sounds good to me.  I'll kick off a custom build and we should do some minimal smoke testing

crogers1:
Build [here](https://openxt.ainfosec.com/buildbot/#/builders/22/builds/66)

jandryuk:
FWIW, I installed build 66 and ran an existing Windows .vhd without problem.  Only minimally tested.

crogers1:
I think that's sufficient.  I'll merge this later today if there aren't any other objections

crogers1 (approved):

PR #1453, opened 2022-06-07
ndvm: Use RDEPEND with network-slave

jandryuk:
network-slave execs brctl, carrier-detect, iptables, and NetworkManager,
so it has direct runtime dependencies on them.  Add them to RDEPENDS.

With that, carrier-detect is not a build depends, so drop it from
DEPENDS.

Since they are set as RDEPENDS, drop the runtime deps from the ndvm
image recipe.


crogers1 (approved):
LGTM.  Should help with build times a bit

PR #1454, opened 2022-08-18
[mle] Write PCRS_GOOD on successful unlock

crogers1:
Over time after upgrades and forward seal operations, PCRS_GOOD
  can become outdated. When you do encounter a measurement failure,
  it doesn't always provide an admin/developer accurate PCR values
  for debugging what PCR measurement failed.

  This commit writes the latest known-good PCR values on each successful
  unlock of the encrypted partition when booting.

Signed-off-by: Chris Rogers <crogers122@gmail.com>


jandryuk (approved):
There's a part of my that doesn't like re-writing things that haven't changed.  Maybe checking the firstboot flag, which is set on upgrade?  But this is so much simpler, and it won't miss any case.

jandryuk:
Will merge later today.

PR #1455, opened 2022-08-19
dbus: Remove Make-the-default-DBus-reply-timeout-configurable.pat

jandryuk:
This patch needs a configure option to change the timeout value.  Since
commit 203c0d2bf8f8 "dbus: Move to upstream OE" (2015!), the configure
option has not been supplied.  Drop the unused patch.


crogers1:
Missed this one too.  Also looks good.  Merging soon

PR #1458, opened 2022-09-13
installer: Generate root-owned control.tar.bz2

jandryuk:
The build user & group id are leaking into the control.tar.bz2 tarball. Specify the owner and group as root to avoid this.

It doesn't have a practical impact in normal OpenXT use since the installer runs as root.


crogers1 (approved):
LGTM

PR #1461, opened 2022-09-22
Xen 4.16 uprev

jandryuk:
Uprev to 4.16.  It first uprevs to 4.14 where some stuff is fixed up, and then jumps to 4.16.

Depends on:
https://github.com/OpenXT/manager/pull/196
https://github.com/OpenXT/libxenbackend/pull/9
https://github.com/OpenXT/xsm-policy/pull/40
~~https://gitlab.com/vglass/glass/-/merge_requests/38~~
https://gitlab.com/vglass/glass/-/merge_requests/39


jandryuk:
It would be good for someone to double check the grant table kconfig changes.  Maybe also the PCI device cleanup.

I am seeing an issue with some machines where "shutdown" reboots the computer.  I'm guessing it's a Xen crash that then reboots.  It doesn't trigger consistently.  It seems to happen when the UI is running, but not when shutting down from console mode.

jandryuk:
Preliminary testing shows the shutdown -> reboot issue may have been resolved by https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-5.15.y&id=5359524251d27d5fb50e425bac0108a134ed682c

crogers1:
Ack.  @tsirakisn and I will start reviewing these.

recipes-extended/xen/xen-common.inc
```diff
@@ -47,7 +47,6 @@ SRC_URI_append = " \
     file://libxl-pci-passthrough-fixes.patch \
     file://libxl-vwif-support.patch \
     file://libxl-atapi-pt.patch \
-    file://libxl-iso-hotswap.patch \
```
crogers1, line 79: Did you have the async friendly way already mapped out or is that something we'll have to implement?

recipes-extended/xen/files/libxl-const-stubdom-is-linux.patch
```diff
@@ -0,0 +1,11 @@
+--- a/tools/libxl/libxl_internal.h
```
crogers1: This patch needs some kind of header, obviously the description is trivial but specifying the removal plan is important.  I would imagine this is very unstreamable.

recipes-extended/xen/files/blktap3-build-integration.patch
```diff
@@ -1,35 +1,22 @@
...
 --- a/tools/Makefile
 +++ b/tools/Makefile
-@@ -20,7 +20,7 @@ SUBDIRS-y += console
+@@ -19,6 +19,7 @@ SUBDIRS-$(CONFIG_X86) += firmware
+ SUBDIRS-y += console
  SUBDIRS-y += xenmon
  SUBDIRS-y += xenstat
- SUBDIRS-$(CONFIG_Linux) += memshr 
--SUBDIRS-$(CONFIG_BLKTAP2) += blktap2
 +SUBDIRS-$(CONFIG_BLKTAP2) += blktap3
```
crogers1: Is this stuff still needed? blktap3 is a separate repo, you've already rm'd the xen-tools-blktap recipes, does this and the ```XEN_BLKTAP2``` need to make it into Xen's tools Makefile?

recipes-extended/xen/files/libxl-const-stubdom-is-linux.patch
jandryuk: I already upstreamed it:
commit e7cfcdc6719d586eb7cdb62d40275a7d17fe6760
Author: Jason Andryuk <jandryuk@gmail.com>
Date:   Wed Mar 30 14:17:22 2022 -0400

    libxl: constify libxl__stubdomain_is_linux

recipes-extended/xen/files/blktap3-build-integration.patch
jandryuk: It is removed in:
commit d3189b5328c3c9362d90058d8efcf59c14825f93
Author: Jason Andryuk <jandryuk@gmail.com>
Date:   Fri Nov 19 14:16:04 2021 -0500

    Build blktap3 outside xen

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
```diff
@@ -0,0 +1,551 @@
...
++    fi
++
++    claim_lock "block"
++    if find_device; then
++        dev="/dev/xen/blktap-2/tapdev$minor"
++        check_device_sharing "$dev" "$mode"
++    else
++        TAPDISK3_CRYPTO_KEYDIR=$( db-read /xenmgr/platform-crypto-key-dirs )
++        : ${TAPDISK3_CRYPTO_KEYDIR:=/config/platform-crypto-keys}
++        USER_KEYDIRS=$( xec-vm -n uivm get crypto-key-dirs )
```
crogers1: This was what we were avoiding by calling tapEnv in xenmgr, is this a separate case where the block script is being called outside of xenmgr?

jandryuk: It gets removed here:
commit c8547eed43cd1764b87701b1b47dfe8a6fb327c5
Author: Jason Andryuk <jandryuk@gmail.com>
Date:   Fri Nov 19 12:02:39 2021 -0500

    xen: convert-block-tap-for-shared.patch

    count_sharing needs some more updating.

The commit message doesn't mention it :/

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
```diff
@@ -0,0 +1,551 @@
+From 164436bdff5620e605268ec82860728014dab7cb Mon Sep 17 00:00:00 2001
```
crogers1, line 1: Reminder for me to come back and finish reviewing this patch

recipes-extended/xen/xen-common.inc
crogers1: ack ty.

recipes-extended/xen/files/blktap3-build-integration.patch
crogers1: ack thanks. I'm going through these commit-by-commit.

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
crogers1: Gotcha.

recipes-extended/xen/xen-tools_git.bbappend
```diff
@@ -14,6 +14,8 @@ B = "${S}"
 
 DEFAULT_PREFERENCE = "1"
 
+EXTRA_OECONF+="--disable-golang"
```
crogers1, line 16: What's the motivation for this change?

recipes-extended/xen/files/libxl-const-stubdom-is-linux.patch
jandryuk: Sorry, it is needed - it will only be in 4.17.  I updated to use the full `git format-patch` output as the patch and mention 4.17 in the commit message.

recipes-extended/xen/xen-tools_git.bbappend
jandryuk: I forget - I'll investigate and try enabling golang.  I wouldn't have added this without a reason, so there may have been a build issue.  The flipside is since OpenXT doesn't use Go, I may have disabled it to cut down on build time.

jandryuk: In my test, configure didn't find "go" and successfully compiled.  I will remove this.

recipes-extended/xen/files/libxl-const-stubdom-is-linux.patch
crogers1: :+1: will expect to drop it when we go to 4.17.

recipes-extended/xen/xen-tools_git.bbappend
crogers1: Interesting, and thanks.  We've been using golang downstream and could see upstreaming some of that work; not sure if it matters for xen or any of the xen tools specifically though.

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
jandryuk: 0001-convert-block-tap-for-shared.patch is introduced and fixed up in later commits.  I think they can and should all be squashed together, especially if we have xenmgr set TAPDISK3_CRYPTO_KEYDIR at the time of PR merging

recipes-extended/xen/xen-common.inc
```diff
@@ -72,6 +72,7 @@ SRC_URI_append = " \
     file://libxl-add-ovmf-varstored.patch \
     file://libxl-const-stubdom-is-linux.patch \
     file://libxl-disable-dom0-qemu.patch \
+    file://libxl-xen-disable-dom0-qemu.patch \
```
tsirakisn, line 68: Not necessarily an issue, but in the interest of keeping the patch queue trimmed, do we want to combine these two qemu patches? They provide the same fix essentially.

recipes-extended/xen/files/libxl-openxt-helpers.patch
```diff
@@ -195,22 +197,3 @@ PATCHES
...
--                int stubdom_domid = dcs->dmss.pvqemu.guest_domid;
--                libxl__qmp_initializations(gc, stubdom_domid, d_config);
--            }
-+            /* OpenXT: for stubdom-enabled guests, qmp_helper creates the same
-+             * unix socket than QEMU would for non-stubdom guests.
-+             * Therefore there's no need to dissociate the two cases */
-+            libxl__qmp_initializations(gc, domid, d_config);
-         }
-     }
-
```
tsirakisn, line 216: There are a lot more references to the qmp helper in this patch. Are we leaving that logic intact intentionally?

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
```diff
@@ -0,0 +1,551 @@
...
++        USER_KEYDIRS=$( xec-vm -n uivm get crypto-key-dirs )
++        if [ -n "$USER_KEYDIRS" ]; then
++            TAPDISK3_CRYPTO_KEYDIR="$USER_KEYDIRS,$TAPDISK3_CRYPTO_KEYDIR"
++        fi
++        CRYPTO_USER=$( xec-vm -n uivm get crypto-user )
++        if [ -n "$CRYPTO_USER" ]; then
++            TAPDISK3_CRYPTO_KEYDIR="/config/sec/s-$CRYPTO_USER,$TAPDISK3_CRYPTO_KEYDIR"
++        fi
++        export TAPDISK3_CRYPTO_KEYDIR
++        dev=$( ( eval "exec $_lockfd>&-" ; tap-ctl create $tap_opts -a $target ) )
```
crogers1, line 506: Doesn't
```
eval "exec $_lockfd>&-"
```
close the FD in the current process? ```release_lock``` is still ok with a closed FD?

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
crogers1: ok I'm done with this, resolving

recipes-extended/xen/files/defconfig
```diff
@@ -29,19 +42,28 @@ CONFIG_GRANT_TABLE=y
 # CONFIG_GRANT_TABLE_EXOTIC is not set
 CONFIG_HAS_ALTERNATIVE=y
 CONFIG_HAS_EX_TABLE=y
+CONFIG_HAS_FAST_MULTIPLY=y
 # CONFIG_MEM_ACCESS_ALWAYS_ON is not set
 # CONFIG_MEM_ACCESS is not set
-# CONFIG_HAS_MEM_PAGING is not set
-# CONFIG_HAS_MEM_SHARING is not set
+CONFIG_HAS_MEM_PAGING=y
```
crogers1, line 48: This one stands out to me since it seems to apply to HVM guests, what do we need it for? Or better question, what is it useful for?

recipes-extended/xen/files/defconfig
```diff
@@ -29,19 +42,28 @@ CONFIG_GRANT_TABLE=y
...
 CONFIG_NEEDS_LIBELF=y
+
+#
+# Speculative hardening
+#
+CONFIG_SPECULATIVE_HARDEN_ARRAY=y
+CONFIG_SPECULATIVE_HARDEN_BRANCH=y
+# end of Speculative hardening
+
+CONFIG_HYPFS=y
```
crogers1, line 63: If we compile with this on now, do you have another commit that is mounting/using the hypervisor file system?  Is this something we maybe shouldn't provide from a security perspective?

recipes-extended/xen/files/defconfig
```diff
@@ -84,24 +114,26 @@ CONFIG_HAS_PCI=y
...
 CONFIG_VGA=y
 CONFIG_HAS_VPCI=y
+# end of Device Drivers
 
-#
-# Deprecated Functionality
-#
-# CONFIG_PV_LDT_PAGING is not set
-CONFIG_DEFCONFIG_LIST="arch/x86/configs/x86_64_defconfig"
+CONFIG_EXPERT=y
```
crogers1, line 119: I didn't see where EXPERT was a requirement in the Kconfig for argo.  Is it required due to some config dependencies or am I just missing where it says it outright?

recipes-extended/xen/files/libxl-disable-dom0-qemu.patch
```diff
@@ -0,0 +1,27 @@
+libxl: do not start qemu in dom0 just for extra consoles
```
crogers1, line 1: Is this completely blocking qemu from running in dom0 (ie. HVM without stubdomain) or just the console support?  If it's the latter, the patch name should be something like ```libxl-disable-qemu-console-helpers-in-dom0.patch```.

recipes-core/initrdscripts/initramfs-stubdomain/init.sh
```diff
@@ -77,8 +77,18 @@ export INTEL_DBUS=1
...
-/usr/bin/qemu-system-i386 ${dmargs}
+
+/usr/bin/qemu-system-i386 ${dmargs} -chardev socket,server,nowait,path=/tmp/qemu.qmp,id=m2 -mon chardev=m2,mode=control &
+
+device_model="device-model/$target"
+
+echo "Starting vchan-socket-proxy"
+vchan-socket-proxy -m server 0 $device_model/qmp-vchan /tmp/qemu.qmp &
+
+wait
```
crogers1: In the current implementation, QEMU is a blocking process. On HVM domain shutdown, somehow it exits cleanly and then the stubdom ```poweroff -f```'s.  With the ```wait``` and the backgrounded QEMU process, will the stubdom still shutdown in the same fashion?

recipes-extended/xen/files/libxl-stubdom-allow-msi-enable.patch
```diff
@@ -0,0 +1,20 @@
+--- a/tools/libxl/libxl_pci.c
++++ b/tools/libxl/libxl_pci.c
+@@ -1417,6 +1417,17 @@ static void pci_add_dm_done(libxl__egc *
+             rc = ERROR_FAIL;
+             goto out;
+         }
++    } else if (libxl_is_stubdom(ctx, domid, NULL)) {
```
crogers1, line 7: This is in support of the HVM ndvm usecase?

recipes-extended/qemu-dm/qemu-dm.inc
```diff
@@ -41,6 +41,7 @@ SRC_URI += "file://compile-time-stubdom-flag.patch \
             file://block-remove-unused-block-format-support.patch \
             file://net-Remove-unused-network-options.patch \
             file://round-pci-region-sizes.patch \
+            file://0001-vnc-stubs-Allow-vnc-none.patch \
```
crogers1, line 44: Noting here, when we uprev OE  to kirkstone, we will be able to drop this patch as it's included in v6.2.0 (current qemu version in kirkstone)

recipes-extended/xen/files/defconfig
jandryuk: The "*HAS*" ones like CONFIG_HAS_MEM_PAGING just means the architecture has support for the feature - not that is it used/enabled.

Oh, defconfig might need a refresh for 4.16 since I don't see it down in the 4.16 build directory.

recipes-extended/xen/files/defconfig
jandryuk: That's a good question.  While it has FS in the name, it's not a mountable FS.  But it is a way to expose some stuff from the hypervisor to ~dom0.  XSM restricts it to dom0_type only, so I think it's okay.

    Remove XEN_SYSCTL_set_parameter
    
    xen commit 02e9a9cf2095 "xen: remove XEN_SYSCTL_set_parameter support"
    says "The functionality of XEN_SYSCTL_set_parameter is available via
    hypfs now, so it can be removed."  Mirror the change.

So I think it may be needed in place of the above?

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
```diff
@@ -0,0 +1,551 @@
...
+     if [ -z "$pid" ] || [ -z "$minor" ]; then
+-        fatal "cannot find required parameters"
++        return 1
++    fi
++
++    return 0
++}
++
++##
++# count_sharing device mode
```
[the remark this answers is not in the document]
jandryuk: You are right - Thanks

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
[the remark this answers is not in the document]
jandryuk: The commit message introducing it has:
    xen: Patch block-tap script
    
    This brings in a patched block-tap script.  It provides the device
    sharing checks previously done in libxl.
    
    Calling tap-ctl passes an open file descriptor to tapback which messes
    up locking.  That is why we have `eval "exec $_lockfd>&-` to close the
    flock fd.
    
    We also play games for the keydir.  This tries to mimic what's in xenmgr
    already... I guess we could let xenmgr set those variables before
    execing xl...

I can expand the patch description with something like:
Make block-tap usable with OpenXT

Xen has a block-tap script to handle vhds, but it needs some changes to be usable by OpenXT.

It needs to handle device sharing checks which were previously done in libxl.  This is partially accomplished by moving and reusing functions from Xen's block script to common code.

The call to tap-ctl needs to change to prevent leaking the open _lockfd file descriptor into the tapdisk daemon.  The `eval "exec $_lockfd>&-` call closes the fd to prevent that leak.  Such a leak blocks any other block-tap script from running thereby blocking any other VM from starting.
 
It relies on TAPDISK3_CRYPTO_KEYDIR being set in the environment by xenmgr to allow tapdisk to unlock encrypted VHDs.

recipes-extended/xen/xen-common.inc
jandryuk: The two patches come from Qubes - I've wondered why they are separate there.  Since they come from elsewhere, I think we should just leave them as-is.

recipes-extended/xen/files/libxl-openxt-helpers.patch
jandryuk: I think you are right that qmp-helper reference can be removed since we are using vchan-socket-proxy.  If qmp-helper wants to be resurrected, it can be done by specifying via `configure --stubdom-qmp-proxy` and making it compatible argument wise.

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
crogers1: Thanks for the extra explanation.

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
```diff
@@ -428,6 +428,8 @@ Subject: [PATCH] convert block-tap for shared
 +#                requested mode; or
 +#    'ok':       the device may be used.
 +#
++#    '$n':       Number of instances using the tapdev
```
crogers1, line 428: I just realized that commit ```c8547eed43cd1764b87701b1b47dfe8a6fb327c5``` " xen: convert-block-tap-for-shared.patch " makes changes to the block scripts, I think it could probably be squashed into the original commit?

recipes-extended/xen/files/0001-tools-xl-Add-stubdomain_cmdline-option-to-xl.cfg.patch
```diff
@@ -0,0 +1,87 @@
+From ce233b10319dfb77ab5e0102202190a5911ae8f3 Mon Sep 17 00:00:00 2001
```
crogers1: I assume this is also something that's going to be in 4.17 on release?

recipes-extended/xen/files/defconfig
jandryuk: I think ARGO used to be behind EXPERT.  Now it is behind UNSUPPORTED.

recipes-extended/xen/files/libxl-disable-dom0-qemu.patch
jandryuk: The patch is only disabling QEMU from providing the extra consoles.  Non-stubdom will still launch svirt-interpose/qemu.

The patch name comes from Qubes.  I would not change it, but I can if you definitely want that.

recipes-core/initrdscripts/initramfs-stubdomain/init.sh
jandryuk: The poweroff hasn't seemed necessary in my testing.  I think QEMU exists on QMP command.  Later commits add the console, which has the init 1 shell waiting.  Things seem to work fine and there aren't any messages about the stubdom crashing.

recipes-extended/xen/files/libxl-stubdom-allow-msi-enable.patch
jandryuk: In the 256b62a3224423f6293163dde7e1d6f7eb3801fd commit message, I wrote
    Add this Qubes patch to allow stubdoms to enable MSI.  Without it,
    e1000e were reporting that they could not enable MSI and falling back to
    regular IRQs.

Qubes upstreamed some other changes to how IRQs are handled.  I think this piece hasn't gone in since it's for linux stubdom pci passthrough which isn't upstream.

jandryuk: d7bcc139d6d61c750f234db278ad9a28bcfadda0 on this branch.

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
jandryuk: I think all convert-block-tap-for-shared.patch changes should be squashed together since they are fixes and we agreed on the xenmgr env var approach for TAPDISK3_CRYPTO_KEYDIR.

recipes-extended/xen/files/0001-tools-xl-Add-stubdomain_cmdline-option-to-xl.cfg.patch
jandryuk: It was needed for 4.14 and goes away for 4.16.

recipes-extended/xen/xen-common.inc
tsirakisn: Ack

recipes-extended/xen/files/libxl-fix-flr.patch
```diff
@@ -125,66 +124,75 @@ PATCHES
...
-@@ -1501,10 +1500,6 @@ skip1:
-         fclose(f);
+@@ -1596,7 +1595,8 @@ void libxl__device_pci_add(libxl__egc *e
+         goto out;
      }
- out:
+ 
+-    libxl__device_pci_reset(gc, pcidev->domain, pcidev->bus, pcidev->dev, pcidev->func);
++    /* Ugh this needs to go.  All resets must happen before any are assigned. */
++    /* libxl__device_pci_reset(gc, pcidev->domain, pcidev->bus, pcidev->dev, pcidev->func); */
```
crogers1: Based on your commit message, under what circumstances would a device be hot-added to a domain in openxt?  Currently all devices/functions are reset and assigned prior to domain start, and then reset after domain death.  I can't think of a scenario where one function gets reset independently of a 2nd function, while the guest is running.

recipes-core/initrdscripts/initramfs-stubdomain/init.sh
```diff
@@ -82,7 +82,12 @@ dmargs=$( xenstore-read $(xenstore-list -p "$vm_uuid/image/dm-argv" | sort ))
 echo "target $target vm_uuid $vm_uuid"
 echo "Invoking qemu with dmargs       = ${dmargs}"
 
+# $dm_args and $kernel are separated with \n to allow for spaces in arguments.
+IFS=$'\n'
```
crogers1, line 192:
```
OIFS=$IFS
IFS='\n'

... do whatever ...
IFS=$OIFS
```
```unset``` will clear IFS in the current scope, but IFS comes in with an initial value.  In the past I've saved the incoming IFS and then restored it when I was done handling the data to avoid clobbering it.

File: recipes-core/initrdscripts/initramfs-stubdomain/init.sh [hunk 22]
```diff
@@ -97,6 +97,8 @@ device_model="device-model/$target"
 echo "Starting vchan-socket-proxy"
 vchan-socket-proxy -m server 0 $device_model/qmp-vchan /tmp/qemu.qmp &
 
-wait
-
-poweroff -f
+while true; do
```
tsirakisn, line 212: +1 for a stubdom shell. I've done this successfully in dev environments by adding:

`/sbin/getty 115200 hvc0 -n -l /bin/sh &`

right before qemu is called. This did not require cttyhack or setsid and can be utilized with a simple `xl console <dm_name>`. Is there a reason behind the ctty hack? I'm not too familiar with it.

recipes-kernel/linux/5.15/patches/xen-txt-add-xen-txt-eventlog-module.patch
```diff
@@ -277,7 +277,7 @@ Subject: [PATCH] xen-txt: add xen txt eventlog module
  #define __HYPERVISOR_xc_reserved_op       39 /* reserved for XenClient */
  #define __HYPERVISOR_xenpmu_op            40
  #define __HYPERVISOR_dm_op                41
-+#define __HYPERVISOR_txt_op               42
++#define __HYPERVISOR_txt_op               43
```
tsirakisn, line 280: Is there no patch fuzz associated with this? Shouldn't there be a line for the new hypfs?

recipes-extended/xen/files/allow-mwait-cstate.patch
```diff
@@ -1,98 +0,0 @@
-################################################################################
```
crogers1, line 1: So I agree PV kernels can't actually execute MWAIT or MONITOR directly, I agree.  But doesn't this patch just advertise to PV guests whether the host supports it?  Not that it is claiming PV guests can execute the instruction?

I have to imagine at some point, someone wanted to be able to make a decision about deep C-states from inside a PV guest, perhaps by querying the cpuid when otherwise such a query would not say whether the host supported MONITOR/MWAIT.  Whether or not we have anything making use of such a query is up for debate.  I'd have to grep and get lucky I think.

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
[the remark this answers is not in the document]
tsirakisn: Appreciate the explanation! Since you already wrote it you might as well add it.

recipes-extended/xen/files/defconfig
crogers1: _EXPERT_ to _UNSUPPORTED_ eh? I can only imagine reading between those lines :joy:

recipes-extended/xen/files/libxl-fix-flr.patch
jandryuk: I might have messed this up?

For an HVM domain with a stubdom, devices are PV PCI hotplugged into the running PV stubdom, and then QMP assigned to QEMU before booting the HVM.

I thought the issue was that with "thorough reset" potentially resetting a slot, reset cannot be done while any devices are assigned.  The patch description mentioned "delayed until the domain has fully been destroy."

I thought the patch was to move all resets to only the shutdown side.  That way, any devices are reset when they move from the guest back to the system, and the potential reset issues are avoided.  Was the reset during startup supposed to remain?

If the reset during startup remains, would a thorough reset be a problem?
2 devices at startup:
Reset 1: Ok
Reset 2: Potentially bad

The devices wouldn't be in use by the HVM guest (though they would by in stubdom via pcifront), so this is probably okay?  The real guest isn't using them yet, and the stubdom shouldn't use them on its own.

Seems like the best flow would be:
1 reset all PCI devices
2 assign PCI devices
3 run - shutdown vm
4 deassign PCI
5 reset all PCI

I also wonder about cross-VM affects of the PCI thorough reset if the slot reset happens.  Is that a concern?

(libxl-fix-flr.patch gets modified a few times in subsequent patches, but not in a way relevant to this change.)

recipes-core/initrdscripts/initramfs-stubdomain/init.sh
jandryuk: Sounds good.  I'll make the change.

recipes-core/initrdscripts/initramfs-stubdomain/init.sh
jandryuk: cttyhack is take from Qubes.  The qubes way of doing things made it upstream to Xen, so aligning the approach seems preferable.  Also, this way will respawn a new shell if the first is closed.

From https://github.com/brgl/busybox/blob/master/shell/cttyhack.c
cttyhack avoids the "can't access tty; job control turned off" error message, which typically appears when one tries to use a shell with stdin/stdout on /dev/console.

recipes-kernel/linux/5.15/patches/xen-txt-add-xen-txt-eventlog-module.patch
jandryuk: No, log.do_patch doesn't mention any.  This is patching a header in linux source, and that has not been updated with hypfs_op yet.

recipes-extended/xen/files/allow-mwait-cstate.patch
jandryuk: I agree that at some point, some PV kernel must have used this.  But we don't have a patch in our tree to do something based on it.  Any other distro is upstream-based these days, and per the comment shouldn't be using this.

Dom0 is still allowed, and that is the one that can do some C-state uploading.

So it just seems like something we should drop.  If only someone documented the *why* for this patch originally...

recipes-core/initrdscripts/initramfs-stubdomain/init.sh
crogers1: Ok, sounds good.

recipes-core/initrdscripts/initramfs-stubdomain/init.sh
tsirakisn: Okay, if it made it upstream then it's good enough for me

recipes-extended/xen/files/libxl-disable-json-updates.patch
```diff
@@ -1,88 +0,0 @@
-################################################################################
```
crogers1, line 1: I think the point of this patch was to not let libxl modify the ```/tmp/xenmgr-xl-<uuid>``` conf file that gets passed as the ```xl create``` arg.  Xenmgr is the authoritative source for what should be in those files; every time a domain is started, xenmgr will overwrite whatever is there with fresh values read from the db.

recipes-extended/xen/files/prune-vga-acpi-dev.patch
```diff
@@ -1,59 +0,0 @@
-################################################################################
```
crogers1, line 1: To be clear, removing this patch has no effect on the state of S3 for Windows 10?

recipes-extended/xen/files/libxl-syslog.patch
```diff
@@ -287,12 +287,15 @@ PATCHES
...
-@@ -419,6 +424,8 @@ int main(int argc, char **argv)
+@@ -419,6 +424,11 @@ int main(int argc, char **argv)
          xtl_flags |= XTL_STDIOSTREAM_SHOW_DATE | XTL_STDIOSTREAM_SHOW_PID;
      logger = xtl_createlogger_stdiostream(stderr, minmsglevel, xtl_flags);
      if (!logger) exit(EXIT_FAILURE);
 +    /* OpenXT: we use syslog, we could probably comment out the above. */
 +    openlog("xl", LOG_CONS, LOG_USER);
++    /* OpenXT: xl writes to stderr, which can be a dangling pipe.  We don't
++     * want that terminating the process, so ignore it. */
++    signal(SIGPIPE, SIG_IGN);
```
crogers1, line 298: Couldn't we just redirect stdout and stderr to /dev/null if everything is making it to syslog by default?

recipes-extended/xen/files/libxl-disable-json-updates.patch
jandryuk: The `/tmp/xenmgr-xl-$uuid` files are different from the json in `/var/lib/xen/userdata-d.$domid.$uuid.json`.  xenmgr still overwites the xl.cfg files and that is fine.  I'm not sure about libxl re-writing the xl.cfg files.  There is some talk of it in the xl man page, but I don't know about it happening in practice.  Xenmgr would still overwrite anyway.

The json data is something libxl uses for its dynamic state tracking.  I am not totally sure what it is for, but there is some logic for comparing it with xenstore when outputting devices.  That leads to some weirdness with `xl pci-list` where it doesn't list any devices for the HVM, but it does list them for the PV stubdom.  The Xenstore entries are only present for the PV stubdom.

Anyway, the patch is affecting the json side and not the xl.cfg.

Xen commit e6afc7063afe06f4609fb2d18fde28f3b54e995e seems to be the switch to the .json stuff.  Looks like that is basically about changes taking affect on reboot.  If we pursue my changes to switch out xenmgr reboot handling to shutdown & start, then this libxl stuff is irrelevant.

Back to this patch.  The description implies failure because the json files are not present.  But they are present now.  I guess I can try switching networks at run time to see that hasn't broken.  Otherwise, I haven't seen issue with this patch dropped.

jandryuk: For a running VM, the UI network tab doesn't allow adding or removing networks.  So I'm doubly confused with what this is supposed to be addressing.

recipes-extended/xen/files/prune-vga-acpi-dev.patch
jandryuk: S3 isn't allowed on Windows 10 with or without this patch - the lack of S3 support for the graphics card is listed.

Run cmd as Administrator and run `powercfg /a`:
```
Standby (S3)
    An internal system component has disabled this standby state.
            Graphics
```

recipes-extended/xen/files/libxl-argo-xswait.patch
```diff
@@ -85,22 +85,37 @@ PATCHES
...
 +    sdss->multidev.callback = spawn_stub_launch_dm;
 +    libxl__add_disks(egc, ao, dm_domid, dm_config, &sdss->multidev);
-+    libxl__multidev_prepared(egc, &sdss->multidev, 0);
++    libxl__multidev_prepared(egc, &sdss->multidev, rc);
++
++    return;
++
++ out:
++    libxl__xswait_stop(gc, xswait);
++    qmp_proxy_spawn_outcome(egc, sdss, rc);
```
tsirakisn, line 118: The control flow of these changes confuses me a little. The early returns in the error cases mean that the `libxl__*` calls are never executed, which might mean we don't clean up properly in the case of an error? Also, the new `out:` seems more like an `err:` now as the success case never hits that block.

recipes-core/initrdscripts/initramfs-stubdomain/init.sh
```diff
@@ -82,10 +82,122 @@ dmargs=$( xenstore-read $(xenstore-list -p "$vm_uuid/image/dm-argv" | sort ))
...
+    ip link set "$vif" up
+}
+
+(
+set +x
+for arg in $dmargs; do
+if [ $(echo "$arg" | grep -cim1 "ifname=") -eq 1 ]; then
+    (
+    # XXX Remove after debugging
+    set -x
```
tsirakisn, line 175: I assume this `set -x` can be removed

jandryuk: :) Yes, I'll remove that.

recipes-core/initrdscripts/initramfs-stubdomain/init.sh
```diff
@@ -82,10 +82,122 @@ dmargs=$( xenstore-read $(xenstore-list -p "$vm_uuid/image/dm-argv" | sort ))
...
+    fi
+
+    if [ ! -e "/sys/class/net/$eth" ] ; then
+        echo "qemu_ifup: PV interface $eth does not exist."
+
+        return 1
+    fi
+
+    if [ -e "/sys/class/net/$br" ] ; then
+        echo "qemu_ifup: Bridge $br already exists"
```
tsirakisn, line 140: Should there be an early return here (or similar)? If the bridge exists, we still attempt to `addbr` down below. Looking at the original script, there are a few operations we opt-out of when the bridge is already present.

recipes-extended/xen/files/libxl-argo-xswait.patch
jandryuk: Yes, it's a little confusing - took me a second to remember.  argo_xswait_cb() is called immediately on xenstore registration, so the `if (!p)` case is not an error - we need to keep waiting to give it a second chance.  For `if (strcmp(p, "1")) {`, I chose to not fail on invalid values and give it more time to come get set.  i.e. it'll timeout if it doesn't change to the expected value.  It shouldn't happen, but it isn't necessarily fatal which is why it just returns.  Since the xswait hasn't been stopped, return means just keep waiting.  This is what happens programming with libxl's async paradigm.

jandryuk: libxl has 436 "out:" vs. 36 "err:", so I think I was mainly following surrounding code.  I can change it if you want.

recipes-core/initrdscripts/initramfs-stubdomain/init.sh
jandryuk: I'll add a `return 1` for that case.  It's one of those "should never happen" things.

recipes-extended/xen/files/libxl-argo-xswait.patch
jandryuk: How about changing the log messages to:
"Unable to read argo-firewall-ready - keep waiting"
and
"Unexpected argo-firewall-ready value\"%s\" - keep waiting"
?

recipes-extended/xen/files/defconfig
crogers1: I think it might just be
```
CONFIG_MEM_PAGING
```
now.

recipes-extended/xen/files/libxl-stubdom-allow-msi-enable.patch
crogers1: We don't see e1000e complaining today though, even in the HVM ndvm case.  Looking at a recent openxt build
```
00:04.0 Ethernet controller: Intel Corporation Ethernet Connection (11) I219-LM
	Subsystem: Hewlett-Packard Company Device 8753
	Flags: bus master, fast devsel, latency 64, IRQ 56
	Memory at f1000000 (32-bit, non-prefetchable) [size=128K]
	Capabilities: [c8] Power Management version 3
	Capabilities: [d0] MSI: Enable+ Count=1/1 Maskable- 64bit+
	Kernel driver in use: e1000e
	Kernel modules: e1000e
```
MSI: Enable+, and nothing in dmesg concerning MSI.

recipes-extended/xen/files/libxl-argo-xswait.patch
tsirakisn: I think those log additions would make things a lot clearer, +1. The label name is less important, but looking at it some more I'm not sure if that block is needed anymore. We only jump to it once, and that's during the proper error case. Since it's just two lines you might as well just call them in that `if` block directly.

recipes-extended/xen/files/shim-support-for-shim-lock-measure.patch
```diff
@@ -81,38 +77,68 @@ Signed-off-by: Tamas K Lengyel <lengyelt@ainfosec.com>
...
-         /*
+@@ -1387,10 +1431,23 @@ efi_start(EFI_HANDLE ImageHandle, EFI_SY
+      * verify it.
+      */
+     if ( kernel.ptr &&
+-         !EFI_ERROR(efi_bs->LocateProtocol(&shim_lock_guid, NULL,
+-                                           (void **)&shim_lock)) &&
+-         (status = shim_lock->Verify(kernel.ptr, kernel.size)) != EFI_SUCCESS )
+-        PrintErrMesg(L"Dom0 kernel image could not be verified", status);
++         shim_lock && !kernel_already_measured) {
```
tsirakisn, line 122: Out of curiosity, why duplicate here if we know that our measurement has already happened? In what case would we be hitting this second block without a measured kernel?

tsirakisn:
Finally got through this whole thing. Awesome work, this looks like a ton of effort. Appreciate the time.

recipes-extended/xen/files/defconfig
jandryuk: The .config down in the build directory has:
`# CONFIG_MEM_PAGING is not set`

I'll add a commit refreshing the defconfig for 4.16.

recipes-extended/xen/files/libxl-argo-xswait.patch
jandryuk: Yes, you are correct that the block could be moved.  However, I was trying to follow the libxl "style" using the out label.  With all the async callbacks in libxl, they try to funnel exiting to the next step at the end of a callback.  The other side of that is the "keep waiting" cases return return in the middle as this particular callback will be called again for success or failure.

So I'd leave it with the "libxl" style, but I can change it if you really want.

tsirakisn: Ack. If that's the paradigm then it is what it is.

Btw - for some reason I don't have a `Resolve` button under my comments. If you have the means you can feel free to resolve whatever comment chains are done at this point.

recipes-extended/xen/files/shim-support-for-shim-lock-measure.patch
jandryuk: I think the re-ordering in https://github.com/OpenXT/xenclient-oe/pull/1461/commits/ad0555604b2f01b4ee3753869ea5b52d91332e95 is unnecessary.  I confused myself that the movement of the kernel measuring would change the PCR calculation.  However, moving the kernel is okay since it is measured into PCR 4 while the others are measured into PCR 8.  An HP laptop failed to unseal which made me think this was the problem, but that laptop has some other issue with PCR4.

I'm going to test out removing `kernel_already_measured` and just having the single kernel location.

recipes-extended/xen/files/libxl-stubdom-allow-msi-enable.patch
jandryuk: I may have been running with pci_permissive=0 - I've been testing that off and on.  With pci_permissive=1, the guest can enable MSI-X in the PCI config space, but that is denied with pci_permissive=0.  This is the Qubes commit adding the patch: https://github.com/QubesOS/qubes-vmm-xen/commit/9774d93bb602052c1eb16f1bc9639e3edb16ed06

Supporting pci_permissive=0 is something I'd like to introduce as at least an option.  Should I drop this for the time being?  For xenmgr, my quick patch just makes a per-VM pci_permissive setting.  Per-PCI device seems like it would be better, but implementing that is more complicated.

recipes-extended/xen/files/libxl-fix-flr.patch
crogers1: So I was digging through this stuff today and I may have been mistaken about resets.  Devices _are_ reset before they're assigned to a domain but the libxl code that reset is attached to I believe is for like, ```xl pci-attach```, not the normal domain creation with pci devices specified in the pci list in the domain's xl.conf.

That being said, I'm not 100% sure how those pci devices in the domain xl.conf list get added to the domain during creation.  I will keep looking, because this would give us a hard answer as to whether or not devices are reset before domain start

That flow sounds correct to me.  We avoid cross-VM problems in the downstream by enforcing that all functions on the device get passed through to the same VM.  No assigning the GPU video function to one Windows guest and GPU Audio function to a different Windows guest.  That way all functions on the slot (device) are released after domain shutdown and if FLR fails, slot (device) reset should succeed.

recipes-extended/xen/files/allow-mwait-cstate.patch
crogers1: Agreed, missing a lot of context from the original commit pre-Github.  We can do regression testing and keep an eye on this, but yea for the most part I doubt anything is using this capability.

jandryuk: I'll expand the commit message with: "At some point, some PV kernel must have used this, but we don't have a patch in our tree to do something based on it. Any other distro is upstream-based these days, and per the comment shouldn't be using this.  Dom0 is still allowed, and that is the one that can do some C-state uploading."

recipes-extended/xen/files/libxl-fix-flr.patch
jandryuk: Domain create calls the same functions as xl pci-attach.  Domain create does:
```
libxl__add_pcis()
  loop
    libxl__device_pci_add()
```
whereas `xl pci-attach` calls:
```
libxl_device_pci_add()
  libxl__device_pci_add()
```

Upstream Xen does a reset in libxl__device_pci_add() and pci_remove_detached().

Anyway, I think you are right that libxl__device_pci_reset() should be retained in libxl__device_pci_add().  During startup, given "The real guest isn't using them yet, and the stubdom shouldn't use them on its own.", I think it's fine to just leave the reset and not try to re-work the reset flow.

recipes-extended/qemu-dm/qemu-dm/qemu-dm-wrapper
```diff
@@ -41,4 +41,25 @@ export QEMU_ALSA_VOL_CTRL=vm-${ALSA_VMID}
...
+# We have to manually setup a pipe equilavent with a fifo and run the exec
+# directly from the main script.
+
+# Dummy commands to create a pipeline.
+sleep 9999 | sleep 9999 &
+# save the process ids
+PID2=$!
+PID1=$(jobs -p %+)
+# hijack the pipe's file descriptors using procfs
+exec 3</proc/$PID2/fd/0 4>/proc/$PID1/fd/1
```
crogers1, line 53: Per our side discussions and your testing, I would like to see the qemu-dm-wrapper here trap on the SIGHUP from libxl and exit the qemu and logger processes, instead of the anonymous pipe approach.

tsirakisn: Same here fwiw. Chris and I had talked about this offline as well.

jandryuk: The trap/SIGHUP implementation is more straightforward, but this pipe stealing is kinda amazing...

recipes-extended/xen/files/libxl-stubdom-allow-msi-enable.patch
crogers1: It's not directly needed to support 4.16, and I also think it deserves having its own PR(s).

recipes-extended/xen/files/libxl-fix-flr.patch
crogers1: > Domain create calls the same functions as xl pci-attach. Domain create does:
Must have completely glossed over where those get called.  Interesting.

+1 for leaving flow as is for now.

recipes-extended/xen/files/libxl-disable-json-updates.patch
crogers1: > For a running VM, the UI network tab doesn't allow adding or removing networks. So I'm doubly confused with what this is supposed to be addressing.

Interesting.  Possibly it was done through the commandline instead of the UI with ```xl network-attach```?  Also another case could be when the ndvm gets restarted underneath a running VM there is some vif switching due to the domid of the ndvm changing.

recipes-extended/xen/files/prune-vga-acpi-dev.patch
crogers1: Ack.  We will look into the S3 support issues after xen 4.16 gets merged.

recipes-extended/xen/files/libxl-syslog.patch
crogers1: Ack.

crogers1:
A few last convos to finish up and then quick final review when you submit the changes.  This was a huge amount of work and I just wanted to give you some major thanks and kudos for spearheading this effort @jandryuk!

recipes-extended/qemu-dm/qemu-dm/qemu-dm-wrapper
crogers1: It is a pretty cool trick, honestly didn't know you could do that sort of things with pipes.

recipes-extended/xen/files/libxl-fix-flr.patch
jandryuk: "Leave flow as is" meaning we keep the reset in libxl__device_pci_add(), yes?  That is what I am testing locally.

recipes-extended/qemu-dm/qemu-dm/qemu-dm-wrapper
jandryuk: trap/SIGHUP seems to be working, so I'm switching to that.

recipes-extended/xen/files/libxl-fix-flr.patch
crogers1: Yea, to be clear:
- Leave the reset in pci add
- Also reset after the domain is destroyed

recipes-extended/xen/files/shim-support-for-shim-lock-measure.patch
jandryuk: Forward seal & unseal failed when I removed kernel_already_measured.  I tested on a Dell is typically good for sealing & unsealing.

tsirakisn: I guess I'm still confused on the two cases then. The implication with that variable is that sometimes the kernel will already be measured at this point and sometimes it won't be. In what case will you have one vs the other? And why isn't it universal across all platforms?

jandryuk: Well, I was trying to make it work for x86 or ARM.  Maybe I shouldn't bother with that and just stick to x86.

shim-support-for-shim-lock-measure.patch doesn't have a problem with the order of PCR4 & 8 extends, but tboot-measure-and-launch-from-xen-efi.patch extends tboot into PCR4.  So that is what throws off PCR4 xen.efi->tboot vs tboot->xen.efi

I can re-spin without kernel_already_measured and break the ARM devicetree path to just have the 1) xen.efi 2) tboot extend order.  Does that sound good?

tsirakisn: I would think ARM is a moot consideration seeing as how we don't support it for the larger usecase. Personally I would say to focus on x86 to avoid confusion. Curious to hear from @crogers1 though.

jandryuk: xen commit 048de2c47075d6106136bee110d4e930049a0fe34.15.0-rc4-717-g048de2c470 , in 4.16, re-ordered the kernel measurement location in efi_start from inside the config file loop to after that loop and the device tree parsing.

With tboot-measure-and-launch-from-xen-efi.patch, a faux diff of pseudo code is roughly:
```
 efi_start
-    measure kernel
     measure things in PCR8
     arch_late
         measure tboot
         measure xen.efi
+    measure kernel
```


```
 pcr4_objs=(
     /boot/shimx64.efi
     /boot/xen.efi
-    /boot/bzImage
     /boot/tboot
     /boot/xen.efi
+    /boot/bzImage
 )
```

So I think the pcr4_objs should just be re-ordered and shim-support-for-shim-lock-measure.patch can just patch the new "measure kernel" location.  Since we forward seal from a chroot, we'll use the new pcr4 order and things should be fine.

This also makes pcr4 follow the actual execution order since the boot goes through xen -> tboot -> xen -> bzImage.

recipes-extended/xen/files/libxl-disable-json-updates.patch
crogers1: ```xl network-list <guest domid>``` should show the updated backend though, and the network in the guest should come back up after the ndvm finishes rebooting.  Xl wouldn't use the ```/tmp/xenmgr-xl``` values until it a new ```xl create``` process is run, or the exiting ```xl create``` does a ```goto``` back to the top where it re-parses the config options (but I think that only happens on a domain reboot, during which xenmgr would re-write the ```/tmp/xenmgr-xl``` file).

jandryuk: I tried to differentiate `xec-vm reboot` and guest-internal reboot.  `xec-vm reboot` definitely re-writes the config, and it is available for re-parsing.  For guest-internal, it looks like the XenMgr synchronization signal delays the "goto" until after the xl config is re-written.

network-daemon has moveToNetwork which calls xl network-attach/detach which does the re-attach for dom0 when an NDVM reboots.  Looks like xenmgr calls xl network-detach/attach for other domUs when an NDVM reboots.

xl network-list output looks shows the backend changing.

Using xl network-attach/detach should do the right thing with respect to the json since it's going through libxl.

So what would you like to see with respect to this change?

recipes-extended/xen/files/shim-support-for-shim-lock-measure.patch
tsirakisn: +1 for aligning the order with upstream. Like you said forward seal should be unaffected.

recipes-extended/xen/files/libxl-disable-json-updates.patch
crogers1: Hmm, I guess if all the edge cases we discussed are working correctly without this patch and we don't see any failures, then it's fine to drop it.

recipes-extended/xen/files/shim-support-for-shim-lock-measure.patch
crogers1: Also +1

recipes-extended/xen/files/libxl-openxt-helpers.patch
jandryuk: It is okay to remove qmp-helper from libxl-openxt-helpers.patch in its own patch closer to the end of the queue?  I'd rather not have to deal with rebasing and all the patch offset conflicts if I don't have to.  A repo of patches doesn't rebase easily.

jandryuk: I removed qmp-helper earlier in the series and then fixed up from there.  Some of the intermediate points might be slightly off for their patch offsets.

jandryuk:
The force-pushed diff shows not too much has changed. https://github.com/OpenXT/xenclient-oe/compare/32d6c869bfae1f55c01a4d8ff72a17a2119832bb..6fe4748503dfafe19c8c27a7206a2957b30aeb41

I added some missing signed-off-by lines

The end has three new commits:
https://github.com/OpenXT/xenclient-oe/pull/1461/commits/dca38083a9d074bc7e49bec944eed658853689ea
https://github.com/OpenXT/xenclient-oe/pull/1461/commits/19521675befd512d032a6f9828e3f70bb36e5abc
https://github.com/OpenXT/xenclient-oe/pull/1461/commits/6fe4748503dfafe19c8c27a7206a2957b30aeb41

The defconfig refresh is self explanatory.  The others address some PCI attachment issues.  The linux patch was accepted into a staging tree.  The vchan marker ones were submitted to Xen but have yet to be applied upstream.  Rebuild both dom0 and stubdom with them or you can't get stubdom-ed HVMs to start.

I didn't refresh all the Xen patches.  There is some fuzz/offsets but that would add noise to the review.  I can refresh later.

vglass vkb/vinput change is still pending and this PR now expects that other vglass piece.

jandryuk:
The problem is passing 2+ PCI devices to a single VM can have the 2nd (or
later) assignment fail with a timeout.

libxl opens a QMP connection to QEMU, QMP handshakes, sends commands and
closes it.  Non-stubdom, this is all QMP over UNIX sockets inside dom0.
The closing and opening of the QMP UNIX socket is important for QEMU to
reset the QMP state and expect a new handshake.

For the stubdom case in dom0, vchan-socket-proxy listens on a UNIX socket.
When it receives a connection, it connects to the stubdom
vchan-socket-proxy vchan server,  The stubdom vchan-socket-proxy then UNIX
socket connects to QEMU.  Then the QMP handshake, commands and close happen.

The problem is that while dom0 vchan-socket-proxy closed and re-opened, the
stubdom side vchan-socket-proxy may not notice, so it doesn't UNIX close
and connect again.  There is no handshake and libxl doesn't get the
response it expects leading to a timeout.

These patches use an in-band marker to signal the connection reset to have
vchan-socket-proxy do the close() & connect() again.

tsirakisn:
@jandryuk appreciate the explanation. I'm following. Something is still odd about it to me but I don't know what...in any case, LGTM.

jandryuk:
> The problem is that while dom0 vchan-socket-proxy closed and re-opened, the
stubdom side vchan-socket-proxy may not notice, so it doesn't UNIX close
and connect again. 

It's a limitation of libxenvchan.  libxenvchan's server side re-uses the same granted out page.  There is only a "connected" indication.  If the client disconnects and a new one reconnects, there is no expression of that event.  The server can only catch it if it checks and observes a disconnected state in between.  When it is missed, QEMU/QMP in the stubdom is out-of-sync with libxl in dom0.  libxl doesn't get the expected QMP messages and it times out.

The reconnect-marker byte clearly indicates that a re-connection occurred, which lets the server side disconnect and reconnect to the QMP socket.  That resets everything to the expected state for libxl and it works.

tsirakisn:
> It's a limitation of libxenvchan. libxenvchan's server side re-uses the same granted out page. There is only a "connected" indication. If the client disconnects and a new one reconnects, there is no expression of that event.

Ah okay. That makes a lot more sense. Thanks!

jandryuk:
Added a compatibility fix for when stable-4.16 advances.

tsirakisn (approved):

PR #1464, opened 2022-11-08
tboot fixes

jandryuk:
I made these two changes while working on the Xen 4.14 & 4.16 uprevs.  I got OpenXT's UEFI -> xen -> tboot -> xen running in QEMU so I could debug it.

[tboot/pe: Fix PE load address calculation](https://github.com/OpenXT/xenclient-oe/commit/9bcd088fdb35205c12c8fd190a682f139b090362) is a bug fix.  `efi_memmap_get_highest_sized_ram()`'s return value isn't checked,so the uninitialized ram_size is used.

[tboot/PE: Or-in PAE flag to cr4](https://github.com/OpenXT/xenclient-oe/commit/872e442d5d685be5f22df14df52e9e633495ab09) fixes running tboot under QEMU.  My QEMU instance didn't emulate enough of CR4 that it faulted when trying to set all the values.  In particular, QEMU doesn't support TXT(SMX), so that can't be set in cr4.  This change makes it a little more proper to just or-in the desired cr4 PAE flag instead of using the hard coded set.

vGlass (and maybe disman) don't run because SSE3 is missing where I was testing.  Specifying a sufficient cpu like `qemu-system-x86_64 -cpu Skylake-Client-v4,+ssse3` lets them run.

OpenXT hardcodes NDVM as HVM and UIVM as PVH, so neither can run by default under QEMU/KVM.  If they are changed to PV, they will start and you can have the OpenXT gui in QEMU.  Nested virt could work, but I did not investigate.  I logged in via the serial console, used db-write to change virt-type and started them manually.

### DIFF
--- a/recipes-security/tboot/tboot-1.9.12/0010-Add-support-for-launching-64-bit-PE-kernels.patch
+++ b/recipes-security/tboot/tboot-1.9.12/0010-Add-support-for-launching-64-bit-PE-kernels.patch
@@ -418,7 +418,7 @@ new file mode 100644
 index 0000000..bafb574
 --- /dev/null
 +++ b/tboot/common/pe.c
-@@ -0,0 +1,221 @@
+@@ -0,0 +1,227 @@
 +/*
 + * pe.c: support functions for manipulating PE binaries
 + */
@@ -606,12 +606,18 @@ index 0000000..bafb574
 +
 +    uint32_t limit = get_lowest_mod_start(lctx);
 +
-+    efi_memmap_get_highest_sized_ram(image_max + (1 << 21), (u32)limit,
-+                          &ram_base, &ram_size);
-+
-+    if ( ram_size < image_max ) {
-+        printk(TBOOT_ERR"Not enough memory to load kernel.\n");
-+        return false;
++    /*  NOTE: the memory maps have been modified already to reserve critical
++        memory regions (tboot memory, etc ...). get_highest_sized_ram
++        will return a range that excludes critical memory regions. */
++    if (!efi_memmap_get_highest_sized_ram(image_max + (1 << 21), (u32)limit,
++                                          &ram_base, &ram_size)) {
++        if (!e820_get_highest_sized_ram(image_max + (1 << 21), (u32)limit,
++                                        &ram_base, &ram_size)) {
++            printk(TBOOT_INFO"ERROR No memory area found for image"
++                             "relocation!\n");
++            printk(TBOOT_INFO"required 0x%X\n", image_max + (1 << 21));
++            return false;
++        }
 +    }
 +
 +    if ( ram_base + ram_size > (u32)image )
--- a/recipes-security/tboot/tboot-1.9.12/0010-Add-support-for-launching-64-bit-PE-kernels.patch
+++ b/recipes-security/tboot/tboot-1.9.12/0010-Add-support-for-launching-64-bit-PE-kernels.patch
@@ -188,7 +188,7 @@ new file mode 100644
 index 0000000..39b8422
 --- /dev/null
 +++ b/tboot/common/paging_64.c
-@@ -0,0 +1,224 @@
+@@ -0,0 +1,225 @@
 +/*
 + * paging_64.c: Enable 64-bit paging and jump to PE entry points
 + */
@@ -348,7 +348,8 @@ index 0000000..39b8422
 +     "lgdtl %[gdt]\n"
 +
 +    // Enable PAE
-+     "mov %[cr4], %%eax\n"
++     "mov %%cr4, %%eax\n"
++     "or  %[cr4_pae], %%eax\n"
 +     "mov %%eax, %%cr4\n"
 +
 +    // Load our page tables
@@ -397,7 +398,7 @@ index 0000000..39b8422
 +     [lme] "i" (1 << _EFER_LME),
 +     [cr0_nopg] "i" (CR0_PE | CR0_MP | CR0_NE),
 +     [cr0] "i" (CR0_PE | CR0_MP | CR0_NE | CR0_PG),
-+     [cr4] "i" (CR4_DE | CR4_PAE | CR4_MCE | CR4_FXSR | CR4_XMM | CR4_SMXE),
++     [cr4_pae] "i" (CR4_PAE),
 +     [ds64] "i" (__BOOT_DS64),
 +     [cs64] "i" (__BOOT_CS64),
 +     "S" (&pe_data)


crogers1:
For the tboot changes, what kind of hardware did you test?  I'd like to try this out on a couple different machines to see what the effect is.  Code wise, all this looks fine, though I'm a little rusty with tboot at the moment.

jandryuk:
I've tested with a Dell laptop with Intel 8th gen processor, Dell laptop with 10th gen, and an HP with 8th gen, in addition to QEMU.

[tboot/PE: Or-in PAE flag to cr4](https://github.com/OpenXT/xenclient-oe/pull/1464/commits/872e442d5d685be5f22df14df52e9e633495ab09) seems more correct to me, and it done elsewhere in tboot like:
https://github.com/tklengyel/tboot/blob/c9ca81f41b4a3eaa8e00c12a3d1d6d3c783782d4/tboot/common/paging.c#L218

I think this change will allow the UEFI -> tboot path to work on more systems, as it working in QEMU shows.  The old setting of CR4_SMXE (SMX/TXT) and CR4_XMM would fault when they aren't present, effectively hard coding their requirement.  This wouldn't affect a particular system, but just allow OpenXT to run on more hardware.

[tboot/pe: Fix PE load address calculation](https://github.com/OpenXT/xenclient-oe/pull/1464/commits/9bcd088fdb35205c12c8fd190a682f139b090362) is correcting an actual bug, but systems could stop working now that a correct check is performed.

crogers1:
Looking closer at the cr4 change, why wouldn't we want those OR'd in during the host loading of tboot?

recipes-security/tboot/tboot-1.9.12/0010-Add-support-for-launching-64-bit-PE-kernels.patch
```diff
@@ -397,7 +398,7 @@ index 0000000..39b8422
 +     [lme] "i" (1 << _EFER_LME),
 +     [cr0_nopg] "i" (CR0_PE | CR0_MP | CR0_NE),
 +     [cr0] "i" (CR0_PE | CR0_MP | CR0_NE | CR0_PG),
-+     [cr4] "i" (CR4_DE | CR4_PAE | CR4_MCE | CR4_FXSR | CR4_XMM | CR4_SMXE),
```
crogers1, line 400: I'm actually not 100% convinced this section here is only for turning on PAE.  This is the last thing that happens before we jump back to a 64-bit PE.  In our case it's the xen.efi, but it really could be any PE.  Perhaps DeGraf wanted to leave the control register in a somewhat default (?) state (for an Intel vPRO chip) for the next PE to load after this call.

By the time we get to ```launch_kernel``` (and eventually this function) we've already done SENTER, which needs SMXE in CR4.  I guess since we aren't making any more SMX-required calls of the processor until we shutdown the platform (that I know of), and SMXE is guaranteed to be OR-d back into CR4 from a line in ```txt_shutdown```, ultimately it may not matter if we don't set it in CR4 before launching xen.efi the second time.  But I think this would represent a significant runtime difference in that on all versions prior to this, we have always run OpenXT with SMXE and these other vPRO flags in CR4, and taking this change would make that no longer the case.

jandryuk: From memory, I think the code constructs page tables with the large bit set, so they are 2MB mappings.  CR4 needs PAE to support that.

tboot will boot without SMX/TXT - that is how I can get through here in QEMU.  I think you are supposed to check CPUID before you set it in CR4.  So this unilateral setting is wrong.  When SMX support is not detected, tboot shortcuts through most of its code.

My guess is de Graaf printed CR4 at some point and set it to that since it worked for the test system.  Since all vPro systems have the ~same capabilities, that was fine for real hardware.

Going through the CR4 bits:
I think these two made QEMU print out all the floating point registers (though maybe it was doing that regardless of these):
CR4_FXSR - Operating system support for FXSAVE and FXRSTOR instructions
CR4_XMM - Operating System Support for Unmasked SIMD Floating-Point Exceptions
tboot doesn't do floating point, so they don't need to be enabled, AFAIK.

CR4_DE - Debugging Extensions
These shouldn't be needed in normal operation, again AFAIK.
CR4_MCE Machine Check Exception
Not sure, but if this triggers interrupt generation, we don't want it set when transitioning to new code.

CR4_PAE - I think this is needed for [Physical Address Extension](https://en.wikipedia.org/wiki/Physical_Address_Extension) compatible page tables.

CR4_SMX - shouldn't be needed moving forward.

We're transitioning to a new PE, so it's a new software environment.  Xen has it's own code to handle CR4.

I see your point that this is a change in the processor CR4 configuration going forward.  I guess I could print CR4 with and without this change and see what changes in practice.  I guess my thought is that any new code needs to query CPUID and handle CR4 as it sees fit, so the new PE shouldn't be depending on the old value.

Real hardware that had SMX set will continue to have it set.  tboot is the only thing that does SMX/TXT instructions anyway, so having it flipped off outside of tboot shouldn't matter.  I expect Xen/Dom0 handle the other bits as necessary.

I'll report back in a bit with a finding on CR4 with and without this change.

jandryuk: On an i7-8665U, when entering jump_pe_image, cr4 is 0x00044240

```
                    0x00044240
#define CR4_MCE     0x00000040 /* Machine check enable */
#define CR4_FXSR    0x00000200/* Fast FPU save/restore used by OS */
#define CR4_SMXE    0x00004000/* enable SMX */
#define CR4_OSXSAVE 0x00040000 /* enable xsave and xrestore */
```

So my change would drop CR4_DE and CR4_XMM (Called CR4_OSXMMEXCPT elsewhere).

CR4_DE
"Debugging Extensions"
"If set, enables debug register based breaks on I/O space access."

CR4_OSXMMEXCPT
"Operating System Support for Unmasked SIMD Floating-Point Exceptions"
"If set, enables unmasked SSE exceptions."

Xen __start_xen() sets CR4_OSXMMEXCPT during boot.

So CR4_DE is let unset.  From the SDM:
17.2.2 Debug Registers DR4 and DR5
Debug registers DR4 and DR5 are reserved when debug extensions are enabled (when the DE flag in control register CR4 is set) and attempts to reference the DR4 and DR5 registers cause invalid-opcode exceptions (#UD).  When debug extensions are not enabled (when the DE flag is clear), these registers are aliased to debug registers DR6 and DR7.

I think it's fine to leave this to Dom0/Xen to toggle as they need.  Xen looks to mainly reference CR4_DE for guest handling.  CR4_DE is included in:
```
/*
 * Bits which a PV guest can toggle in its view of cr4.  Some are loaded into
 * hardware, while some are fully emulated.
 */
#define PV_CR4_GUEST_MASK \
    (X86_CR4_TSD | X86_CR4_DE | X86_CR4_FSGSBASE | X86_CR4_OSXSAVE)
```

Linux itself clears CR4_DE for 64bit kernels (arch/x86/kernel/cpu/common.c).

Xen would also end up clearing out CR4_DE.  It starts with `mmu_cr4_features =  XEN_MINIMAL_CR4 = (X86_CR4_PGE | X86_CR4_PAE)` and  set_in_cr4() or-s in options.  It's in the later parts of __start_xen(), but this all means Xen doesn't rely on the inherited cr4.

crogers1 (approved):
LGTM.

PR #1465, opened 2023-03-13
Build fixes

jandryuk:
Build fix for networkmanager uprev in upstream OE
Patch context update for Xen hvm-rtc.patch


recipes-connectivity/networkmanager/networkmanager_1.22.%.bbappend
```diff
@@ -1,4 +1,4 @@
-FILESEXTRAPATHS_prepend := "${THISDIR}/${BPN}-${PV}:${THISDIR}/files:"
+FILESEXTRAPATHS_prepend := "${THISDIR}/${BPN}-1.22:${THISDIR}/files:"
```
crogers1, line 1: OE supports `-%` in the dir name, so this can stay `${PV}` and the patches dir gets named to ```networkmanager-1.22.%```.

crogers1: Actually I guess it's the same effect as naming it `networkmanager-1.22`, the real benefit would be to set it to `networkmanager-1.%` to handle minor and micro version changes.

jandryuk: I think I assumed that did not work...  Thanks!

crogers1 (approved):
LGTM. Merging soon

PR #1466, opened 2023-03-22
Default to read-only blktap

jandryuk:
Change the default for `tap-ctl create` to read-only, the former -R option.  Add a new -W option to use when read-write is desired.

Now that tap-ctl defaults to read-only, we need to specify -W when opening writable .vhds.  For the read-only cases, explicity add -R to highlight the use.

OpenXT is changing to hash .vhds directly.  To do so, we need to only open .vhds as read-only as read-write can modify the metadata even if the contents do not change.  To avoid user error, default to read-only so explicit action is needed which will potentially modify the .vhd.

### DIFF
--- a/recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
+++ b/recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
@@ -412,7 +412,7 @@ xenmgr to allow tapdisk to unlock encrypted VHDs.
      if ! tap-ctl check >& /dev/null ; then
  	fatal "Blocktap kernel module not available"
      fi
-@@ -81,23 +81,124 @@ find_device()
+@@ -81,23 +81,126 @@ find_device()
      done
  
      if [ -z "$pid" ] || [ -z "$minor" ]; then
@@ -511,6 +511,8 @@ xenmgr to allow tapdisk to unlock encrypted VHDs.
 +
 +    if [ "$mode" = "r" ]; then
 +        tap_opts="-R"
++    else
++        tap_opts="-W"
 +    fi
 +
 +    claim_lock "block"
--- a/recipes-openxt/vhd-scripts/vhd-scripts/vhd-copy
+++ b/recipes-openxt/vhd-scripts/vhd-scripts/vhd-copy
@@ -84,7 +84,7 @@ trap atexit EXIT
 
 size=$(vhd-util query -n $srcvhd -v)
 [ $? -eq 0 ] || error couldn\'t query size of src $srcvhd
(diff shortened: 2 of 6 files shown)


crogers1:
Build here: https://openxt.ainfosec.com/buildbot/#/builders/22/builds/80

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
```diff
@@ -511,6 +511,8 @@ xenmgr to allow tapdisk to unlock encrypted VHDs.
 +
 +    if [ "$mode" = "r" ]; then
 +        tap_opts="-R"
++    else
++        tap_opts="-W"
```
crogers1, line 515: This would cover the non-measured, HVM (Windows/Linux) use case?

recipes-openxt/vhd-scripts/vhd-scripts/vhd-copy
```diff
@@ -84,7 +84,7 @@ trap atexit EXIT
 
 size=$(vhd-util query -n $srcvhd -v)
 [ $? -eq 0 ] || error couldn\'t query size of src $srcvhd
-srctap=$(TAPDISK2_CRYPTO_KEYDIR="$srckeydir" TAPDISK3_CRYPTO_KEYDIR="$srckeydir" tap-ctl create -a "vhd:$srcvhd")
+srctap=$(TAPDISK2_CRYPTO_KEYDIR="$srckeydir" TAPDISK3_CRYPTO_KEYDIR="$srckeydir" tap-ctl create -R -a "vhd:$srcvhd")
```
crogers1, line 87: Are these `-R`'s needed if the patch to tap-ctl.c directly initializes flags to readonly?

recipes-extended/xen/files/0001-convert-block-tap-for-shared.patch
jandryuk: Yes.  read-only/read-write is independent of measured, but measured only measures read-only disks.

recipes-openxt/vhd-scripts/vhd-scripts/vhd-copy
jandryuk: The -R's are not needed, but I added them to be explicit (and that was mentioned in the change log).

I don't like adding a blktap patch to implement this, but I don't know how else to do it.  It isn't strictly necessary, but it is a little safer to default to read-only.

crogers1: Claiming a VHD is read-only and measured but not measuring the metadata is "fine", but what this change does is strengthen the read-only+measured assertion to extend to the VHD metadata.  It removes a weird edge case and increases the security posture of the system.  Maintaining a small patch to better ensure a VHD isn't mounted read-write except when an admin/owner explicitly wants it is an acceptable price in my opinion, given how disastrous user-error can be.

crogers1 (approved):
LGTM.

PR #1467, opened 2023-03-22
Add grub-xen-pvh & grub-xen-pv64

jandryuk:
This switches to using grub as a PVH/PV64 boot loader for booting VMs.  It removes extracting kernels from disk images, though that functionality still remains.

grub2 doesn't support a 32bit version, and PV32 shouldn't be used these days, AFAIUI.

https://github.com/OpenXT/manager/pull/205 changes the templates to use grub-xen-pvh/pv64.  I think this could go in first without the templates change and things would keep working.

### DIFF
--- a/recipes-bsp/grub/grub-xen-pvh_2.04.bb
+++ b/recipes-bsp/grub/grub-xen-pvh_2.04.bb
@@ -1,0 +1,4 @@
+require grub-xen.inc
+GRUBPLATFORM = "xen_pvh"
+GRUB_TARGET = "i386"
+GRUBEXT = "xen-pvh"
--- a/recipes-bsp/grub/files/grub-xen.cfg
+++ b/recipes-bsp/grub/files/grub-xen.cfg
@@ -1,0 +1,30 @@
+insmod part_msdos
+insmod part_gpt
+insmod search
+insmod configfile
+insmod legacycfg
+if search -s root -f /boot/grub2/grub.cfg ; then
+  echo "Reading ${root}/boot/grub2/grub.cfg"
+  configfile /boot/grub2/grub.cfg
+elif search -s root -f /boot/grub/menu.lst ; then
+  legacy_configfile /boot/grub/menu.lst
+elif search -s root -f /grub2/grub.cfg ; then
+  echo "Reading ${root}/grub2/grub.cfg"
+  configfile /grub2/grub.cfg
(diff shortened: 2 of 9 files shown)


crogers1:
Build here: https://openxt.ainfosec.com/buildbot/#/builders/22/builds/80

recipes-bsp/grub/grub-xen-pvh_2.04.bb
```diff
@@ -0,0 +1,4 @@
+require grub-xen.inc
+GRUBPLATFORM = "xen_pvh"
```
crogers1, line 2: Is `xen_pvh` turned on by default in grub2 now? The xen instructions here from 4.12 make a note of compiling it using the `--with-platform=xen_pvh` flag. https://wiki.xenproject.org/wiki/Xen_4.12_RC_test_instructions.

Just making sure we don't need to do anything to the base grub2 recipe

recipes-bsp/grub/grub-xen-pvh_2.04.bb
```diff
@@ -0,0 +1,4 @@
...
+GRUB_TARGET = "i386"
```
crogers1, line 3: Why does xen_pvh build as a 32 bit binary?  This is just how xen supports it or was there some other special reason?

recipes-bsp/grub/grub-xen-pvh_2.04.bb
jandryuk: GRUBPLATFORM is the variable that is passed to `--with-platform`:

https://github.com/openembedded/openembedded-core/blob/f45ddfbf007de858327eef0ffefd5840ef4c69b8/meta/recipes-bsp/grub/grub2.inc#L77

crogers1 (approved):
LGTM merging soon

jandryuk:
One thing - In the embedded config, it has `elif search -s root -f
/@/boot/grub2/grub.cfg ; then`. I copied that from somewhere (Xen wiki?),
But I don't think that @ is useful for us. I should probably drop that and
re-push.

crogers1:
Good catch. You want me to run it through the autobuilder one more time?

jandryuk:
I think it's okay to skip going through the autobuilder.  I did test locally, and the diff looks clean.

crogers1:
Merging in a few then.

PR #1470, opened 2023-05-11
Support BIOS SINIT modules

jandryuk:
Intel doesn't really post SINIT modules anymore.  Also, with Converged BootGuard and TXT (CBnT), the BIOS-provided ACM doesn't double duty for both BootGuard and TXT.  It's something like there can be only a single BIOS ACM.  Since it has to be there for BootGuard, it also does TXT.

Since there may not be an SINIT ACM on on disk to match against, we can't rely on it to parse the ACM caps.  Instead (kinda ugly) grep it out of txt-stat output.  It's logged there by tboot and we just need a single 32bit value.

pcr-calc is updated to accept ACM caps on the command line.

### DIFF
--- a/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions
+++ b/recipes-openxt/openxt-measuredlaunch/openxt-measuredlaunch/ml-functions
@@ -510,10 +510,20 @@ calculate_drtm_pcrs() {
         else
             acm="$(match_acm ${root})"
         fi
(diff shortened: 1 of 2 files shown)


dpsmith:
Unfortunately this is not ideal, but until a method is crafted to allow dom0 either access (direct memory) or details (expand tboot hypercall), this is probably the best way in the short term.

crogers1 (approved):

crogers1:
Merging soon.

PR #1474, opened 2023-06-14
dunfell: Set SRCREVs for dunfell release

crogers1:


recipes-openxt/surfman/surfman.inc
```diff
@@ -1,4 +1,4 @@
 PV = "0+git${SRCPV}"
 
-SRCREV = "${AUTOREV}"
+SRCREV = "46bfde5850d92399d36edb24313f2e404d53d4dd"
```
jandryuk, line 3: Doesn't need to be done now, but this recipe should be deleted.

jandryuk (approved):
LGTM - I double checked all the hashes, too.

recipes-openxt/surfman/surfman.inc
crogers1: I can only think it was kept for "optional build" purposes.

crogers1:
No functional change, merge whenever.

PR #1475, opened 2023-07-05
Mark devicemodel-stubdom & xenclient-stubdomain-kernel as dom0.

jandryuk:
Bitbake can log messages when the SRC_URI files are missing for these recipes.  As the are only expected to be used for dom0, use COMPATIBLE_MACHINES to mark them as such.


crogers1 (approved):
LGTM, merging soon.

PR #1477, opened 2023-08-17
grub-xen: Fix binary layout to avoid faulting

jandryuk:
After the recent grub updates in oe-core, we have faults on startup in the two pv-grubs.

grub-xen-pv64:
(XEN) d17v0 Unhandled invalid opcode fault/trap [#6, ec=ffffffff] (XEN) domain_crash_sync called from entry.S: fault at ffff82d0402fdfc8 x86_64/entry.S#create_bounce_frame+0x14f/0x167

grub-xen-pvh:
(XEN) MMIO emulation failed (1): d13v0 32bit @ 0000:ffffffff ->
(XEN) d13v0 Triple fault - invoking HVM shutdown action 1

GCC is generating a .text.unlikely section with terminate_arg.cold(), and that is placed ahead of the normal _start code in the text segment. The entry points still point at the beginning of the text segment, which faults as it is terminate_arg.cold().

Passing -fno-reorder-functions inhibits use of .text.unlikely, so terminate_arg.cold() is left in the main .text section.  This leaves _start at the beginning of the text segment and things work as expected.

### DIFF
--- a/recipes-bsp/grub/grub-xen.inc
+++ b/recipes-bsp/grub/grub-xen.inc
@@ -9,6 +9,11 @@ require conf/image-uefi.conf
 # Need to be disabled otherwise there are wint_t type not defined errors
 SECURITY_CFLAGS = ""
 
+# Prevent generation of .text.unlikely which would be placed ahead of the
+# expected _start at the beginning of the .text section.  .text.unlikely
+# would then get executed and fault.
+CFLAGS_append = "-fno-reorder-functions"
+
 DEPENDS_append_class-target = " grub-efi-native"
 RDEPENDS_${PN}_class-target = "grub-common virtual/grub-bootconf"
 


crogers1:
Built and tested locally. UIVM boots as expected using grub-xen-pvh, no triple faults observed.  Merging soon.

crogers1 (approved):

jandryuk:
```
(XEN) avc:  denied  { memorymap } for domid=4 scontext=system_u:system_r:uivm_t tcontext=system_u:system_r:uivm_t tclass=mmu
```

This is seen, but it actually started with the original https://github.com/OpenXT/xenclient-oe/pull/1467 adding grub-xen-pvh.

grub-xen-pvh works without the permission to grab the memory map, so it's not necessary to add it.  It could be silenced with dontaudit, but that would hide it if it started to matter in the future.  It doesn't hurt anything, so just leave it for now.

jandryuk:
I looked into this grub XSM denial because I worried it might be a lurking issue.  It isn't an issue AFAICT.  My earlier comment is incorrect - the denied `memorymap` is for then the guest is trying to call `XENMEM_set_memory_map` to *update* its e820 memory map in the hypervisor (not grab it - that is allowed).  Typically, dom0 sets the memory map when building the domain and the domain doesn't have to touch it, which is why we don't allow the permission today.

When grub runs as Xen PVH, grub has to map in the shared_info page and the grant table page.  The logic is to look for an unused memory range:

grub grabs the e820 table, finds unused areas to map the pages, maps them, and then updates the e820 table.  When finished, grub stores the new e820 table back into the hypervisor.

For the unused areas, grub looks for a hole in the e820 table and uses that.  If none is found, then it tries to use `the highest RAM page below 4G and reserve it`.  In that case, grub updates the e820 table to reserve the memory.  This is important so that the next launched guest doesn't start scribbling over the live memory of the shared_info for grant_table pages - Xen is still treating them as such.

However, my testing shows that a libxl-created PVH domain never changes the e820 table.  Any RAM size up to 0xfc000000 is in the first entry.  Any additional RAM is allocated above 4GB.  Like this:

```
0: 0000000000000000 - 00000000fc000000 RAM
1: 00000000fc000000 - 00000000fc009000 ACPI
2: 00000000feff8000 - 00000000ff000000 Reserved
3: 0000000100000000 - 0000000104000000 RAM
```

So there are always enough unassigned space available for grub to map the two special pages.  Here the two pages are allocated at 0xfc009000 and 0xfc00a000.  If there is less RAM, the pages get allocated in the gap between the first RAM entry and the ACPI entry.

Hmmm, libxl actually provides an un-ordered non-page-aligned e820:
```
0: 0000000000000000 - 00000000fc000000 RAM
1: 00000000feff8000 - 00000000ff000000 Reserved
2: 00000000fc008000 - 00000000fc008040 ACPI
3: 00000000fc000000 - 00000000fc001000 ACPI
4: 00000000fc001000 - 00000000fc008000 ACPI
5: 0000000100000000 - 0000000104000000 RAM
```

pvh-grub takes that and sorts and page-aligns it.  The grub version is what is passed to Linux and it seems to use.

So not being able to update the e820 table will not be a problem for a PVH domain since it is never changed.  Still, should we allow guests to have the XSM permission.  It looks like the memory map is just stored by Xen and doesn't get processed or consulted by Xen internally.  There is a size limit applied to the new value, so the guest can't consume unbounded memory.  grub could also be patched to skip re-writing the e820 table if it hasn't changed.  Given the sorting, skipping the set_memory_map hypercall would require some fudging.  Thoughts?

PR #1478, opened 2023-08-17
Update ACMS for DRTM measured launch

crogers1:
This PR introduces Intel's latest ACMs released here: https://www.intel.com/content/www/us/en/developer/articles/tool/intel-trusted-execution-technology.html version 630744 into OpenXT.

The first commit updates the URLs to these ACMs on the mirror, and it also gives them new names which are abbreviations of the cpu codename they apply to.  It also updates the hashes and starts tracking a license file from a recent ACM since the 630744 version no longer distributes the ACMs bundled with their error codes and individual license files.

The second commit updates their usage in isolinux and pxelinux cfgs.  The installer.git code doesn't need to be updated as the ACMs are built into the dom0 image under /boot, and copied to the ESP for EFI boot in the function `copy_to_esp()` in `part2/stages/functions/install-main`


jandryuk:
Instead of adding `license.txt`, should `Legal Disclaimer.pdf` be used for the "license"?  It's not a license per-say, but it is included in the source zip file.

recipes-openxt/xenclient-dom0-tweaks/xenclient-dom0-tweaks/openxt.cfg needs to be updated for the new modules?  Hmm. if you OTA, the old modules are probably left on the ESP.  I don't think we have any support for clearing out old files from there.

This is dropping some support for older systems?  That is fine from my end, but we should be clear about it.

However, I have an issue with at least one modern system.  When I use CFL_SINIT_20211019_PRODUCTION_REL_NT_O1_1.10.0.bin with it, it reboots, seemingly in tboot.  i.e. it enters a boot loop.  Normally it uses the BIOS-provided SINIT.  With Converged Bootguard and TXT (CBnT), TXT-enabled systems are shipping with an embedded SINIT module.  I'm wondering if these external SINIT modules are really needed for newer systems.  As I've seen, they can be problematic.

crogers1:
> Instead of adding `license.txt`, should `Legal Disclaimer.pdf` be used for the "license"? It's not a license per-say, but it is included in the source zip file.
> 
Thought about that, but since it wasn't actually a license, didn't think it qualified.

> recipes-openxt/xenclient-dom0-tweaks/xenclient-dom0-tweaks/openxt.cfg needs to be updated for the new modules? Hmm. if you OTA, the old modules are probably left on the ESP. I don't think we have any support for clearing out old files from there.
> 
from `part2/stages/Functions/install-main`:
```
 623     # make sure the ESP is starting from a clean slate                                                                                                                                                          
 624     # if we're upgrading. prevents outdated files from                                                                                                                                                          
 625     # older versions from populating the partition.                                                                                                                                                             
 626     if [ "${INSTALL_MODE}" = "upgrade" ]; then                                                                                                                                                                  
 627         do_cmd rm -rf ${ESP_MOUNT}/EFI/BOOT >&2                                                                                                                                                                 
 628         do_cmd rm -rf ${ESP_MOUNT}/EFI/OpenXT >&2                                                                                                                                                               
 629     fi
 ```
That should cover the old .acms
> This is dropping some support for older systems? That is fine from my end, but we should be clear about it.
> 
I can't be 100% sure about that.  Based on the support matrix in the .zip it seems like they combined some platforms into a single SINIT.  It however doesn't say anything about Xeon processors, for which we had separate SINIT's defined in the recipe.  In practice they might just be BIOS provided but we'd need to test them out for sure.

> However, I have an issue with at least one modern system. When I use CFL_SINIT_20211019_PRODUCTION_REL_NT_O1_1.10.0.bin with it, it reboots, seemingly in tboot. i.e. it enters a boot loop. Normally it uses the BIOS-provided SINIT. With Converged Bootguard and TXT (CBnT), TXT-enabled systems are shipping with an embedded SINIT module. I'm wondering if these external SINIT modules are really needed for newer systems. As I've seen, they can be problematic.

That's an issue then, if there's a txt errorcode you can pull that might tell us what's wrong. Otherwise I was under the impression it would prefer BIOS provided SINITs over the ones we provide.  Maybe it has something to do with whichever version is newer.  Are there any firmware updates for your system that's using the CFL SINIT?

**Comment by crogers1** [comment]
Updated with latest Intel ACMs released today.

**Comment by crogers1** [comment]
Updated to include missing changes to the openxt.cfg file referencing the names of the new SINITs.

**Comment by crogers1** [comment]
Added commit that uses `acminfo` to check if we're processing an ACM rather than looking at the fingerprint, since the fingerprint no longer exists for newer ACMs.

File: recipes-security/acms/acms.bb [hunk 1]
```diff
@@ -3,63 +3,42 @@ HOMEPAGE = "https://software.intel.com/en-us/articles/intel-trusted-execution-te
...
-    ${OPENXT_MIRROR}/7th_gen_i5_i7-SINIT_74.zip;name=kbl \
-    ${OPENXT_MIRROR}/8th_gen_i5_i7-SINIT_76.zip;name=cfl \
+    file://license.txt \
+    ${OPENXT_MIRROR}/SNB_IVB_SINIT_20190708_PW.bin;downloadfilename=snb_ivb.bin;name=snb_ivb \
+    ${OPENXT_MIRROR}/BDW_SINIT_20190708_1.3.2_PW.bin;downloadfilename=bdw_hsw.bin;name=bdw_hsw \
+    ${OPENXT_MIRROR}/SKL_KBL_AML_SINIT_20211019_PRODUCTION_REL_NT_O1_1.10.0.bin;downloadfilename=skl_kbl_aml.bin;name=skl_kbl_aml \
+    ${OPENXT_MIRROR}/CFL_SINIT_20221220_PRODUCTION_REL_NT_O1_1.10.1_signed.bin;downloadfilename=cfl_wkl_cml.bin;name=cfl_wkl_cml \
+    ${OPENXT_MIRROR}/CML_S_SINIT_1_13_33_REL_NT_O1.PW_signed.bin;downloadfilename=cml_s.bin;name=cml_s \
+    ${OPENXT_MIRROR}/CMLSTGP_SINIT_v1_14_39_20211214_REL_NT_O1.PW_signed.bin;downloadfilename=cml_s_tgp.bin;name=cml_s_tgp \
+    ${OPENXT_MIRROR}/RKLS_SINIT_v1_14_46_20220818_REL_NT_O1.PW_signed.bin;downloadfilename=rkls.bin;name=rkls \
```
jandryuk, line 19: RKLS failed to download - it's 2022081**9**.

630744_003 also has:
CML_RKL_S_SINIT_v1.13.33_REL_NT_O1.PW_signed.bin
CMLSTGP_SINIT_v1_14_46_20220819_REL_NT_O1.PW_signed.bin
which are not on the mirror or included here in SRC_URI.

jandryuk: The three CML_* files are identical:
```
5670905f72d7415fad97eb02bd994ae75c15ceec  630744_002/CML_S_SINIT_1_13_33_REL_NT_O1.PW_signed.bin
5670905f72d7415fad97eb02bd994ae75c15ceec  630744_003/CML_RKL_S_SINIT_v1.13.33_REL_NT_O1.PW_signed.bin
5670905f72d7415fad97eb02bd994ae75c15ceec  630744_003/CML_S_SINIT_1_13_33_REL_NT_O1.PW_signed.bin
```

jandryuk:
Something else is wrong with an OTA.
```
updatemgr: cp -f /mnt/upgrade/dom0/boot/xen.efi /mnt/upgrade/dom0/boot/shimx64.efi /mnt/upgrade/dom0/boot/tboot /mnt/upgrade/dom0/boot/bzImage /mnt/upgrade/dom0/boot/initrd /mnt/upgrade/dom0/boot/7th_gen_i5_i7-SINIT_74.bin /mnt/upgrade/dom0/boot/8th_gen_i5_i7-SINIT_76.bin /mnt/upgrade/dom0/boot/Xeon-5600-3500-SINIT-v1.1.bin /mnt/upgrade/dom0/boot/Xeon-E7-8800-4800-2800-SINIT-v1.1.bin /mnt/upgrade/dom0/boot/adl.bin /mnt/upgrade/dom0/boot/bdw_hsw.bin /mnt/upgrade/dom0/boot/cfl_wkl_cml.bin /mnt/upgrade/dom0/boot/cml_s.bin /mnt/upgrade/dom0/boot/cml_s_tgp.bin /mnt/upgrade/dom0/boot/microcode_20230808.bin /mnt/upgrade/dom0/boot/microcode_intel.bin /mnt/upgrade/dom0/boot/rkls.bin /mnt/upgrade/dom0/boot/skl_kbl_aml.bin /mnt/upgrade/dom0/boot/snb_ivb.bin /mnt/upgrade/dom0/boot/tgl.bin /mnt/upgrade/dom0/boot/*.BIN /mnt/upgrade/dom0/etc/xen/xenrefpolicy/policy/policy.24 /mnt/upgrade/dom0/usr/share/xenclient/bootloader/openxt.cfg /mnt/upgrade/dom0/usr/share/xenclient/bootloader/BOOT.CSV /mnt/upgrade/esp/EFI/OpenXT
updatemgr: cp: cannot stat '/mnt/upgrade/dom0/boot/*.BIN': No such file or directory
updatemgr: Failed to copy files to ESP
updatemgr: umount /mnt/upgrade/esp
updatemgr: ERROR: Failure committing dom0 with file dom0-rootfs.ext3.gz.
```

Ugh, the installer: https://github.com/OpenXT/installer/blob/e20749637d9a8016af6721d4bc3968cd5978c60a/part2/stages/Functions/install-main#L639

recipes-security/acms/acms.bb
crogers1: My mistake on the 20220819. Pushing a fix for that. The change to v1_14_46 for CMLSTGP slipped past me, pushing a fix for that as well.  It'll also go up on the mirror.

Not sure why they have 2 versions of the same sinit, but the CML_RKL_S isn't on the 630744_ACM_PlatformList_003.pdf, only the CML_S_SINIT_1_13_33 is.  So I'm going to advocate for only listing that one in the `SRC_URI` and chalk up the inclusion of the other one to user error on Intel's part.

crogers1:
> Ugh, the installer: https://github.com/OpenXT/installer/blob/e20749637d9a8016af6721d4bc3968cd5978c60a/part2/stages/Functions/install-main#L639

Hmm, guess we'll have to drop that line.

dpsmith (approved):
Acked-by: Daniel P. Smith <dpsmith@apertussolutions.com>

PR #1479, opened 2023-08-18
Update OpenXT refpolicy-mcs to align with meta-selinux

crogers1:
Meta-selinux recently dropped the 2.2019* refpolicy-mcs recipe, leaving only the general refpolicy-mcs_git.bb they intend to maintain going forward. That recipe builds refpolicy 2.20200229, and since there are 80 patches in the meta-selinux queue, it is more reasonable to simply update our recipes to support this version rather than try to downrev back 2.2019*.  This PR introduces the changes to directory structure, patch context, and new interfaces/file contexts needed to patch, build, and deploy the new refpolicy without introducing new AVC denials.

Commits are split:
- first commit is patches with hunk context changes only
- middle commits are selinux context and interface updates due to the change in refpolicy version
- last commit is updating the patch directory structure, modifying the .bb to remove patches that no longer apply + add new patch, and updating the PREFERRED_VERSION


recipes-security/refpolicy/refpolicy-mcs-2.%/patches/policy.modules.services.acpi.diff
```diff
@@ -1,14 +1,13 @@
...
-+#
-+# OpenXT local policy
-+#
-+# Early devtmpfs, before udev relabel.
-+dev_dontaudit_rw_generic_chr_files(apmd_t)
++gen_require(`
++    type device_t;
++')
++
++allow acpid_t device_t:dir { watch };
```
jandryuk: I think this should be:
dev_watch_dev_dirs(acpid_t);

And then you don't need the gen_require.

recipes-security/refpolicy/refpolicy-mcs-2.%/patches/policy.modules.services.networkmanager.diff
```diff
@@ -306,3 +304,15 @@
...
+ 
+ term_dontaudit_use_console(wpa_cli_t)
++
++require {
++    type lib_t;
++    type etc_t;
++}
++allow NetworkManager_t NetworkManager_etc_rw_t:dir { watch };
++allow NetworkManager_t lib_t:dir { watch };
++allow NetworkManager_t etc_t:dir { watch };
```
jandryuk, line 318: files_watch_etc_dirs(NetworkManager_t);

jandryuk (commented):
Looks great.  Thanks for breaking up the commits like you did - it made review easier.

Two small fix ups where you can use existing interfaces.

policy.modules.services.consolesetup.diff:
Does the upstream consolesetup still have xserver_read_xkb_libs when it shouldn't exist?  Oh it is just that we have xserver stuff disabled, so it should be an optional_policy for OpenXT?

jandryuk:
Maybe we want modules-upstream.conf to disable consolesetup `consolesetup = off`?  

cryfs, knot, stubby, and wireguard are all new modules which may want being turned off.

There is also a new tpm2 module, but that covers abrmd which we don't use.

recipes-security/refpolicy/refpolicy-mcs-2.%/patches/policy.modules.services.acpi.diff
crogers1: Done.

recipes-security/refpolicy/refpolicy-mcs-2.%/patches/policy.modules.services.networkmanager.diff
crogers1: Done.

crogers1:
Removed the allow rules, used recommended interfaces instead. Dropped consolesetup, it was added sometime in the 2.2020 release and we don't need to support it, so we drop the patch as well.  That `xserver` rule issue I was seeing was absolutely due to the fact xserver is `off` in the module conf.  Disabled the other newly added modules @jandryuk recommended via modules-upstream.conf since we're not using those either.

jandryuk (approved):
LGTM - thanks. Will merge soon.

PR #1482, opened 2023-10-31
xen-vif-scripts-ndvm: Trim down the script

jandryuk:
When this vif script runs, it can race with network-slave with both executing iptables commands.  iptables tries to grab a lock to synchronize access, but by default it fails the command if the lock cannot be grabbed:
"Another app is currently holding the xtables lock." "Perhaps you want to use the -w option?"

The `-w` flag to iptables makes it wait for the lock.  This is important so we don't lose iptables commands that define chains used for other commands.  For instance, this was seen:
network-slave: exception: "cannot add rule -A FORWARD -i brbridged -o brbridged -m physdev --physdev-in eth0 -j FORWARD_brbridged : iptables v1.8.4 (legacy): Couldn't load target FORWARD_brbridged':No such file or directory\n\nTry iptables -h' or 'iptables --help' for more information.\n" while processing RPC "configure"

The vif script could gain `-w` for all of the iptables commands, but in fact it is basically unused.

The firewall stuff is actually handled by network-slave.  xenmgr writes to:
/vm/$UUID/firewall-rules/$num/
While ndvm tries to read from:
/vm/$UUID/config/nic/$devid/firewall-rules/$num

ndvm can't access that top-level path since it gets constrained to it's own subtree (/vm/$ndvm_uuid).

network-daemon will write from /vm/$UUID/firewall-rules/$num/ into the NDVM's /vm/$ndvm_uuid/firewall-rules/$guest_uuid/$devid and network-slave will pick those up.

INPUT_vifs and FORWARD_vifs don't exist, so the can never be appended to.

INPUT_${vif} and FORWARD_${vif} don't get used by anything else.

network-slave already create FORWARD_${vif}_{IN,OUT} for its own use.

So all the firewall and iptables stuff can be removed.

The XAPI stuff is legacy and not used by anymore.  Remove it as well.

Goes with https://github.com/OpenXT/network/pull/28

### DIFF
--- a/recipes-openxt/xen-vif-scripts/xen-vif-scripts-ndvm/vif
+++ b/recipes-openxt/xen-vif-scripts/xen-vif-scripts-ndvm/vif
@@ -17,95 +17,16 @@
 # Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 #
 
-TYPE=`echo ${XENBUS_PATH} | cut -f 2 -d '/'`
-DOMID=`echo ${XENBUS_PATH} | cut -f 3 -d '/'`
-DEVID=`echo ${XENBUS_PATH} | cut -f 4 -d '/'`
-
-XAPI=/xapi/${DOMID}/hotplug/${TYPE}/${DEVID}
-
 case "$1" in
     online)
         if [ "x${vif}" = "x" ]; then
             exit
         fi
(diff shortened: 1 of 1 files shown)


jandryuk:
force-pushed to removed some additional unused variables.

crogers1 (approved):
LGTM. Merging soon.

PR #1484, opened 2023-11-15
[ndvm] Drop rules file for vif scripts

crogers1:
Commit c3a29af664ff9b6359845038e123d39d914a434f removed
  xen-vif-backend.rules from xenclient-oe, it also needs to be
  removed from this recipe.


jandryuk:
Did I break the build?  I think we need to keep the .rules file for NDVM, unless it's coming from somewhere else.  NDVM needs its vif script to run, and udev is the way that happens today.

jandryuk:
Sorry if I broke the build.

crogers1:
All good, I'm just stomping out build issues as I go.

I can restore the .rules file, we just want it included for the vif-scripts in the ndvm, but not for dom0 right?

jandryuk:
Yeah, only an ndvm one should exist.

PR #1486, opened 2023-11-29
Fix installer efivarfs mounting and drop efivars from fstab

jandryuk:
Installation is currently broken (after the Linux 6.1 uprev I believe) because efivarfs is not mounted, so efibootmgr fails to write the boot variables.

The first patch fixes the problem.  It manually mounts efivarfs for the installer.

The second patch removes efivarfs from /etc/fstab.  dom0's is mounted early in /init from the initramfs-framework.  With the first patch, the installer manually mounts it in the same fashion.  The fstab entries are therefore unnecessary.  They were always problematic on legacy boots, so just drop them from fstab.

If the second patch is deemed unnecessary, then only the first patch can be taken to fix the issue.


crogers1 (approved):
LGTM. Custom build here: https://openxt.ainfosec.com/buildbot/#/builders/22/builds/91

PR #1493, opened 2024-04-19
dom0-tweaks: Add delayed-seal initscript

dpsmith:
delayed-seal is an initscript to run on shutdown after VMs have been stopped.  We don't want dom0 doing MMIO operations with the TPM while a guest is running that could potentially snoop on the data.

It has additional code to try destroying any still-running VMs as a safety measure.

Needs the installer to touch /tmp/upgrade-delayed-seal to trigger the sealing.


crogers1:
Back in 2014 it looks like `updatemgr` was open sourced with disabled logic to detect running guest VMs.  At this point it's set to just error out but we could easily have it shutdown non-UIVM guests instead before proceeding with the update stage in the installer.git code.  Despite it being a haskell change, it should be relatively small, and it avoids adding the initscript and signaling with the /tmp/upgrade-delayed-seal flag.

I can code up something for `updatemgr` if you're cool with that @dpsmith

PR #1496, opened 2024-04-22
UIVM partitioned disk

dpsmith:
This PR converts the UIVM over to being a fully partitioned disk to enable the UIVM to be ran as HVM.


crogers1 (commented):
UIVM is currently PVH, I would think that's preferable to HVM in most cases for a guest like the UIVM.  Would this be a convenience change to support HVM UIVM in a downstream version?

shahs-ais (approved):
Looks OK to me, also curious about the motivation for the change (assume this doesn't affect PVH mode and just adds an option for HVM if needed down the line?)

PR #1498, opened 2025-04-04
Update patches/recipes

shahs-ais:
Updating patch contexts and package URLs to sync with external changes.

### DIFF
--- a/recipes-core/varstored/varstored_git.bb
+++ b/recipes-core/varstored/varstored_git.bb
@@ -46,7 +46,7 @@ SRC_URI += " \
     https://uefi.org/sites/default/files/resources/dbxupdate_x64.bin;downloadfilename=dbx.auth \
 "
 
-SRC_URI[sha256sum] = "46ba1f2a0a2ed7aabe20f9b7b2a8d717cb0b514cea83c7a1a24fe25f6b208784"
+SRC_URI[sha256sum] = "2378fdfe035a8373529ce9acb013fc31b59d3a71d4f9bbbc590bfc8536f90787"
 
 USERADD_PACKAGES = "${PN}"
 USERADD_PARAM_${PN} = "--system --no-create-home \
--- a/recipes-devtools/dojo/dojosdk-native.inc
+++ b/recipes-devtools/dojo/dojosdk-native.inc
@@ -2,7 +2,7 @@ DEPENDS = "icedtea7-native"
 
 inherit native
 
-SRC_URI = "http://download.dojotoolkit.org/release-${PV}/dojo-release-${PV}-src.tar.gz"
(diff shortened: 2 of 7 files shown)


crogers1:
Two of those commits should be squashed, the whitespace one and the "undo local change".  Also we don't usually commit binary blobs to the OE layers.  I will put the anthy package tarball up on our build mirror, and then your recipe change should point to that URL.

recipes-core/varstored/varstored_git.bb
```diff
@@ -46,7 +46,7 @@ SRC_URI += " \
     https://uefi.org/sites/default/files/resources/dbxupdate_x64.bin;downloadfilename=dbx.auth \
 "
 
-SRC_URI[sha256sum] = "46ba1f2a0a2ed7aabe20f9b7b2a8d717cb0b514cea83c7a1a24fe25f6b208784"
+SRC_URI[sha256sum] = "2378fdfe035a8373529ce9acb013fc31b59d3a71d4f9bbbc590bfc8536f90787"
```
crogers1, line 49: This is needed due to the patch context changes or for some other reason?

recipes-devtools/dojo/dojosdk-native.inc
```diff
@@ -2,7 +2,7 @@ DEPENDS = "icedtea7-native"
 
 inherit native
 
-SRC_URI = "http://download.dojotoolkit.org/release-${PV}/dojo-release-${PV}-src.tar.gz"
+SRC_URI = "https://download.dojotoolkit.org/release-${PV}/dojo-release-${PV}-src.tar.gz"
```
crogers1, line 5: Was this to fix a warning or an error?

recipes-core/varstored/varstored_git.bb
shahs-ais: It is because the UEFI revocation list changed earlier this year (see https://uefi.org/revocationlistfile) so the checksum is different for the download file now

recipes-devtools/dojo/dojosdk-native.inc
shahs-ais: This was an error

crogers1:
I still need to upload the anthy package to the mirror, but once that happens I think all the commits can be squashed into a single commit.

shahs-ais:
> I'm not the best commit message writer in the world, but the commit message only talks about anthy but the commit has changes all over the place. And then looking at the changes, you have formatting changes, src rev id changes, and src uri changes. I would not have smashed these all into a single commit.

Yes you're right, I accidentally axed some of the older commit messages when squashing, I can split it up into more logical commits/messages

dpsmith (approved):

PR #1499, opened 2025-05-13
Update nm applet and backend

shahs-ais:
Uprev to network-manager v1.16.0 and network-manager-applet v1.22.10 and revising associated patches


dpsmith:
Your commit really needs to be broken up to provide a clean evolution. libnma recipe should be its own commit followed by a commit that moves the applet patch and cleans up the applet recipe. And why did you lock the files folder to .10 for networkmanager instead of .%. Doing this will cause unnecessary patch churn if we have to do a bug update.

shahs-ais:
> Your commit really needs to be broken up to provide a clean evolution. libnma recipe should be its own commit followed by a commit that moves the applet patch and cleans up the applet recipe. And why did you lock the files folder to .10 for networkmanager instead of .%. Doing this will cause unnecessary patch churn if we have to do a bug update.

Will break it up into more reasonable slices, thanks. I was under the impression that we wanted to have the package versions locked for stability reasons, maybe I misunderstood this - will discuss with Chris as well.

dpsmith (approved):

crogers1 (approved):

